PR #8, opened 2017-08-18
Initial Nightwatch Refactor Changes

### LINKED ISSUE #7: Improve Nightwatch Testing
## Summary
Our team's goal is to improve the ease of using the terra-toolkit, starting with nightwatch testing. To improve nightwatch testing, we are going to:
- Provide simpler nightwatch configuration that allows easy customization
- Remove the test scripts needed to start nightwatch testing
- Provide a responsive test helper 
- Clean up documentation on nightwatch test setup

emilyrohrbough:
## Summary
Here is the first phase of the nightwatch refactor. This PR removes the current scripts and configuration from the toolkit to only provide nightwatch configuration and responsive test helper functions. This closes https://github.com/cerner/terra-toolkit/issues/7.

### Things to Note About Configuration
 - Headless chrome is the default (and only) test environment provided
 - Screenshots are enabled -  use `browser.saveScreenshot('imagepath')` in a test step

### Additional Notes
Documentation updates will come later.

## Examples of Use
**Setup Example:**
Nightwatch configuration is available via the  `nightwatchConfig` function in `nightwatch.config.js`. Takes nightwatch config, array of test directory paths and port number (optional). 

screen shot 2017-08-18 at 4 51 42 pm

**Test Suite Example:**
The nightwatch configuration now sets `browser.launchUrl` to `http://localhost:${port}` and then supplying the test steps to the responsive test helper `resizeTo` removes the need to include the beforeEach and after hooks. To test responsive elements, use `screenWidth(browser)`.

screen shot 2017-08-18 at 2 19 22 pm


emilyrohrbough:
One thing to note: the test_settings configuration `detailed_output` is set to true. This give explict feedback when a failure occurs saying what was expected vs what was received. This writes to the console as tests are ran.

emilyrohrbough:
Added `setup-helper.js` in my last commit to help with nightwatch configuration setup. Since the test_settings `filter` configuration only accepts a string, all subdirectories would also need to be listed for the nightwatch `srcFolders` configuration. Updated sample configuration above.

src/nightwatch/nightwatch.config.js
```diff
@@ -0,0 +1,80 @@
...
+let port = 8080;
+
+const nightwatchConfig = (webpackConfig, srcFolders, providedPort) => {
+  if (providedPort) {
+    port = providedPort;
+  }
+
+  const webpackServer = new WebpackDevServer(webpack(webpackConfig), { quiet: true, hot: false, inline: false });
+  const stopWebpackDevServer = done => webpackServer.close(() => done());
+  const startWebpackDevServer = done => webpackServer.listen(port, '0.0.0.0', () => done());
```
mhemesath: If we want to support multiple ports we can use something like https://github.com/indexzero/node-portfinder

I know this node developer, he is highly reputable, so the dependency should be legit.

src/nightwatch/setup-helper.js
```diff
@@ -0,0 +1,51 @@
+const path = require('path');
+const fs = require('fs');
+const loadJsonFile = require('load-json-file');
+const globSync = require('glob').sync;
+
+const interatePackageDirectories = (lernaJSON) => {
+  const srcFolders = [];
```
mhemesath: nitpic: JavaScript is a very functional language and usually a [declarative pattern is preferred over an imperative one](https://medium.com/javascript-scene/master-the-javascript-interview-what-is-functional-programming-7f218c68b3a0#c9b9). One sign that a function is imperative is usually the declaration of a variable at the top that is appended/modified throughout a series of operations. 

Declarative implementation:
```js
return  loadJsonFile.sync(lernaJSON).packages  // array of directories
  .map(globSync) // array of array of file paths
  .reduce((a,b) => a.concat(b), []) // Merge file paths to single array
  .map(packagePath => `${packagePath}/tests/nightwatch/`) // Get test file path
  .filter(fs.existsSync); // Filter to valid test paths
```

We shouldn't have to search sub-directories. If we do, its likely because our filter is `*-spec.js` and should actually be `**/*-spec.js`
https://github.com/nightwatchjs/nightwatch/blob/master/test/nightwatch.json#L11

src/nightwatch/setup-helper.js
```diff
@@ -0,0 +1,51 @@
...
+            srcFolders.push(subDir);
+          }
+        });
+      }
+    });
+  });
+  return srcFolders;
+};
+
+const interateDirectories = (basePath) => {
```
mhemesath: per my comment above, this may not be needed if we change our filter to use a glob pattern

src/nightwatch/responsive-helpers.js
```diff
@@ -0,0 +1,39 @@
+const screenWidth = (browser) => {
+  const breakpoint = browser.currentTest.name.match(/\[@(.*?)\]/)[1];
```
mhemesath, line 6: This is a good idea.. can I suggest we turn this into a [custom command](http://nightwatchjs.org/guide#writing-custom-commands)?   

```js
const breakpoint = browser.breakpoint();
```

It may also be nice to have custom commands that do something similar to

```js
browser.whenBreakpoint('small', () => {
  // todo assert things when screensize is small
});
```

Don't have to name the function exactly that, but a utility to make it easier to scope tests. One could also argue that rather than that function.. the assertions should just be broken into separate tests that only resize to what they are asserting.

src/nightwatch/nightwatch.config.js
```diff
@@ -0,0 +1,80 @@
...
+            medium: [838, 768],
+            large: [1000, 768],
+            huge: [1300, 768],
+            enormous: [1500, 768],
+          },
+          asyncHookTimeout: 30000,
+          before: startWebpackDevServer,
+          after: stopWebpackDevServer,
+        },
+        filter: '*-spec.js',
```
mhemesath: if we change this to `**/*-spec.js` we may not have to search sub-directories.

src/nightwatch/setup-helper.js
emilyrohrbough: I just retested this, I must have written `*/*-spec.js` initially because my first thought was also to use this filter pattern. Will update.

> ... usually a declarative pattern is preferred over an imperative one.

Good point. I will also make this change.

src/nightwatch/nightwatch.config.js
emilyrohrbough: I will take a look at this package. I think this would be valuable to include. Currently if a port is busy, nightwatch ends up hanging because webpack does not return any information when a listen error occurs, so we cannot close the server.

mhemesath: the only challenge this might create is not knowing which port to use.. not sure how we can set that for each process.

emilyrohrbough: Yeah, I took a look at this. I found I was able to hook in and search for an unused port and successfully start the webserver, however within the nightwatch configuration, we specify the urlLauncher as `http://localhost:${port}`. At this point, port becomes static to the initial value we assign and if the portFinder increments/assigns a different unused port, the tests  will still look for the initial assigned port.

I feel at this point this would be a nice extra, but what we have is probably fine since the current configuration giving users the option to specify their desired port number & makes it possible to run multiple instances from the command line.

src/nightwatch/responsive-helpers.js
emilyrohrbough: > const breakpoint = browser.breakpoint();

I took a look at updating to use this custom command, however, I think the current implementation is what we want. It seems custom commands are useful for pages interactions such as login steps, or repeated click events, etc. Writing a custom command to solely return the breakpoint value ends up returning the entire test object (which is what we started with). The alternative is to write a command with a callback function, but I have found if an error occurs, assertions within the callback neither pass nor fail.

emilyrohrbough: > browser.whenBreakpoint('small', () => {
  // todo assert things when screensize is small
});

At this point with our tests, I do not think a custom command like this would add any value given responsive component has different assertions depending on screen size and the current if-else is pretty simple and easy to read. Can you think of any specific examples where this would be better?

src/nightwatch/nightwatch.config.js
ryanthemanuel: Agreed.  I think we should log an issue to explore this in the future, but it's not worth holding up the PR.

src/nightwatch/responsive-helpers.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  browser.url(browser.launchUrl);
+  browser.resizeWindow(screenSize[0], screenSize[1], done);
+};
+
+const resizeTo = (breakpoints, suite) => {
+  const newTests = {};
+  breakpoints.forEach((breakpoint) => {
+    newTests[`Adjust Broswer to [@${breakpoint}]`] = () => resize(breakpoint);
```
mhemesath: mispelled browser

src/nightwatch/responsive-helpers.js
```diff
@@ -0,0 +1,42 @@
...
+    Object.keys(suite).reduce((tests, key) => {
+      const resizedTests = {};
+      const value = suite[key];
+      if (typeof value === 'function') {
+        resizedTests[`${key} [@${breakpoint}]`] = value;
```
mhemesath: I wonder if a slightly more elegant approach is to do a resize as part of the first test rather than have a separate test to do resizes.

emilyrohrbough:
It seems there are browser resizing issues with the current implementation which may do with the chromedriver. I am going to close this PR to investigate and reopen when things are cleared up.

This could just be an issue with my local chromedriver.

emilyrohrbough:
My local chromedriver was affecting test updates, but I also fixed the browser resizing issues.

src/nightwatch/responsive-helpers.js
emilyrohrbough: Agreed. Updated [here](https://github.com/cerner/terra-toolkit/pull/8/files#diff-0d9e98c0c20f0e65e7db09f64ca6e407R20)

ryanthemanuel:
Can we add a .gitignore folder similar to what we have in the other repos?

src/nightwatch/nightwatch.config.js
```diff
@@ -0,0 +1,80 @@
...
+            large: [1000, 768],
+            huge: [1300, 768],
+            enormous: [1500, 768],
+          },
+          asyncHookTimeout: 30000,
+          before: startWebpackDevServer,
+          after: stopWebpackDevServer,
+        },
+        filter: '**/*-spec.js',
+        screenshots: {
```
emilyrohrbough: @ryanthemanuel This configuration provides screenshots, which addresses https://github.com/cerner/terra-toolkit/issues/9. The outputted screenshot name is the test name.

emilyrohrbough:
@ryanthemanuel Here is the [link](https://github.com/cerner/terra-toolkit/blob/master/.gitignore) to our current .gitignore file. Is there anything in particular that we should add?

EDIT: Added `package-lock.json` for node update.

ryanthemanuel:
Oops.  I meant the .github folder.  I was noticing that we don't have the pull request templates and such.

emilyrohrbough:
@cerner/terra This PR should be good to go. Changes in core & clinical have been made and tested.

README.md
```diff
@@ -25,115 +24,7 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
-```javascript
-## Run the default driver (phantomjs)
-node ../../node_modules/terra-toolkit/lib/scripts/nightwatch-non-parallel.js
-
-## Run the safari driver
-node ../../node_modules/terra-toolkit/lib/scripts/nightwatch-non-parallel.js safari
-```
-
-The default nightwatch testing environment will be used unless the `WEBPACK_CONFIG_PATH` is specified.
+Updates are coming soon.
```
mjhenkes, line 27: lol

mjhenkes (approved):

bjankord (approved):

src/nightwatch/responsive-helpers.js
ryanthemanuel: I'm ok with leaving this as a potential enhancement in the future for now.

mhemesath (approved):

tests/nightwatch/static-suite-spec.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  'Runs the test suite correctly statically': (browser) => {
+    browser
+      .url(browser.launchUrl)
+      .assert.containsText('.test', 'Test');
```
ryanthemanuel: Let's make this an expect in case anyone looks at it for an example.

ryanthemanuel (approved):

PR #15, opened 2017-10-11
Lock webpack-dev-server at latest version that supported IE10 (2.7.1)

emilyrohrbough:
### Summary
Webpack > 2.7.1 dropped IE 10 & 11 support.
See https://github.com/cerner/terra-core/pull/901
Thanks for contributing to Terra. 
@cerner/terra


bjankord (approved):

package.json
```diff
@@ -48,7 +48,7 @@
     "react": "^15.4.2",
     "react-dom": "^15.4.2",
     "webpack": "^3.6.0",
-    "webpack-dev-server": "^2.9.1",
+    "webpack-dev-server": "2.7.1",
```
mhemesath: I don't think we should do this.. Example my team isn't locked down to IE10 and we can upgrade. The peer dependency should be flexible

emilyrohrbough: @mhemesath Updated.

mjhenkes (approved):

Issue #9, opened 2017-08-23
Add a Screenshot Grabbing Utility


ryanthemanuel:
# Issue Description
We need to add the ability to capture screenshots.  There's a lot of potential here, and this may need to be split up into several issues, but for documentation purposes the functionality we would like to have is:

* A screenshot grabber that captures screenshots
* Some mechanism to compare screenshots against an "official version"

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

emilyrohrbough:
The following pieces are needed:
- [ ] Ability to capture screenshot and compare
- [ ] Script to handle obsolete snapshots
- [ ] Script to handle snapshot management (i.e. stale snapshots)

emilyrohrbough:
#### Proposed Snapshot File Structure:
``` __snapshots__/locale/browser/snapshot-name```

#### Proposed Snapshot Name:
```ExampleName.TestId.SnapshotName.BrowserSize```

- TestID is identified via the test step name as the test step number inclosed in brackets. If a test step number is not provided, fallback to use test step name.

- Snapshot Name is an optional unique name to describe the particular instance of a test step. For example, 'Selected' or 'Disclose'. Defaults to the name 'Default'.

emilyrohrbough:
Resolved in #18.

PR #16, opened 2017-10-12
added snapshot command

mhemesath:
Added a snapshot utility to capture screenshots.

#9 

DO NOT MERGE - I still need to work out some details but this mostly works right now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
   "dependencies": {
     "chromedriver": "^2.31.0",
     "glob": "^7.1.1",
-    "gm": "^1.23.0",
     "jimp": "^0.2.28",
     "load-json-file": "^2.0.0"
   },
--- a/package.json
+++ b/package.json
@@ -42,6 +42,8 @@
   "dependencies": {
     "chromedriver": "^2.31.0",
     "glob": "^7.1.1",
+    "gm": "^1.23.0",
+    "jimp": "^0.2.28",
     "load-json-file": "^2.0.0"
   },
   "peerDependencies": {
--- a/src/nightwatch/snapshot.js
+++ b/src/nightwatch/snapshot.js
@@ -1,0 +1,37 @@
+const Jimp = require('jimp');
+const fs = require('fs');
+const path = require('path');
+
+
+module.exports = function (browser, done) {
+  const testDirectory = path.dirname(module.parent.filename);
+  const currentTest = browser.currentTest;
+
+  browser.execute('document.documentElement.getAttribute("lang")', [], (locale) => {
+    const snapshotDirectory = path.join(testDirectory, '__snapshots__', locale.value || 'en');
+
+    browser.screenshot(false, (data) => {
+      Jimp.read(Buffer.from(data.value, 'base64')).then((snapshot) => {
+        const name = currentTest.name.replace(/\s+/g, '-');
+        const browserName = browser.capabilities.browserName;
+        const moduleName = currentTest.module.split('/').slice(-1)[0];
+        const fileName = `${moduleName}_${name}_${browserName}${snapshot.bitmap.width}.jpg`;
+        const filePath = path.join(snapshotDirectory, fileName);
+
+        // Found a baseline
+        if (fs.existsSync(filePath)) {
+          Jimp.read(filePath).then((baseline) => {
+            const diff = Jimp.diff(snapshot.autocrop(), baseline);
+            browser.assert.equal(diff.percent, 0, `Baseline is ${diff.percent}% different`);
+            done();
+          });
+        // New Image
+        } else {
+          snapshot.autocrop().write(filePath, () => {
+            done();
+          });
+        }
+      });
+    });
+  });
+};
--- a/tests/nightwatch/MoreTests/nested-test-suite-spec.js
+++ b/tests/nightwatch/MoreTests/nested-test-suite-spec.js
@@ -1,9 +1,11 @@
 const { resizeTo, screenWidth } = require('../../../lib/nightwatch/responsive-helpers');
+const snapshot = require('../../../src/nightwatch/snapshot');
 
 module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous'], {
   'Runs the test suite correctly when responsive': (browser) => {
     browser
       .url(browser.launchUrl)
+      .perform(snapshot)
       .expect.element('.test').text.to.equal('Test');
   },
 
@@ -16,6 +18,6 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
       // eslint-disable-next-line no-console
         throw new Error(`The test breakpoint width is: ${testWidth} and the browser width is ${result.value.width}`);
       }
-    });
+    }).perform(snapshot);
   },
 });
--- a/tests/nightwatch/responsive-test-suite-spec.js
+++ b/tests/nightwatch/responsive-test-suite-spec.js
@@ -1,9 +1,11 @@
 const { resizeTo, screenWidth } = require('../../lib/nightwatch/responsive-helpers');
+const snapshot = require('../../src/nightwatch/snapshot');
 
 module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous'], {
   'Runs the test suite correctly when responsive': (browser) => {
     browser
       .url(browser.launchUrl)
+      .perform(snapshot)
       .expect.element('.test').text.to.equal('Test');
   },
 
@@ -16,6 +18,6 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
       // eslint-disable-next-line no-console
         throw new Error(`The test breakpoint width is: ${testWidth} and the browser width is ${result.value.width}`);
       }
-    });
+    }).perform(snapshot);
   },
 });
--- a/tests/nightwatch/static-suite-spec.js
+++ b/tests/nightwatch/static-suite-spec.js
@@ -1,7 +1,9 @@
+const snapshot = require('../../src/nightwatch/snapshot');
+
 module.exports = {
   'Runs the test suite correctly statically': (browser) => {
     browser
       .url(browser.launchUrl)
-      .expect.element('.test').text.to.equal('Test');
+      .perform(snapshot);
   },
 };


src/nightwatch/snapshot.js
```diff
@@ -0,0 +1,37 @@
...
+module.exports = function (browser, done) {
+  const testDirectory = path.dirname(module.parent.filename);
+  const currentTest = browser.currentTest;
+
+  browser.execute('document.documentElement.getAttribute("lang")', [], (locale) => {
+    const snapshotDirectory = path.join(testDirectory, '__snapshots__', locale.value || 'en');
+
+    browser.screenshot(false, (data) => {
+      Jimp.read(Buffer.from(data.value, 'base64')).then((snapshot) => {
+        const name = currentTest.name.replace(/\s+/g, '-');
```
mhemesath, line 15: thinking of biting the bullet here and just importing a module to poperly sanitize file names and camelcase.

src/nightwatch/snapshot.js
```diff
@@ -0,0 +1,37 @@
...
+        // Found a baseline
+        if (fs.existsSync(filePath)) {
+          Jimp.read(filePath).then((baseline) => {
+            const diff = Jimp.diff(snapshot.autocrop(), baseline);
+            browser.assert.equal(diff.percent, 0, `Baseline is ${diff.percent}% different`);
+            done();
+          });
+        // New Image
+        } else {
+          snapshot.autocrop().write(filePath, () => {
```
mhemesath, line 30: need to add a command to allow cropping to be disabled

src/nightwatch/snapshot.js
```diff
@@ -0,0 +1,37 @@
...
+        const browserName = browser.capabilities.browserName;
+        const moduleName = currentTest.module.split('/').slice(-1)[0];
+        const fileName = `${moduleName}_${name}_${browserName}${snapshot.bitmap.width}.jpg`;
+        const filePath = path.join(snapshotDirectory, fileName);
+
+        // Found a baseline
+        if (fs.existsSync(filePath)) {
+          Jimp.read(filePath).then((baseline) => {
+            const diff = Jimp.diff(snapshot.autocrop(), baseline);
+            browser.assert.equal(diff.percent, 0, `Baseline is ${diff.percent}% different`);
```
mhemesath, line 25: need to be able to allow threshold on diff to be configurable.

tests/nightwatch/static-suite-spec.js
```diff
@@ -1,7 +1,9 @@
+const snapshot = require('../../src/nightwatch/snapshot');
+
 module.exports = {
   'Runs the test suite correctly statically': (browser) => {
     browser
       .url(browser.launchUrl)
-      .expect.element('.test').text.to.equal('Test');
+      .perform(snapshot);
```
mhemesath, line 7: not using custom commands because due to the way they load I can't easily determine the filepath from where the test is running.

package.json
```diff
@@ -42,6 +42,7 @@
   "dependencies": {
     "chromedriver": "^2.31.0",
     "glob": "^7.1.1",
+    "jimp": "^0.2.28",
```
mhemesath, line 45: this project is nice because it doesn't require any external dependencies (e.g. imagemagick). Downside is it might be a bit slower though I'm not sure on that.

src/nightwatch/snapshot.js
```diff
@@ -0,0 +1,37 @@
...
+const fs = require('fs');
+const path = require('path');
+
+
+module.exports = function (browser, done) {
+  const testDirectory = path.dirname(module.parent.filename);
+  const currentTest = browser.currentTest;
+
+  browser.execute('document.documentElement.getAttribute("lang")', [], (locale) => {
+    const snapshotDirectory = path.join(testDirectory, '__snapshots__', locale.value || 'en');
```
emilyrohrbough, line 11: Probably be good to fallback to en-US since that is the site's starting locale

emilyrohrbough:
Can we upload the generated test files to see the cropping and naming structure?

src/nightwatch/snapshot.js
```diff
@@ -0,0 +1,37 @@
+const Jimp = require('jimp');
+const fs = require('fs');
+const path = require('path');
+
```
emilyrohrbough, line 4: nit: extra line

src/nightwatch/snapshot.js
emilyrohrbough: We discussed trying for the test step number and then falling back on the testStep name. For either:

`[1] Displays Correctly`  to be `name = [1]` or `Displays Correctly` to be `Displays-Correctly`

tests/nightwatch/responsive-test-suite-spec.js
```diff
@@ -16,6 +18,6 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
       // eslint-disable-next-line no-console
         throw new Error(`The test breakpoint width is: ${testWidth} and the browser width is ${result.value.width}`);
       }
-    });
+    }).perform(snapshot);
```
emilyrohrbough, line 21: We should add an addional parameter to accept a screenshot description. When one isn't provided, screenshot name is 'Default'.

src/nightwatch/snapshot.js
mhemesath: Yup, haven't done that yet

tests/nightwatch/responsive-test-suite-spec.js
mhemesath: yeah, need to update this to take options.. it'll likely look like

`.perform(snapshot(name, {}))`

mhemesath:
@emilyrohrbough I didn't upload screenshots cause its kinda  mess... dumps them all in one directory. When I fix that I can upload them.. jsut didn't want to pollute PR

tests/nightwatch/MoreTests/nested-test-suite-spec.js
```diff
@@ -1,9 +1,11 @@
 const { resizeTo, screenWidth } = require('../../../lib/nightwatch/responsive-helpers');
+const snapshot = require('../../../src/nightwatch/snapshot');
```
emilyrohrbough, line 2: use lib

src/nightwatch/snapshot.js
mhemesath: 👍 was also thinking of allowing this to be configurable in nightwatch config

package.json
emilyrohrbough: Do you have any other information regarding `jimp`?

mhemesath: https://github.com/oliver-moran/jimp

src/nightwatch/snapshot.js
emilyrohrbough: Is there ever a possibly where cropping fails or returns an empty image?

src/nightwatch/snapshot.js
```diff
@@ -0,0 +1,37 @@
...
+        const name = currentTest.name.replace(/\s+/g, '-');
+        const browserName = browser.capabilities.browserName;
+        const moduleName = currentTest.module.split('/').slice(-1)[0];
+        const fileName = `${moduleName}_${name}_${browserName}${snapshot.bitmap.width}.jpg`;
+        const filePath = path.join(snapshotDirectory, fileName);
+
+        // Found a baseline
+        if (fs.existsSync(filePath)) {
+          Jimp.read(filePath).then((baseline) => {
+            const diff = Jimp.diff(snapshot.autocrop(), baseline);
```
emilyrohrbough, line 24: It looks like the default difference threshold for Jimp is 0.1. IMO our threshold should probably be 0. It looks like there is a [bug](https://github.com/oliver-moran/jimp/issues/350) when defining the threshold as 0, but the workaround it using .01

src/nightwatch/snapshot.js
emilyrohrbough: This being said, we crop the image weather we compare or save it. If we provide a failing crop fallback, maybe we should we crop before doing either.

mhemesath: good point,  should probably fail on a zero dimension image.

src/nightwatch/snapshot.js
emilyrohrbough: Or should threshold an optional value that will default to .01?

mhemesath: we can do a perceptual hash compare if we want guaranteed identical. But yeah i'll update that to 0.1 and probably also allow it to be configurable

mhemesath: Actually... i check the diff percent below and verify thats zero...

src/nightwatch/snapshot.js
emilyrohrbough, line 25: nit: `Current snapshot is ${diff.percent}% different than the baseline.`

src/nightwatch/snapshot.js
emilyrohrbough: Yeah, it looks like Jimp uses the pixelmatch package that takes this threshold value to return the numDiffPixels which is used to calculate that percentage. Zero is what we'd want though.

emilyrohrbough:
Has there been work done regarding a snapshot script to offer snapshot updates for obsolete snapshots and to check for stale snapshots?

package.json
emilyrohrbough: Any information on choosing this package over another?

mhemesath: It has crop and compare with no external dependencies

mhemesath:
What is the difference between stale and obsolete

emilyrohrbough:
- Obsolete as in the snapshots failed due to changes we intended to make and need to be updated
- Stale as in a test step was removed or name was changed and this snapshot is no longer relevant/compared

src/nightwatch/snapshot.js
bjankord, line 4: This should be written as a custom nightwatch command: http://nightwatchjs.org/guide#writing-custom-commands

This would allow users to set `"custom_commands_path"` and point to a directory in toolkit which contains our custom commands without needing to import this file, `const snapshot = require('../../../src/nightwatch/snapshot');` in every test they need to perform the snapshot.

tests/nightwatch/static-suite-spec.js
bjankord: Ah I left a comment about this up above, I see the issue is around how we set `testDirectory`. You might check with the value of `__dir` is and see if that can be used in any way to determine where `snapshotDirectory` should be.

bjankord:
Obsolete snapshots could be updated with a command similar to [jest's update snapshot command](https://facebook.github.io/jest/docs/en/snapshot-testing.html#updating-snapshots). 

With the idea of passing options, one possible way would be to pass an option to update the snapshot

`.perform(snapshot(name, {update: true}))`

This would have to be run once to update and then removed so snapshots are not continually updated.

Stale snapshots could be removed manually, or removed via the options object
`.perform(snapshot(name, {remove: true}))`

emilyrohrbough:
@bjankord With the ideas you proposed, would this be something we could add in with default `update: false` and `remove: true` that could be possibly be switched via command line or manually?

tests/nightwatch/static-suite-spec.js
mhemesath: That will give the directory of where the `snapshot.js` file is.

mhemesath:
update and remove should be applied with a command line option IMO. e.g.

```
npm run nightwatch -- --snapshot-u
```

src/nightwatch/snapshot.js
mhemesath: I tried a custom command but I can't determine the location of which test is being executed. I  printed out the nightwatch API object to identify any possible way to figure it out and there is simply nothing there.

The problem is that nightwatch loads these plugins indirectly so when I find the module parent directory it evaluates to nightwatch and not the test running...

bjankord:
@emilyrohrbough @mhemesath Yeah I agree about the CLI usage. Do you know if we can access flags passed like `--snapshot-u` inside of `src/nightwatch/snapshot.js`? If so, we can use that to modify behavior of the `.perform(snapshot)` command to update the snapshot and pass the test disregarding any diffs.

PR #17, opened 2017-10-25
Windows Dev Compatibility

emilyrohrbough:
### Summary
Updated file paths to allow for development on Windows OS. @singleton06 was able to verify these changes.


package.json
```diff
@@ -24,17 +24,17 @@
   },
   "scripts": {
     "clean:all": "npm run clean:compiled && npm run clean:dependencies",
-    "clean:compiled": "rimraf ./lib",
-    "clean:dependencies": "rimraf ./node_modules",
+    "clean:compiled": "rimraf node lib",
```
ryanthemanuel: Should this be rimraf lib (and the next line rimraf node_modules)?

mjhenkes (approved):

package.json
dv297: Perhaps not a concern for Terra, but since rimraf is itself a node_module, this can make the flow chart of "when do I want to use install vs clean:install" a little bit more complicated. You could write a JS script that delete these folders using node itself and have this script committed. Then, you could run npm run clean:install, regardless of whether there is a node_module. It could possibly make on-boarding/ contributing to Terra a little more straight forward, as new devs could always run npm run clean:install to fully handle the after-clone process

StephenEsser (approved):

package.json
emilyrohrbough: Before any development happens, people should be running `npm install`, so I wouldn't think this is a huge issue, since rimraf should be installed by the time a person would use the script `npm run clean:install`. This is a valuable point, though out of the scope of this PR IMO. 

This makes me think we should log an issue to write a 'Contributors Getting Started' that indicates the setup needed to contribute to toolkit. Do you think this would clear up any confusion between these two scripts?

dv297: A getting started guide would be very helpful. Commands like this can often get hidden in a package.json and only used by core contributors who are in the know. 

[An issue for a contributor guide to terra-core already exists](https://github.com/cerner/terra-core/issues/940), so logging a parallel issue and being conscientious about considerations that are shared/unique across each can help consistency for these Cerner OS projects.

emilyrohrbough: Sounds good. I'll go ahead an log an issue for a Startup Guide for Toolkit.

bjankord (approved):

ryanthemanuel (approved):

package.json
emilyrohrbough: Issue Logged: https://github.com/cerner/terra-toolkit/issues/20

PR #18, opened 2017-10-26
Webdriver.io Accessibility and Visual Regression Testing

mhemesath:
### Summary
Added webdriver.io testing services/examples. Need to add README instructions on how to consume this.


src/wdio/visualcompare.js
```diff
@@ -0,0 +1,23 @@
+import path from 'path';
+import { LocalCompare } from 'wdio-visual-regression-service/compare';
+
+function getScreenshotName(ref) {
```
mhemesath: this logic needs to be tweaked a little to add a few requirements @emilyrohrbough had

wdio.conf.js
```diff
@@ -0,0 +1,74 @@
...
+    javascriptEnabled: true,
+    acceptSslCerts: true,
+    chromeOptions: {
+      args: [
+        'headless',
+        'no-sandbox',
+      ],
+    },
+  }, {
+    browserName: 'firefox',
```
mhemesath: headless firefox WOOT WOOT

package.json
```diff
@@ -45,6 +45,7 @@
     "chai": "^4.1.2",
     "chromedriver": "^2.31.0",
     "glob": "^7.1.1",
+    "ip": "^1.1.5",
```
emilyrohrbough: For those curious. This package is described as:

>IP address utilities for node.js

 https://github.com/indutny/node-ip

mhemesath: There may be a better solution than this.. but being we have to run visual tests within a container, this is used to find the IP address of the host machine so the headless browser in the container can access it.

README.md
```diff
@@ -23,8 +23,191 @@ Terra Toolkit is a utility module used to facilitate independent development of
 
 - Install with [npm](https://www.npmjs.com): `npm install terra-toolkit --save-dev`
```
emilyrohrbough, line 26: Lets update the summary of Toolkit to specify it provides Nightwatch Utility, as well as Webdrive.io Utility. Then, we need to update the links above to include 
`- [Webdriver.io Utility](#webdriver.io-utility)

README.md
```diff
@@ -23,8 +23,191 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+// Validate it matches
+expect(screenshots).to.matchReference();
+// Validate it does not match
+expect(screenshots).to.not.matchReference();
+```
+
+
+
+
+## Nightwatch
```
emilyrohrbough: Update to `Nightwatch Utility` to keep the table of contents link & summary reference.

package.json
```diff
@@ -37,12 +37,18 @@
     "release:patch": "npm test && node ./scripts/release/release.js patch",
     "nightwatch": "nightwatch",
     "nightwatch:port-provided": "nightwatch -c tests/test.nightwatch.conf.js",
-    "test": "npm run compile && npm run nightwatch && npm run nightwatch:port-provided"
+    "test": "npm run compile && npm run nightwatch && npm run nightwatch:port-provided && wdio wdio.conf.js",
+    "wdio": "npm run compile && wdio wdio.conf.js"
```
emilyrohrbough: Clean up before pushing. Would expect 

```
"test": "npm run compile && npm run nightwatch && npm run nightwatch:port-provided && npm run wdio",
"wdio": "wdio wdio.conf.js"
```

src/wdio/services/AxeService.js
```diff
@@ -0,0 +1,47 @@
+/* global browser, axe */
+import fs from 'fs';
+
+let axeCoreSrc = fs.readFileSync('node_modules/axe-core/axe.min.js', 'utf8');
+axeCoreSrc = axeCoreSrc.replace(/^\/\*.*\*\//, '');
```
emilyrohrbough: Will need to test this with windows

wdio.conf.js
```diff
@@ -0,0 +1,73 @@
+const localIP = require('ip');
+
+const staticServerPort = 4567;
+const AxeService = require('./lib/wdio/services').Axe;
+const TerraService = require('./lib/wdio/services').Terra;
+const visualRegression = require('./lib/wdio/visualcompare');
+
+// Force firefox into headless mode
+process.env.MOZ_HEADLESS = '1';
```
emilyrohrbough: Is this something we can setup in the config?

src/wdio/services/AxeService.js
mhemesath: why would this change on windows? ALl its doing is stripping out leading comments in the JS so it injects nice

wdio.conf.js
mhemesath: this wouldn't matter anymore as we would run firefox within docker anyway

src/wdio/services/TerraService.js
```diff
@@ -0,0 +1,53 @@
+import chai from 'chai';
```
emilyrohrbough, line 1: Lets document each function in AxeService & TerraService so its easy to understand for consumers using these pieces.

package.json
```diff
@@ -37,12 +37,18 @@
...
   },
   "dependencies": {
+    "axe-core": "^2.4.2",
+    "chai": "^4.1.2",
```
emilyrohrbough: For those interested: https://www.npmjs.com/package/chai
> Chai is a BDD / TDD assertion library for node and the browser that can be delightfully paired with any javascript testing framework.

src/wdio/services/TerraService.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const viewport = (...sizes) => {
+  const widths = {
+    tiny: { width: 470, height: 768 },
+    small: { width: 622, height: 768 },
+    medium: { width: 838, height: 768 },
+    large: { width: 1000, height: 768 },
+    huge: { width: 1300, height: 768 },
+    enormous: { width: 500, hegiht: 768 },
```
emilyrohrbough: oops. Enormous is the incorrect width

src/wdio/services/TerraService.js
```diff
@@ -0,0 +1,72 @@
...
+};
+
+
+export default class TerraService {
+  // eslint-disable-next-line class-methods-use-this, consistent-return
+  onPrepare() {
+    if (!process.env.TRAVIS) {
+      this.containerId = exec('docker run -d -p 4444:4444 selenium/standalone-chrome');
+      // Sleep a few seconds to let selenium startup
+      return new Promise((resolve) => {
```
emilyrohrbough: Clean-up to ensure successful status.

src/wdio/services/TerraService.js
mhemesath: These should be documented in the README.

README.md
```diff
@@ -23,8 +23,191 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+## Webdriver.io
+[Webdriver.io](http://webdriver.io/) is a framework for writing webdriver powered tests to validate functionality in browsers. Webdriver.io also provides easy services for setting up selenium, starting webpack and static servers, as well as accessibilty and visual regression testing.
 
-## Nightwatch Utility
+
+Install the package
+```sh
+npm install webdriverio --save-dev
+```
+Generate a configuration file. [See documentation](http://webdriver.io/guide/testrunner/gettingstarted.html) for more info.
```
ryanthemanuel: I prefer that we provide a "recommended" configuration file.

mhemesath: Recommended for what setup? Jenkins runs, Travis, local, hitting our internal selenium?

package.json
```diff
@@ -71,16 +77,19 @@
...
-    "webpack": "^3.6.0",
-    "webpack-dev-server": "^2.9.1",
-    "node-sass": "^4.5.2",
-    "postcss-loader": "^2.0.6",
+    "sass-loader": "^6.0.6",
     "shelljs": "^0.7.6",
     "style-loader": "^0.19.0",
-    "raw-loader": "^0.5.1",
-    "sass-loader": "^6.0.6"
+    "wdio-mocha-framework": "^0.5.11",
```
ryanthemanuel: If we want to recommend these 'wdio' items, I suggest adding them as dependencies so that they get transitively pulled in.

src/wdio/services/AxeService.js
```diff
@@ -0,0 +1,47 @@
+/* global browser, axe */
+import fs from 'fs';
+
+let axeCoreSrc = fs.readFileSync('node_modules/axe-core/axe.min.js', 'utf8');
```
ryanthemanuel: I would make this fs.readFileSync(fs.readFileSync(path.join(__dirname, '..', '..', '..', 'node_modules', 'axe-core', 'axe.min.js'), 'utf8')

ryanthemanuel:
It looks like the tests here are all just static html.  Is there an example of how this will work to test something within terra-core?  I think that would be beneficial for consumers.

README.md
ryanthemanuel: Yes for all of those cases.  I'd like to provide some recommendations to the 9 questions that get asked by running "./node_modules/.bin/wdio config".  We already have enough problems with people trying to debug "non-standard" configurations.  I'd like to give people a starting point for the more common scenarios so they have to make as few decisions as possible.

ryanthemanuel: So I guess not *one* recommended file, but recommendations for the common scenarios our consumers will be dealing with.

src/wdio/services/AxeService.js
ryanthemanuel: I edited that a few times and there may be a better alternative.  We should make sure this is available in various development scenarios (hitting terra-toolkit locally, from git, etc.)

ryanthemanuel:
FYI, I've been working on getting headless chrome working with our standard build jobs in the orion space internally.  I'm going to play around with this branch to see how it works as well.

src/wdio/services/TerraService.js
emilyrohrbough: Agreed, but for maintainability purposes I think it'd be good to add documentation throughout here as well.

mhemesath:
>  It looks like the tests here are all just static html. Is there an example of how this will work to test something within terra-core? I think that would be beneficial for consumers.

@ryanthemanuel  I'm not going to create an example of running with webpack but I can link out to documentation on how to do that. Basically its just a service you configure. Really I see consumers running tests in several ways:
* Static Server
* Webpack DevServer
* Against an externally hosted environment
* Storybook

Each of those configurations is unique to the test setup and I think its sufficient to point to documentation that already exists on how to set them up.

src/wdio/services/TerraService.js
mhemesath: Added, sorry I misinterpreted this comment. I agree source should be documented.

wdio.conf.js
```diff
@@ -0,0 +1,77 @@
...
+  //
+  // Saves a screenshot to a given path if a command fails.
+  screenshotPath: './errorShots/',
+  //
+  // Set a base URL in order to shorten url command calls. If your url parameter starts
+  // with "/", then the base url gets prepended.
+  baseUrl: `http://${localIP.address()}:${staticServerPort}`,
+  //
+  // Default timeout for all waitFor* commands.
+  waitforTimeout: 10000,
```
bjankord: Is this a number we should lower, I'm thinking something like `3000`.

bjankord (approved):

wdio.conf.js
```diff
@@ -0,0 +1,26 @@
...
+
+const staticServerPort = 4567;
+
+const config = {
+  baseUrl: `http://${localIP.address()}:${staticServerPort}`,
+
+  staticServerPort,
+  staticServerLog: false,
+  staticServerFolders: [
+    { mount: '/', path: './tests/fixtures' },
```
emilyrohrbough: Update to use path.join for windows dev.

src/wdio/services/SeleniumDockerService.js
```diff
@@ -0,0 +1,121 @@
...
+export default class SeleniumDockerService {
+  constructor() {
+    this.getSeleniumStatus = this.getSeleniumStatus.bind(this);
+  }
+
+  getSeleniumStatus(callback) {
+    http.get({
+      host: this.host,
+      port: this.port,
+      path: path.join(this.path || '/wd/hub', 'status'),
```
emilyrohrbough, line 20: thinking we need to split '/wd/hub/' out to 'wd', 'hub'

src/wdio/conf.js
```diff
@@ -0,0 +1,36 @@
+const AxeService = require('./services').Axe;
+const TerraService = require('./services').Terra;
+const visualRegression = require('./visualcompare');
+const SeleniumDockerService = require('./services').SeleniumDocker;
+
+exports.config = {
+  specs: ['./tests/specs/**/*.js'],
```
emilyrohrbough: does wdio handle windows paths or should we update this to regex?

src/wdio/conf.js
```diff
@@ -0,0 +1,36 @@
...
+    {
+      browserName: 'chrome',
+    },
+  ],
+
+  sync: true,
+  logLevel: 'silent',
+  coloredLogs: true,
+  bail: 0,
+  screenshotPath: './errorShots/',
```
emilyrohrbough: ^ same as above

src/wdio/services/AxeService.js
```diff
@@ -0,0 +1,64 @@
...
+
+    browser.addCommand('axe', (options) => {
+      // Conditionally inject axe. This allows consumers to inject it themselves
+      // in the test examples which would slightly speed up test runs.
+      if (axeConfig.inject) {
+        if (!axeCoreSrc) {
+          axeCoreSrc = fs.readFileSync(require.resolve('axe-core'), 'utf8');
+          axeCoreSrc = axeCoreSrc.replace(/^\/\*.*\*\//, '');
+        }
+        if (browser.execute('return window.axe === undefined;')) {
```
emilyrohrbough: Should it make sense to check this if first before creating the axeCoreSrc if its unavailable?

src/wdio/services/AxeService.js
```diff
@@ -0,0 +1,64 @@
...
+      };
+
+      if (viewports.length === 0) {
+        viewports.push(currentViewportSize);
+      }
+
+      // Get accessibility results for each viewport size
+      const results = options.viewports.map((viewport) => {
+        browser.setViewportSize(viewport);
+        // eslint-disable-next-line func-names, prefer-arrow-callback
```
emilyrohrbough: I'm curious. Was there a particular reason to not use the arrow callback syntax?

src/wdio/services/AxeService.js
mhemesath: good idea

src/wdio/services/AxeService.js
mhemesath: that function is injected into the browser... won't work in IE10/11.

tests/specs/axe-spec.js
```diff
@@ -0,0 +1,42 @@
+/* global browser, describe, it, expect, viewport */
+describe('axe', () => {
+  const viewports = viewport('tiny', 'huge');
```
emilyrohrbough: Can one of theses be updated to not use `viewports` specifically to hit this [case](https://github.com/cerner/terra-toolkit/pull/18/files#diff-f1b44da52f6b6b95ff1f06b8ca99cb3cR38).

src/wdio/services/SeleniumDockerService.js
mhemesath: that doesn't really provide any value being `this.path` will be a pre-assembled path as well. Its really the path to the selenium hub endpoint wihch shouldn't change much. I just need to be able to reliably append `status` to it and not have to worry about a double //

src/wdio/services/AxeService.js
```diff
@@ -0,0 +1,64 @@
...
+      }
+
+      const currentViewportSize = browser.getViewportSize();
+      const viewports = options.viewports;
+      const axeOptions = {
+        runOnly: options.runOnly,
+        rules: options.rules,
+      };
+
+      if (viewports.length === 0) {
```
emilyrohrbough: Nit: maybe use `specifiedViewports` to help clarify we'd prefer to use the indicated viewports over the current.

Or possibly add inline comment saying something like // If not specified viewports use currrent

src/wdio/services/AxeService.js
```diff
@@ -0,0 +1,64 @@
...
+        viewports.push(currentViewportSize);
+      }
+
+      // Get accessibility results for each viewport size
+      const results = options.viewports.map((viewport) => {
```
emilyrohrbough: Update `options.viewports` to be `viewports`

src/wdio/services/AxeService.js
```diff
@@ -0,0 +1,64 @@
...
+        browser.setViewportSize(viewport);
+        // eslint-disable-next-line func-names, prefer-arrow-callback
+        return browser.executeAsync(function (context, opts, done) {
+          // eslint-disable-next-line func-names, prefer-arrow-callback
+          axe.run(context || document, opts, function (error, result) {
```
emilyrohrbough: There's a lot happen in these nested functions. Could we break them out?

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+## Webdriver.io Utility
+[Webdriver.io](http://webdriver.io/) is a framework for writing webdriver powered tests to validate functionality in browsers. Webdriver.io also provides easy services for setting up selenium, starting webpack and static servers, as well as accessibilty and visual regression testing.
+
+To make testing easier, Terar toolkit provides default configuration that enables the following:
+
+* Selenium docker instances for a consistent testing environment
+* Axe accessibility Utility
+* Visual regression testing
+* Viewport resizing helpers
+* Mocha test framework.
```
emilyrohrbough: nit: remove `.`

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+
+### Selenium Docker
+Selenium docker is provided as a convenience to make selenium testing easier and more stable. Running selenium in a container ensures a consistent testing environment across testing environments which is critical for visual regression testing.
+
+#### Options
+
+Under the key `seleniumDocker` in your wdio.conf.js you can pass a configuration object with the following structure:
+
+* **cidfile** - The name of the docker cidfile used to manage the docker instance during tests. Defaults to '.docker_selenium_id',
+* **enabled** - Flat to disable selenium docker; useful for CI environments which can startup the docker instance outside of test runs. Defaults to true
```
emilyrohrbough: "Flat to disabled..." Typo here?  EDIT: update to false or 'enables selenium docker'

Also, missing `.` after "Defaults to true"

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+
+#### Example
+```js
+// wdio.conf.js
+const wdioConf = require('terra-toolkit/wdio/conf');
+const localIP = require('ip');
+
+const staticServerPort = 4567;
+
+const config = {
```
emilyrohrbough, line 114: We should prob update these examples to spread the default config first since people will likely copy & paste.

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+    inject: false,
+  },
+  ...wdioConf.config,
+};
+
+exports.config = config;
+```
+
+
+#### Usage
```
emilyrohrbough, line 129: Can we consolidate `inject` with these options since they usage both wrap in the same object?

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+  A css selector to scope the accessibility test to. See [axe documentation](https://www.axe-core.org/docs/)
+
+#### Examples
+```js
+it('ignores inaccessibility based on rules', () => {
+  browser.url('/inaccessible-contrast.html');
+  const rules = {
+    'color-contrast': { enabled: false },
+  };
+  expect(browser.axe({ viewports, rules })).to.be.accessible();
```
emilyrohrbough, line 153: You never specified viewports within the example.

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+```
+
+
+#### Usage
+
+`browser.axe([{options}]);`
+
+The following options are available:
+
+* **viewports**:
```
emilyrohrbough, line 135: In you're tests you specify viewports as 'tiny', etc which maps to this {width, height} obj in the terra service. I would maybe update this to mention when using the terraService this is an option

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+```
+
+### Selenium Docker
+Selenium docker is provided as a convenience to make selenium testing easier and more stable. Running selenium in a container ensures a consistent testing environment across testing environments which is critical for visual regression testing.
+
+#### Options
+
+Under the key `seleniumDocker` in your wdio.conf.js you can pass a configuration object with the following structure:
+
+* **cidfile** - The name of the docker cidfile used to manage the docker instance during tests. Defaults to '.docker_selenium_id',
```
emilyrohrbough: Opps use `.` not `,` after '.docker_selenuium_id'

README.md
emilyrohrbough: Or mention when used with terraService use viewport helper & link to section

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+});
+```
+
+### Test Helpers
+
+The terra toolkit provides custom assertions and globals to make testing easier.
+
+#### Viewports
+
+The `viewports()` function provides a convenience method for fetching an array of viewports to use to resize the browser or even pass directly to the accessibility and visual regression commands.
```
emilyrohrbough: In the code, this is named viewport(). IMO viewports() would is what we want

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+    viewports.forEach(size, () => {
+      // Resize browser to each viewport size
+      browser.setViewportSize(size);
+    });
+  });
+});
+```
+
+#### Assertions
+2 custom assertions are provided to make validating the output of the visual regression and accesibility commands easier.
```
emilyrohrbough: Update '2' to 'Two'

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+
+
+`accessible()`
```
emilyrohrbough: Describe what accessible does

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+
+```js
+// Validate it is accessible
+expect(browser.axe()).to.be.accessible();
+// Validate it is not accessible
+expect(browser.axe()).to.not.be.accessible();
+```
+
+
+`matchReference()`
```
emilyrohrbough, line 249: Describe what matchReference does & that it utilizes capture

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
 
+## Webdriver.io Utility
+[Webdriver.io](http://webdriver.io/) is a framework for writing webdriver powered tests to validate functionality in browsers. Webdriver.io also provides easy services for setting up selenium, starting webpack and static servers, as well as accessibilty and visual regression testing.
+
+To make testing easier, Terar toolkit provides default configuration that enables the following:
+
+* Selenium docker instances for a consistent testing environment
+* Axe accessibility Utility
+* Visual regression testing
+* Viewport resizing helpers
```
emilyrohrbough: Is this what you meant by providing Viewport resizing helpers `https://github.com/cerner/terra-toolkit/pull/18/files#diff-04c6e90faac2675aa89e2176d2eec7d8R210`. I'd maybe reword this

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+* Mocha test framework.
+
+### Installation
```
emilyrohrbough: nit: Setup

README.md
emilyrohrbough: Also, will these assertions work if a user changes the wdio.framework from mocha? Might be good to indicate if not.

src/wdio/conf.js
```diff
@@ -0,0 +1,37 @@
...
+const path = require('path');
+
+exports.config = {
+  specs: [path.join('.', 'tests', 'specs', '**', '*.js')],
+
+  maxInstances: 10,
+
+  capabilities: [
+    {
+      browserName: 'chrome',
```
emilyrohrbough, line 14: The Selenium Service that handles which browser is used (chrome/firefox), should this be applied in the axe service such that the testing will be consistent by default?

src/wdio/services/SeleniumDockerService.js
```diff
@@ -0,0 +1,121 @@
...
+    });
+  }
+
+  /**
+   * Get the docker image to use based on configuration and browser capabilities.
+   * @return The name of the docker image to use.
+   */
+  getImage() {
+    // Use configured image or infer image from browserName (only firefox and safari supported).
+    // TODO: Eventually an entire hub should be stood up which supports all browsers in capabilities config
```
emilyrohrbough, line 97: Is this TODO something we should log an issue to and add reference to here?

src/wdio/services/SeleniumDockerService.js
```diff
@@ -0,0 +1,121 @@
...
+    return this.config.image || `selenium/standalone-${this.browserName}`;
```
emilyrohrbough, line 98: is there possible file path conflict with windows? `selenium/standalone-${this.browserName}`

src/wdio/services/TerraService.js
```diff
@@ -0,0 +1,70 @@
+import chai from 'chai';
+
+/**
+* accessible chai assertion to be paired with browser.axe() tess.
```
emilyrohrbough: type: "tess." to "tests."

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+#### Example
+```js
+// wdio.conf.js
+const wdioConf = require('terra-toolkit/wdio/conf');
+const localIP = require('ip');
+
+const staticServerPort = 4567;
+
+const config = {
+  baseUrl: `http://${localIP.address()}:${staticServerPort}`,
```
emilyrohrbough: Is this required for correct setup?

src/wdio/services/SeleniumDockerService.js
```diff
@@ -0,0 +1,121 @@
...
+   */
+  onPrepare(config, capabilities) {
+    this.config = {
+      cidfile: '.docker_selenium_id', // The docker container id file
+      enabled: true, // True if service enabled, false otherwise
+      cleanup: false, // True if docker container should be removed after test
+      image: null, // The image name to use, defaults to selenium/standalone-${browser}
+      ...(config.seleniumDocker || {}),
+    };
+    this.host = config.host;
```
emilyrohrbough: In the example provided in the README, consumer's specif the browserURL in the config, where is this passed here? Also, is host or path ever set? It defaults weren't provided in `src/wdio/conf.js`

src/wdio/services/SeleniumDockerService.js
```diff
@@ -0,0 +1,121 @@
...
+    this.cidfile = this.config.cidfile || this.cidfile;
+
+    return new Promise((resolve, reject) => {
+      if (this.config.enabled) {
+        const containerId = this.getContainerId();
+        if (!containerId) {
+          exec(`docker run --rm --cidfile ${this.cidfile} -p ${config.port}:4444 ${this.getImage()}`);
+        }
+        // Retry for 500 times up to 5 seconds for selenium to start
+        retry({ times: 500, interval: 10 }, this.getSeleniumStatus, (err, result) => {
```
emilyrohrbough: I feel like we should expose the selenium times & interval in the config.

src/wdio/services/SeleniumDockerService.js
```diff
@@ -0,0 +1,121 @@
...
+    this.port = config.port;
+    this.path = config.path;
+    this.browserName = capabilities[0].browserName;
+    this.cidfile = this.config.cidfile || this.cidfile;
+
+    return new Promise((resolve, reject) => {
+      if (this.config.enabled) {
+        const containerId = this.getContainerId();
+        if (!containerId) {
+          exec(`docker run --rm --cidfile ${this.cidfile} -p ${config.port}:4444 ${this.getImage()}`);
```
emilyrohrbough, line 73: I feel like we should expose the `4444` as an seleniumSetup option in the config

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
 
 - Install with [npm](https://www.npmjs.com): `npm install terra-toolkit --save-dev`
 
+## Webdriver.io Utility
+[Webdriver.io](http://webdriver.io/) is a framework for writing webdriver powered tests to validate functionality in browsers. Webdriver.io also provides easy services for setting up selenium, starting webpack and static servers, as well as accessibilty and visual regression testing.
+
+To make testing easier, Terar toolkit provides default configuration that enables the following:
```
JakeLaCombe: *Terra

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+  browser.url('/inaccessible-text.html');
+  const runOnly = {
+    type: 'tag',
+    values: ['color-contrast'],
+  };
+  expect(browser.axe({ viewports, runOnly })).to.be.accessible();
+});
+
+it('runs only specified context', () => {
+  browser.url('/inaccessible-contrast.html');
```
JakeLaCombe, line 166: Do you need to direct to a specific URL when providing this specific context? Also, this feels more like content than context.

src/wdio/visualcompare.js
```diff
@@ -0,0 +1,37 @@
+import path from 'path';
```
emilyrohrbough: I'm missing it, but where is this file used? There wasn't reference in the README and it didn't seem to have any test cases.

README.md
```diff
@@ -23,6 +24,229 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+  const viewports = viewport('tiny', 'huge');
+  it('checks visual comparison', () => {
+    browser.url('/buttons.html');
+
+    // Verify the button element matches
+    let screenshots = browser.checkElement('button', { viewports });
+    expect(screenshots).to.matchReference();
+
+    // Verify the whole viewport matches
+    let screenshots = browser.checkViewprot({ viewports });
```
JakeLaCombe: *checkViewport

scripts/release/release.js
```diff
@@ -3,7 +3,7 @@ const shell = require('shelljs');
 
 const releaseType = process.argv[2] || 'patch';
 
-if (shell.exec(`npm run compile && npm run lint && npm version ${releaseType} -m "Released version %s" && npm publish && git push`).code !== 0) {
+if (shell.exec(`npm version ${releaseType} -m "Released version %s" && npm publish && git push`).code !== 0) {
```
JakeLaCombe: How come this was removed?

src/wdio/visualcompare.js
mhemesath: Its used in the base config... Its our custom config for the visualregression services. Consumers don't need to do anything with it.

src/wdio/services/SeleniumDockerService.js
mhemesath: I don't think so, this is the image name in docker registry, not a path. With that said, I do need to try to run this on windows

src/wdio/services/SeleniumDockerService.js
mhemesath: Probably yeah

src/wdio/services/SeleniumDockerService.js
mhemesath: The 4444 is the port that runs on the docker container... the external port to map that to is configurable

src/wdio/services/SeleniumDockerService.js
mhemesath: host is the selenium host which is first class config in webdriver.io which is different than the browserURL. I can't really default setup the browser URL in the base config being it depends on external factors (is IP remote or local, what is port).

src/wdio/conf.js
mhemesath: > should this be applied in the axe service such that the testing will be consistent by default?

What do you mean?

README.md
mhemesath: its not so much mocha but rather the chai framework which i configure by default.

README.md
mhemesath: Context is following the `axe` api naming convention for what they describe as scoping tests to an element.

README.md
mhemesath: inject is a `wdio.conf.js` option for axe... The options under usage are to be passed directly into the command.  Its possible we can also pass in the below options globally... I recommend doing that as enhancement to this project after this PR. https://www.axe-core.org/docs/#section-2-api-reference

JakeLaCombe (approved):

README.md
mhemesath: For this project yes.... It is critical to allow the docker container to find the host. This setup also coincidentally allows a remote selenium hub to access the host machine (assuming it can be accessed through the network).

scripts/release/release.js
mhemesath: I didn't do that... that may have been a merge artifact. I can reset it to match master

src/wdio/visualcompare.js
emilyrohrbough: OH duh. Yeah I see its the config which was referenced [here](https://github.com/cerner/terra-toolkit/pull/18/files#diff-d668d1220134a102bb565faaf3b911afR3). Can we update this file name to be something like visual_regression.conf.js? it sorta sounds like a helper which why I thought to look in the test cases

tests/specs/axe-spec.js
mhemesath: 👍

src/wdio/services/AxeService.js
mhemesath: 👍

src/wdio/services/AxeService.js
mhemesath: good catch!

README.md
mhemesath: a yeah, good idea

README.md
mhemesath: not against that, but then the challenge with that is then you can't do this:
```js
// ERROR: viewports is already defined
const viewports = viewports('tiny');
```
Having the variable named viewports is nice cause then you can use object shorthand syntax rather than saying `{ viewports: viewports }` one can simply do `{ viewports }`

src/wdio/services/SeleniumDockerService.js
mhemesath: 👍

README.md
mhemesath: I'll call it out in the example documentation below

README.md
emilyrohrbough: If this is required for correct setup, we should somehow highlight it or instruct people that baseURL must be provided. From looking at this, it sorta looks like an additional custom config variable people might assume we are overriding.

README.md
```diff
@@ -23,6 +24,236 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+
+const staticServerPort = 4567;
+
+const config = {
+  baseUrl: `http://${localIP.address()}:${staticServerPort}`,
+  seleniumDocker: {
+    // Disable if running in Travis
+    enabled: !process.env.TRAVIS,
+  },
+  ...wdioConf.config,
```
emilyrohrbough: Move `...wdioConf.config,` before custom config.

ryanthemanuel: I think the config previously in this file is additional config, not custom config in this case.

mhemesath: both are true. i can update it.

README.md
mhemesath: Some of this comes down to understanding what you are consuming. The examples have baseUrl configured and it should be obvious its needed. It's not something special for toolkit

README.md
emilyrohrbough: ^ what you say makes sense. That sorta makes me want to change the name of `viewport()` to something else then since it doesn't read like it would return multiple viewport sizes. Possible ideas could be 'screenSizes()', 'viewportSizes()', terraViewports(), screenDimensions().

Either way, the documentation is incorrect. 
> The `viewports()` function

should be updated to The _viewport_() function

README.md
```diff
@@ -23,6 +24,236 @@ Terra Toolkit is a utility module used to facilitate independent development of
...
+  });
+});
+```
+
+#### Assertions
+Two custom assertions are provided to make validating the output of the visual regression and accesibility commands easier.
+
+
+`accessible()`
+Convenience method to validate the `axe()`` accessibility checks across all tested viewports are successful.
```
emilyrohrbough: oops. have an extra ` after axe()

README.md
mhemesath: I can add a comment in the code example

src/wdio/conf.js
emilyrohrbough: Okay. Missed these config overlaps.

What I was thinking would have been axe and selenium should use the same config for testing in either chrome or firefox. Missed the fact the wdio config drives the selenium config for this.

src/wdio/services/SeleniumDockerService.js
```diff
@@ -0,0 +1,125 @@
...
+        }
+        // Retry for 500 times up to 5 seconds for selenium to start
+        retry({ times: this.config.retries, interval: this.config.retryInterval },
```
ryanthemanuel, line 76: I ended up getting this running both locally and in jenkins for an internal project.  The one issue that I ran into was that to get it working with our current set up, I needed to know the host that was assigned when the 'docker run' command was executed.  

This wasn't 127.0.0.1 because it was a docker container running (sort of) within another docker container.  I figured out how to determine and update the host dynamically to get this working though.  Not sure if that's something we'll want to do at some point, but I don't think it's worth it for now (unless there are other cases where 127.0.0.1 will not be the host and it can't be predetermined ahead of time).

In the long run I'd like to improve our internal setup so that we can stand up the selenium standalone containers before the test runs even start (similar to how Travis does it and as Mike suggested).  I need to do a little more digging to figure out how to get that working with our set up though.

Here's the jenkins build in case you're curious:  https://jenkins.cerner.com/ion/job/ion-chart/job/ion-chief-complaint-js/job/test_webdriver/46/console

Sorry for the long comment :)

README.md
```diff
@@ -11,9 +11,10 @@
 [![NPM version](http://img.shields.io/npm/v/terra-toolkit.svg)](https://www.npmjs.org/package/terra-toolkit)
 [![Build Status](https://travis-ci.org/cerner/terra-toolkit.svg?branch=master)](https://travis-ci.org/cerner/terra-toolkit)
 
-Terra Toolkit is a utility module used to facilitate independent development of Terra projects. This toolkit provides configuration and helpers needed for nightwatch testing to streamline development of npm packages. [terra-core][@terra-core] and [terra-clinical][@terra-clinical] are two example mono-repos which are utilizing utilities offered in this package.
+Terra Toolkit is a utility module used to facilitate independent development of Terra projects. This toolkit provides configuration and helpers needed for nightwatch and webdriver.io testing to streamline development of npm packages. [terra-core][@terra-core] and [terra-clinical][@terra-clinical] are two example mono-repos which are utilizing utilities offered in this package.
```
mjhenkes, line 14: Do we want to remove nightwatch from this description? Once this is checked in we should be recommending writing tests with webdriver.io instead right?

mhemesath: I'm not against that.. though wasn't sure if we wanted to do a slow transition to webdriver.io in some pilot projects before doing a full fledge switch.

src/wdio/services/SeleniumDockerService.js
mhemesath: no problem, makes sense

emilyrohrbough (approved):

ryanthemanuel (approved):

PR #21, opened 2017-11-07
Updated nightwatch config to use docker

mhemesath:
### Summary
Updated nightwatch to use dockerized selenium to keep consistency in our CI/local testing environments.

### Additional Details
Assuming consumers have docker, this change should be passive.

@cerner/terra

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,6 @@ language: node_js
 sudo: required
 jdk: oraclejdk8
 dist: trusty
-addons:
-  chrome: stable
-  apt:
-    packages:
-      - oracle-java8-set-default
 before_script:
   - npm run compile
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 sudo: required
-jdk: oraclejdk8
 dist: trusty
 before_script:
   - npm run compile
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Unreleased
 * Use path.join to allow for windows development
 * Removed sauce labs config and install java 8 in a different way for travis ci
 * Added webdriver.io testing utilities
+* Update nightwatch to use docker
 
 2.1.0 - (October 5, 2017)
 ------------------
--- a/README.md
+++ b/README.md
@@ -24,6 +24,9 @@ Terra Toolkit is a utility module used to facilitate independent development of 
 
 - Install with [npm](https://www.npmjs.com): `npm install terra-toolkit --save-dev`
 
+Terar toolkit uses docker to run selenium to ensure a consistent testing environment locally and in continuous integration build systems. To use nightwatch with terra toolkit you must install docker on your machine: https://www.docker.com/
+
+
 ## Webdriver.io Utility
 [Webdriver.io](http://webdriver.io/) is a framework for writing webdriver powered tests to validate functionality in browsers. Webdriver.io also provides easy services for setting up selenium, starting webpack and static servers, as well as accessibilty and visual regression testing.
 
@@ -36,9 +39,8 @@ To make testing easier, Terra toolkit provides default configuration that enable
 * Mocha test framework
 
 ### Setup
-1. Install docker on your machine: https://www.docker.com/
 
-2. In your root directory, create a `wdio.conf.js` file that inherits from Terra Toolkit's base config.
+1. In your root directory, create a `wdio.conf.js` file that inherits from Terra Toolkit's base config.
 
 ```js
 const wdioConf = require('terra-toolkit/wdio/conf');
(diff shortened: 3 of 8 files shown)


.travis.yml
```diff
@@ -2,11 +2,6 @@ language: node_js
 sudo: required
 jdk: oraclejdk8
```
mjhenkes: can this be removed to?

mhemesath:
It seems that each test runs about ~100ms slower compared to previous builds and about 60ms slower locally.

.travis.yml
mhemesath: Probably

src/nightwatch/nightwatch.config.js
```diff
@@ -1,27 +1,46 @@
...
+      host: seleniumHost,
+      port: seleniumPort,
+      path: '/wd/hub',
+      seleniumDocker: {
+        enabled: !process.env.TRAVIS && !process.env.CI,
+        cleanup: true,
+      },
+    }, [{ browserName: 'chrome' }]);
+
+    Promise.all([staticServerPromise, dockerPromise]).then(() => done());
```
ryanthemanuel: I think you could just pass in done to then.  What happens if either of the promises error out?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Unreleased
 * Use path.join to allow for windows development
 * Removed sauce labs config and install java 8 in a different way for travis ci
 * Added webdriver.io testing utilities
+* Update nightwatch to use docker
```
ryanthemanuel, line 10: Is this technically non-passive?  Assuming that users of this have to have docker installed?  I guess it would only be non-passive to local developers?  Or consumers that are in travis that don't have this container set up (which is probably 0).

src/nightwatch/nightwatch.config.js
```diff
@@ -1,27 +1,46 @@
...
   }
 
+  const seleniumDocker = new SeleniumDockerService();
   const webpackServer = new WebpackDevServer(webpack(webpackConfig), { quiet: true, hot: false, inline: false });
 
   const startDriverAndServer = (done) => {
-    chromedriver.start();
-    webpackServer.listen(port, '0.0.0.0', () => done());
+    const staticServerPromise = new Promise((resolve) => {
+      webpackServer.listen(port, '0.0.0.0', () => resolve());
```
ryanthemanuel: Same here.  Could pass in resolve.

src/nightwatch/nightwatch.config.js
ryanthemanuel: +1 to starting them in parallel.

CHANGELOG.md
mhemesath: its non-passive from an environment standpoint but passive from a development one (after they install docker)

ryanthemanuel: Should we add that to the ReadMe?

mhemesath: meh, they'll figure it out... .j/k

README.md
```diff
@@ -263,6 +263,9 @@ expect(screenshots).to.not.matchReference();
 ## Nightwatch Utility
 Nightwatch.js is an easy to use Node.js based End-to-End (E2E) testing solution for browser based apps and websites. It uses the powerful W3C WebDriver API to perform commands and assertions on DOM elements. Full documentation regarding nightwatch can be found at http://nightwatchjs.org/.
 
+### Setup
```
emilyrohrbough: Given both nightwatch & wdio will need docker set up, it might be better to add the docker requirement under the Getting Started Section

emilyrohrbough (approved):

mhemesath:
Ran against Terra core and seems to run fine

ryanthemanuel (approved):

mhemesath:
I'm actually seeing one error on terra core when running against this branch
```
 ✖ nightwatch/date-time-picker-spec

   - [@tiny] Displays the Time Clarification dialog (1.872s)
   Timed out while waiting for element <[class*="time-clarification"]> to be present for 1000 milliseconds.  - expected "visible" but got: "not found"
       at Object.Displays the Time Clarification dialog (/Users/mh014308/Projects/terra-core/packages/terra-date-time-picker/tests/nightwatch/date-time-picker-spec.js:65:15)
       at _combinedTickCallback (internal/process/next_tick.js:131:7)
       at process._tickCallback (internal/process/next_tick.js:180:9)
   SKIPPED:
```

Is this concerning to anyone? Should we up the timeout for this test?

emilyrohrbough:
So the date-time picker has a few tests that depend on a [specific time zone](https://github.com/cerner/terra-core/blob/master/.travis.yml#L12), so we will need a way to specify that within the docker.

mhemesath:
@emilyrohrbough added the ability to set docker environment variables in the `wdio.conf.js`. This enables each app to control it how they want. In Nightwatch it defaults to `America/Chicago`.

package.json
```diff
@@ -50,7 +50,7 @@
     "glob": "^7.1.1",
     "ip": "^1.1.5",
     "load-json-file": "^2.0.0",
-    "webdriverio": "^4.8.0",
+    "webdriverio": "4.8.0",
```
bjankord, line 53: What's the logic around needing to lock this version down?

mhemesath: The latest release has a defect which breaks the visual regression testing service we're using. Theres a defect logged and its being looked into

bjankord: Ah ok, that makes sense.

PR #23, opened 2017-11-14
Improve docker selenium service stability

mhemesath:
Slight refactor of docker service internals to make it more stable:
* Dont check selenium status until image pulled and started. Prevents selenium timeouts while waiting for image to download
* Use container meta information to manage running containers over cidfile. Prevents corrupted container ids from lingering when test runs are interrupted.

### Summary
This pull request fixes a few stability issues with using selenium as a container. This should eliminate timeouts caused by pulling the container image for the firsttime as well as corrupted cidfiles from terminated test runs that don't clean up the container as that is all managed by meta information we're able to pull from the containers at runtime. 


wdio.conf.js
```diff
@@ -16,11 +16,11 @@ const config = {
...
   seleniumDocker: {
     enabled: !process.env.TRAVIS,
-    cleanup: true,
+    cleanup: false,
   },
   ...wdioConf.config,
 };
 
 
-config.services = wdioConf.config.services.concat(['static-server', SeleniumDockerService]);
```
mhemesath, line 25: this wasn't needed, seleniumdocker is packaged in the master config

ryanthemanuel (approved):

src/wdio/services/SeleniumDockerService.js
```diff
@@ -65,39 +31,125 @@ export default class SeleniumDockerService {
     this.port = config.port;
     this.path = config.path;
     this.browserName = capabilities[0].browserName;
-    this.cidfile = this.config.cidfile || this.cidfile;
+    this.container = await this.getRunningContainer();
```
ryanthemanuel: Do we even need to do this logic if we're disabled?

mhemesath: no, we can exit early

src/wdio/services/SeleniumDockerService.js
```diff
@@ -65,39 +31,130 @@ export default class SeleniumDockerService {
...
-
-        // Workaround for browser specific docker issues
-        // see https://github.com/SeleniumHQ/docker-selenium#running-the-images
-        let args = '';
-        if (this.browserName === 'chrome') {
-          args = '-v /dev/shm:/dev/shm';
-        } else if (this.browserName === 'firefox') {
-          args = '--shm-size 2g';
+      console.log('Ensuring selenium status is ready');
+      // Retry for 500 times up to 5 seconds for selenium to start
```
emilyrohrbough: Update inline to 10 seconds

mhemesath: i should probably just remove that comment.. but currenlty it only waits 5 seconds, do you want that upped to 10? And if its upped to 10, do you want more checks or a longer wait between checks?

package.json
```diff
@@ -50,10 +50,10 @@
     "glob": "^7.1.1",
     "ip": "^1.1.5",
     "load-json-file": "^2.0.0",
-    "webdriverio": "^4.9.8",
     "wdio-mocha-framework": "^0.5.11",
+    "wdio-static-server-service": "^1.0.0",
```
emilyrohrbough: I feel like `wdio-static-server-service` should be a devDependency

mhemesath: you're right, it should be

src/wdio/services/SeleniumDockerService.js
```diff
@@ -65,39 +31,130 @@ export default class SeleniumDockerService {
...
+      await this.runImage();
+      await this.ensureSelenium();
+    }
+  }
+
+  /**
+   * Clean up docker container after all workers got shut down and the process is about to exit.
+   */
+  async onComplete() {
+    if (this.config.cleanup && this.config.enabled) {
```
emilyrohrbough, line 68: Do we need to check `this.config.enbabled`? above it appears that if not enabled its not kicking it off.

mhemesath: the onComplete will still be called and we don't want to run the docker command if its not enabled.

src/wdio/services/SeleniumDockerService.js
```diff
@@ -65,39 +31,130 @@ export default class SeleniumDockerService {
...
+          } else {
+            resolve(result);
+          }
+        },
+      );
+    });
+  }
+
+  /**
+  * Pulls the configured docker image
```
emilyrohrbough: nit: update to indicate it accounts for browser and environment

wdio.conf.js
```diff
@@ -16,11 +15,11 @@ const config = {
 
   seleniumDocker: {
     enabled: !process.env.TRAVIS,
-    cleanup: true,
+    cleanup: false,
```
emilyrohrbough, line 18: Could just remove this since default is false

mhemesath: probably could.. i have there more for a documentation of default settings

emilyrohrbough:
Everything looks pretty good overall. Only thing I am wondering is if it would be more helpful if we offered more descriptive error messages in any failed promise.

mhemesath:
> if it would be more helpful if we offered more descriptive error messages in any failed promise

It will spit out what the shell will... I can try to wrap it but I feel the raw shell output may be the most beneficial

src/wdio/services/SeleniumDockerService.js
emilyrohrbough: The default retry interval above is set to 10 with retries at 500, but maybe just update the inline to indicate its waiting for selenium to start. 

I don't have an opinion on the wait/check times. I'm not sure what would be ideal, though upping it wouldn't necessarily hurt if its completes before the timeout & just moves on.

package.json
emilyrohrbough: Also side note, do we want to also list webdriverio, wdio-mocha-framework & wdio-visual-regression-service as both devDependencies and peerDependencies so consumers can set their version?

emilyrohrbough:
> I feel the raw shell output may be the most beneficial

Alright, wasn't sure how descriptive the shell output was. If you think it'd cover the basics of where failure occurs I think it's fine. Not need to bloat the console.

emilyrohrbough (approved):

PR #26, opened 2017-11-17
Update SeleniumDockerService to create a hub

mhemesath:
### Summary
This updates docker to use a docker-compose file so that we can start up a hub. Launching a hub provides a few benefits:
* Launching tests in parallel (recommended parallel instances to match available CPUs)
* Launching multiple browsers
* Launching multiple browsers with different configurations (e.g. different languages)

This change should be passive to toolkit consumers. It is non-passive to webdriverio consumers depending on their configuration (see removed options).


src/wdio/selenium-grid.yml
```diff
@@ -0,0 +1,21 @@
...
+    environment:
+      - GRID_THROW_ON_CAPABILITY_NOT_PRESENT=false
+      - GRID_MAX_SESSION=10
+      - GRID_BROWSER_TIMEOUT=100000
+      - GRID_TIMEOUT=90000
+      - GRID_NEW_SESSION_WAIT_TIMEOUT=300000
+    ports:
+      - '4444:4444'
+  chrome:
+    image: 'selenium/node-chrome:latest'
```
alex-bezek: Do you want latest to make sure you aggressively update to not fall behind? If you wanted to be more stable, avoid random failures, and make sure people always have the same version, you could lock it to a specific tag
https://hub.docker.com/r/selenium/node-chrome/tags/

src/wdio/services/SeleniumDockerService.js
```diff
@@ -90,104 +80,41 @@ export default class SeleniumDockerService {
...
-  runImage() {
-    let args = '';
-    if (this.browserName === 'chrome') {
-      args = '-v /dev/shm:/dev/shm';
-    } else if (this.browserName === 'firefox') {
-      args = '--shm-size 2g';
-    }
+  startHub() {
+    const scale = Object.keys(this.config.instances).map(key => `--scale ${key}=${this.config.instances[key]}`).join(' ');
+    const command = `docker-compose -f ${this.config.composeFile} up -d ${scale}`;
```
alex-bezek: https://docs.docker.com/compose/extends/#example-use-case
if you wanted, instead of them having to add their own compose file, which should allow them to override existing environment variables or configs in your file, or just tack on additional browsers

src/wdio/services/SeleniumDockerService.js
```diff
@@ -17,46 +23,30 @@ export default class SeleniumDockerService {
    * @param {Object} config wdio configuration object
    * @param {Array.<Object>} capabilities list of capabilities details
    */
-  async onPrepare(config, capabilities) {
+  async onPrepare(config) {
+    // If a custom compose file is provided, dont' default instances as we don't
+    // know the service names.
```
alex-bezek: you can get the service names via
`docker-compose config --services`
https://docs.docker.com/compose/reference/config/

README.md
```diff
@@ -66,12 +66,11 @@ Selenium docker is provided as a convenience to make selenium testing easier and
...
 Under the key `seleniumDocker` in your wdio.conf.js you can pass a configuration object with the following structure:
 
+* **composeFile** - The docker compose file to use to standup the hub. Defaults to a compose file with chrome instances on port 4444.
 * **enabled** - Flag to disable selenium docker; useful for CI environments which can startup the docker instance outside of test runs. Defaults to true.
-* **cleanup** - Destroy the docker container after the test run. Defaults to false.
-* **image** - The docker image to use for test runs. Defaults to `selenium/standalone-chrome` or `selenium/standalone-firefox` based on browser capabilities specified in config.
 * **retries** - Retry count to test for selenium being up. Default 1000.
 * **retryInterval** - Retry interval in milliseconds to wait between retries for selenium to come up. Default 10.
-* **env** - An Object representing environment variables to set within the docker instance. Defaults to {}
+* **instances** - A map of services in the docker compose file to scale up. Defaults to scaling chrome and nodes to the count of CPUs on the host machine.
```
alex-bezek: You could force them to provide a config file that sets the scale if you wanted
https://docs.docker.com/compose/compose-file/compose-file-v2/#scale
i guess it depends on the more common use case

src/wdio/services/SeleniumDockerService.js
ryanthemanuel: From looking at that link it looks like extends got removed in docker-compose v3.

alex-bezek: hmmm, while it is true it says that using the keyword extends from your dockerfile is no longer supported, I wonder if maybe using overrides in the case of specifying multiple files still works. Mike's current default compose file is version 2, but this brings up a good question of if versions overriding/extending are compatible, or if this should just keep using a single compose file to not worry about versions

mhemesath: So I think I'm going to go with the approach of providing a sane default (firefox/chrome) and letting consumers of the service pass in their own config if they want/need. This is the simplest and most customizable approach IMO.

README.md
mhemesath: Decided to go with a defualt compose file and allow consumers to override if they want.

README.md
```diff
@@ -66,12 +66,11 @@ Selenium docker is provided as a convenience to make selenium testing easier and
 
 Under the key `seleniumDocker` in the wdio.conf.js you can pass a configuration object with the following structure:
 
+* **composeFile** - The docker compose file to use to standup the hub. Defaults to a compose file with 5 chrome instances on port http://localhost:4444/wd/hub.
 * **enabled** - Flag to disable selenium docker; useful for CI environments which can startup the docker instance outside of test runs. Defaults to true.
-* **cleanup** - Destroy the docker container after the test run. Defaults to false.
```
mhemesath: removed this option because docker highly recommends removing containers after every test run

src/wdio/docker-compose.yml
```diff
@@ -0,0 +1,26 @@
...
+      - /dev/shm:/dev/shm
+    ports:
+      - '5555:5555'
+    deploy:
+      replicas: 5
+    environment:
+      HUB_PORT_4444_TCP_ADDR: selenium_hub
+      HUB_PORT_4444_TCP_PORT: 4444
+      NODE_MAX_SESSION: 1
+      TZ: 'America/Chicago'
```
mhemesath, line 26: Including america/chicago timezone by default. Once we move this out into a standalone project, we'll have toolkit provide its own compose file which will include this setting for terra test runs

mhemesath:
Tested this locally against terra-core and the tests pass. @emilyrohrbough can you give it a try to make sure the docker config works for your machine? Thanks

README.md
emilyrohrbough: +1 to default. The more 'out-of-the-box' I think the better with options to do what they want with it.

emilyrohrbough:
@mhemesath I pulled this down, ran clean installed & npm run test. This is the error/hang up I am seeing.

screen shot 2017-11-22 at 9 10 29 am

mhemesath:
@emilyrohrbough I think I found the cause. The docker stack can take a second or two to come down and if another stack is brought up in that time there is a conflict trying to create a network. This can happen in a few scenarios

* Executing nightwatch then wdio tests in tandem 
* Killing a wdio/nightwatch test run mid-way, then restarting it which triggers a stack cleanup. 

I think I found a fix which is to wait until the network is removed before rebuilding the stack. Testing it locally now

alex-bezek (approved):

bjankord (approved):

mhemesath:
@ryanthemanuel  are you good to merge this?

ryanthemanuel (approved):

ryanthemanuel:
@mhemesath yup good to go.

wdio.conf.js
```diff
@@ -20,6 +20,9 @@ const config = {
   seleniumDocker: {
     enabled: !process.env.TRAVIS,
     cleanup: false,
+    env: {
```
emilyrohrbough: Do we still need this config if its in the compose file?

emilyrohrbough (approved):

PR #28, opened 2017-11-27
Toolkit Docs

emilyrohrbough:
### Summary
This PR breaks up the Toolkit README by providing a Nightwatch and Webdriver.io Utility Guide.
These both have the following sections:
- Getting Started
- Configuration Setup 
- Writing Tests
- Running Tests

Additionally broke each Terra Wedriver.io service into their own doc files that specifies the config options with examples.

Most of this is 1-to-1 or similar to the documentation before. Just make the docs feel clean and easier to read by breaking it up. 

###  Anything unclear please indicate!

@cerner/terra

PR #29, opened 2017-11-28
Added test helpers

mhemesath:
### Summary
This adds test helpers for accessibility and screenshot comparison. If you think the syntax looks good, I'll write up doc and add it to the PR.


tests/wdio/resize-spec.js
```diff
@@ -0,0 +1,21 @@
+/* global browser, describe, it, expect, before, Terra */
+describe('resize utility', () => {
+  before(() => browser.url('/compare.html'));
+
+  it('resizes to tiny and huge', () => {
+    Terra.resize('tiny', 'huge', (name, viewport) => {
```
emilyrohrbough: Yesterday you said that webdriver.io commands have to exist within an it() or before/after() hook so we can't wrap a bunch of it() calls in a resize helper. And then looking at this test and the resize implementation, I am wondering if we are providing much value by providing this helper. Webdriver's `setViewportSize()` is simple to implement and `should.beAccessible()` and `should.beComparable()` accept an array of viewport sizes.

src/wdio/services/TerraService.js
```diff
@@ -37,21 +46,41 @@ function matchReference() {
...
+};
+
+const should = {
+  beAccessible(options) {
+    global.it('is accessible', () => {
+      global.expect(global.browser.axe(options)).to.be.accessible();
+    });
+  },
+
+  beComparable({ name = 'default', css = '[data-reactroot]', viewports = {} }) {
```
emilyrohrbough: A few things:
- IMO `matchScreenshot` or `matchSnapshot` better describes this helper. `beComparable` could be interpreted differently and I feel like we want an explicit name for a test helper
- How would you feel about breaking name and css out of the param obj? name is specific to our helper and then css and options would be more consistent with visual regression's checkElement func allowing more flexibility to pass in more than viewports
- I feel like the term `element` or `selector` might be clearer than `css`

src/wdio/services/TerraService.js
```diff
@@ -37,21 +46,41 @@ function matchReference() {
...
+    global.it(`[${name}] checks visual comparison`, () => {
```
emilyrohrbough: To go along with my first bullet, maybe `[${name}] matches snapshot`.

src/wdio/services/TerraService.js
```diff
@@ -64,7 +93,9 @@ export default class TerraService {
   before() {
     global.expect = chai.expect;
     global.Terra = {
-      viewports,
+      viewports: getViewports,
+      resize,
+      should,
```
emilyrohrbough: This is a total NIT, but it might be easier to reference the should helpers if defined like:
```
global.Terra = {
    should: {
          beAccessible,
          beComparable,
    }
}
```
I'm thinking in terms of consumers checking which helpers are available and if we were to add more helpers in the future.

tests/wdio/resize-spec.js
mhemesath: yeah, i'm fine if we want to get rid of this helper. It was a worthy experiment to try it out

src/wdio/services/TerraService.js
mhemesath: > How would you feel about breaking name and css out of the param obj? name is specific to our helper and then css and options would be more consistent with visual regression's checkElement func allowing more flexibility to pass in more than viewports

Yeah I had that at first but the problem then becomes if you want to change the viewports but not change the name or default element....  you are now forced to reset them. Thoughts?

tests/wdio/resize-spec.js
emilyrohrbough: If we could do multiple nesting of its() I think it could be valuable. Good to know this concept works if needed later.

src/wdio/services/TerraService.js
emilyrohrbough: hmm what about something like:
```
beComparable(name = 'default', { css = '[data-reactroot]', options  = { viewports = {} } })
```
or
```
beComparable(name = 'default', { css = '[data-reactroot]',  viewports = {}, options  = {})
// then merge options & viewports
```
side note: Should we default to `viewports = VIEWPORTS`?

mhemesath:
```
beComparable(name = 'default', { css = '[data-reactroot]',  viewports = {}, options  = {})
beComparable(name = 'default')
beComparable({ css = '[data-reactroot]',  viewports = {}, options  = {})
```

I wonder how hacky it is to support all 3? Rules would be as follows:

1. 1st param can be object or String. If string then its name, else its options
2. 2nd param, if present is always assumed to be options.

Thoughts?

ryanthemanuel:
.DS_Store shouldn't be checked in.

ryanthemanuel:
Let's update the ReadMe with these new helpers

mhemesath:
Yeah I’m going to update readme once we all agree on API for helpers

src/wdio/services/TerraService.js
```diff
@@ -37,21 +47,37 @@ function matchReference() {
...
+    options = param2 || options;
+  } else {
+    options = param1 || options;
   }
 
-  return sizes.map(size => widths[size]);
+  const selector = options.selector || '[data-reactroot]';
+  const viewports = options.viewports || [];
+
+  global.it(`[${name}] matches screenshot`, () => {
```
emilyrohrbough, line 77: Is there a way to grab the test step name from this instance? that was how we originally proposed the naming structure
https://github.com/cerner/terra-toolkit/blob/master/tests/wdio/compare-spec.js#L7

mhemesath: This is the test step.... this is a helper to inject it so it needs a name passed in

mhemesath: This is the test step.... this is a helper to inject it so it needs a name passed in

emilyrohrbough: duh right. I was thinking in terms grabbing the describe for a test doing something like:

[1 / default] matches screenshot
[1 / default] displays modal currently on click
[1 / default] displays modal on exit

If not, I think we might want to change param1 to be test_step which does not have a default message.

emilyrohrbough: @ryanthemanuel Is this how you'd expected Orion testing to output for screenshots? Part of the reason for the brackets was for requirements mapping but this only gives one test description as is

ryanthemanuel: Let me sync up with some folks on the Orion side to see if there's anything specific they're expecting.  I think we could still encapsulate the requirements inside of the name being passed in if necessary.

ryanthemanuel:
```
Terra.should.matchScreenshot('345626, 3452534 Test that button responds accordingly', { selector: 'button', viewports });
```

ryanthemanuel:
I am good with these changes at this point (with the ReadMe updates of course :))

mhemesath:
Once #28 is merged, I'll update docs for this PR.

bjankord (changes requested):
Overall looks good to me, I'd remove the `.DS_Store` file and either add that to repo's gitignore file or add that to your personal global gitignore

JakeLaCombe (approved):

docs/Wdio_Utility.md
```diff
@@ -85,12 +85,24 @@ Then, to assist with testing, the TerraService provides the custom assertions `a
 - `Terra.viewports(name)` takes the viewport key name(s) and returns an array of { height, width } objects representing the respective terra viewport size(s).
     - Use this function to resize the browser or pass the viewport sizes to the accessibility and visual regression commands.
     - By default returns all viewports if not name key are provided.
-
+- `Terra.should.beAccessible()` convenience method that injects an axe test. Takes the same arguments as the `axe()` utility. See [beAccessible-spec.js](https://github.com/cerner/terra-toolkit/blob/master/tests/wdio/beAccessible-spec.md) for examples.
+- `Terra.should.matchScreenshot(name, options)` convenience method that injects a screenshot test. See [matchScreenshot-spec.js](https://github.com/cerner/terra-toolkit/blob/master/tests/wdio/matchScreenshot-spec.md) for example usage..
```
emilyrohrbough: oops. extra period here

docs/Wdio_Utility.md
```diff
@@ -110,9 +122,34 @@ describe('Basic Test', () => {
       browser.setViewportSize(size);
     });
   });
+
+  Terra.should.beAccessible();
 });
 ```
 
+### Testing multiple viewports.
```
emilyrohrbough: Ah! This is a good one to include. 😄

docs/Wdio_Utility.md
```diff
@@ -85,12 +85,24 @@ Then, to assist with testing, the TerraService provides the custom assertions `a
 - `Terra.viewports(name)` takes the viewport key name(s) and returns an array of { height, width } objects representing the respective terra viewport size(s).
     - Use this function to resize the browser or pass the viewport sizes to the accessibility and visual regression commands.
     - By default returns all viewports if not name key are provided.
-
+- `Terra.should.beAccessible()` convenience method that injects an axe test. Takes the same arguments as the `axe()` utility. See [beAccessible-spec.js](https://github.com/cerner/terra-toolkit/blob/master/tests/wdio/beAccessible-spec.md) for examples.
```
emilyrohrbough: convenience method that utilizes the accessible()  assertion

docs/Wdio_Utility.md
emilyrohrbough: convenience method that utilizes the matchReference() assertion

docs/Wdio_Utility.md
```diff
@@ -85,12 +85,24 @@ Then, to assist with testing, the TerraService provides the custom assertions `a
...
+
+  Terra.should.beAccessible();
+  Terra.should.matchScreenshot();
+
+  it('custom test', () => {
+    expect('something').to.equal('something');
+  });
+});
+
+describe('Advanced Test', () => {
```
emilyrohrbough: Wondering if we give them the information on the Terra object then give the basic example, then explain the accessible() and matchReference() assertions & give the advance to break it up and emphasis the Terra global?

Otherwise looks good.

docs/Wdio_Utility.md
```diff
@@ -110,9 +122,34 @@ describe('Basic Test', () => {
       browser.setViewportSize(size);
     });
   });
+
+  Terra.should.beAccessible();
```
emilyrohrbough: Above in this example we have 

```
   it('checks accessibility', () => {
     expect(browser.axe()).to.be.accessible();
   });
```

prob don't need to add `Terra.should.beAccessible();`

mhemesath: yup removed

docs/Wdio_Utility.md
mhemesath: 👍  updated

docs/Wdio_Utility.md
```diff
@@ -78,19 +78,46 @@ There are a few things to note about the webdriver.io configuration provided by
 - Test files should use `*-spec.js` naming format. The default spec search path is `./tests/wdio/**/*-spec.js`.
 - Use `/test_url_path` to direct test urls. This is appended to the `baseUrl` provided in the config.
 
-Then, to assist with testing, the TerraService provides the custom assertions `accessible` and `matchReference` and a Terra global helper to make testing easier:
 
-- `accessible()` validates the `axe()` accessibility assertions on the specified viewports are successful.
-- `matchReference()` validates the `checkElement` visual regression assertions on the specified viewports are successful.
+
+Then, to assist with testing, the TerraService provides custom methods to make testing easier:
```
emilyrohrbough: Then, to assist with testing, the TerraService provides the Terra global helper to make testing easier:

docs/Wdio_Utility.md
```diff
@@ -78,19 +78,46 @@ There are a few things to note about the webdriver.io configuration provided by
...
+  Terra.should.matchScreenshot({ viewports });
+
+  it('custom test', () => {
+    expect('something').to.equal('something');
+  });
+});
+```
+
+
+If more control is needed over the assertions, the TerraService provides the custom assertions `accessible` and `matchReference` and a Terra global helper to make testing easier:
```
emilyrohrbough: If more control is needed over the assertions, the TerraService alos provides the custom assertions `accessible` and `matchReference`

emilyrohrbough (approved):

bjankord (approved):

PR #37, opened 2018-01-03
Add support for screenshot subtitles.

hs025828:
### The Problem
If a single wdio test captures multiple screenshots, then all screenshots following the first will be compared against the first because they all share the same file name. 

### A Proposed Solution
This enhancement allows consumers to subtitle the file names of their screenshots to ensure uniqueness.

I have implemented this as an optional feature for consumers.  Here is a usage example:

```
let screenshots = browser.checkViewport({ viewports, subtitle: 'this is my subtitle' });
expect(screenshots).to.matchReference();
```

@cerner/terra

### DIFF
--- a/src/wdio/visualRegressionConf.js
+++ b/src/wdio/visualRegressionConf.js
@@ -3,15 +3,19 @@ import { LocalCompare } from 'wdio-visual-regression-service/compare';
 
 const testIdRegex = /\[([^)]+)\]/;
 
-function testName(parent, title) {
+function testName(parent, title, subtitle) {
   const matches = testIdRegex.exec(title);
   const parentName = parent.replace(/[\s+.]/g, '_');
   let name = title.trim().replace(/[\s+.]/g, '_');
   if (matches) {
     name = matches[1];
   }
+  let subtitleName = '';
+  if (subtitle) {
+    subtitleName = `[${subtitle.trim().replace(/[\s+.]/g, '_')}]`;
+  }
 
-  return `${parentName}[${name}]`;
+  return `${parentName}[${name}]${subtitleName}`;
 }
 
 function getScreenshotName(ref) {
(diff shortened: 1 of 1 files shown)


emilyrohrbough:
> If a single wdio test captures multiple screenshots, then all screenshots following the first will be compared against the first because they all share the same file name.

We provide a screenshot helper function which can be called `Terra.should.matchScreenshot()` which allows you to pass a custom name to your screenshot. 

Using this helper, writing a test would result in 

```
const viewports = Terra.viewports('tiny');

describe('Menu', () => {
       before(() => {
        browser.url('/#/tests/menu-tests/default');
      });
  
      Terra.should.matchScreenshot({ viewports });
      // resulting filename: Menu[default].470x768.png

      it('opens menu, () => {
        browser.click('#trigger-menu-button');
      });
      Terra.should.matchScreenshot('opens',  { viewports });
      // resulting filename: Menu[opens].470x768.png
```

Does this meet your needs?

hs025828:
`Terra.should.matchScreenshot` will not generate a screenshot if it's called from within an `it`.  Is this intentional?

emilyrohrbough:
`Terra.should.matchScreenshot` is a custom [`it` statement](https://github.com/cerner/terra-toolkit/blob/master/src/wdio/services/TerraService.js#L91). We are still in the process of determining what we need for wdio helper, so there are ideas on how we can make this service better. One of the possible future options is to add a `matchScreenshot` helper that is not a custom `it`.

hs025828:
For my immediate use case I'm fine with splitting up my `it` and capturing screenshots in between.  Since that solves my problem my issue can be closed.

emilyrohrbough:
Sounds good! While you work through using webdriver.io, if you find a helper that could be useful in the toolkit, feel free to log an issue as a potential future enhancement! 😄

PR #48, opened 2018-01-22
Add the ability to add custom selectors to wdio themeable tests

JakeLaCombe:
### Summary
React 16 removed the [data-reactroot] (see https://github.com/facebook/react/issues/10971) element we defined for looking for react components. As a result, we need to add the ability to specify a custom selector that can be passed into the themeable test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changelog
 Unreleased
 ----------
 * Removed React dependency
+* Add options argument to themeEachCustomProperty method
 
 2.6.0 - (January 4, 2018)
 ----------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ Changelog
 Unreleased
 ----------
 * Removed React dependency
-* Add options argument to themeEachCustomProperty method
+* Add selector argument to themeEachCustomProperty method
+* Add default document global selector for Terra service to wdio config.
 
 2.6.0 - (January 4, 2018)
 ----------
--- a/src/wdio/services/TerraService.js
+++ b/src/wdio/services/TerraService.js
@@ -65,12 +65,13 @@ const beAccessible = (options) => {
 /**
 * Generates a test for each themed property given and runs a screenshot assertion.
 * @property {Object} customProperties - An object containing the CSS custom properties to assert.
+* @property {Object} options - Additional options for testing.
 */
-const themeEachCustomProperty = (customProperties) => {
+const themeEachCustomProperty = (customProperties, options = {}) => {
   Object.entries(customProperties).forEach(([key, value]) => {
     global.it(`themed [${key}]`, () => {
       global.browser.execute(`document.documentElement.style.setProperty('${key}', '${value}')`);
-      global.expect(global.browser.checkElement('[data-reactroot]')).to.matchReference();
+      global.expect(global.browser.checkElement(options.selector || '[data-reactroot]')).to.matchReference();
     });
   });
 };
--- a/src/wdio/services/TerraService.js
+++ b/src/wdio/services/TerraService.js
@@ -67,11 +67,21 @@ const beAccessible = (options) => {
 * @property {Object} customProperties - An object containing the CSS custom properties to assert.
 * @property {Object} options - Additional options for testing.
 */
-const themeEachCustomProperty = (customProperties, options = {}) => {
-  Object.entries(customProperties).forEach(([key, value]) => {
+const themeEachCustomProperty = (customSelector, customProperties) => {
+  let selector = global.browser.options.terra.selector;
+  let styleProperties = {};
+
+  if (typeof customSelector === 'string') {
+    selector = customSelector;
+    styleProperties = customProperties || styleProperties;
+  } else {
+    styleProperties = customSelector || styleProperties;
+  }
+
+  Object.entries(styleProperties).forEach(([key, value]) => {
     global.it(`themed [${key}]`, () => {
       global.browser.execute(`document.documentElement.style.setProperty('${key}', '${value}')`);
-      global.expect(global.browser.checkElement(options.selector || '[data-reactroot]')).to.matchReference();
+      global.expect(global.browser.checkElement(selector)).to.matchReference();
     });
   });
 };
@@ -79,6 +89,7 @@ const themeEachCustomProperty = (customProperties, options = {}) => {
 const matchScreenshot = (param1, param2) => {
   let name = 'default';
   let options = {};
+
   if (typeof param1 === 'string') {
     name = param1;
     options = param2 || options;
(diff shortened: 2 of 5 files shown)


bjankord (approved):

src/wdio/services/TerraService.js
```diff
@@ -65,12 +65,13 @@ const beAccessible = (options) => {
 /**
 * Generates a test for each themed property given and runs a screenshot assertion.
 * @property {Object} customProperties - An object containing the CSS custom properties to assert.
+* @property {Object} options - Additional options for testing.
 */
-const themeEachCustomProperty = (customProperties) => {
+const themeEachCustomProperty = (customProperties, options = {}) => {
```
emilyrohrbough: Is command is pretty straight forward. Do you thinking we need to allow an options object or could we directly pass the selector?

```
const shouldTheme = (customProperties, selector) => {
  const selectedElement = selector || '[data-reactroot]';
  Object.entries(customProperties).forEach(([key, value]) => {
    it(`themed [${key}]`, () => {
      browser.execute(`document.documentElement.style.setProperty('${key}', '${value}')`);
      expect(browser.checkElement(selectedElement)).to.matchReference();
    });
  });
};
```

JakeLaCombe: I kept it to options to add better passivity to it, and keep it consistent with the API the other functions have. If we ever had to add additional parameters to the method, that could make a potential non passive change, as you'd have to re arrange the arguments of the method.

JakeLaCombe:
I can confirm these changes to the api allow tests to pass in terra-core. Here is a snapshot of passing terra-badge wdio tests. 
screen shot 2018-01-22 at 3 04 44 pm

src/wdio/services/TerraService.js
mhemesath: It might also be nice to allow an option to be passed in `wdio.conf.js` to set a default selector globally. This will prevent users from having to set this option everywhere.

Look at the AxeService for an example on how to set/read global properties in wdio.

emilyrohrbough: So I think this looks just fine, though I honestly, I would be more comfortable siting down and mapping out a wishlist of expected functions to be added to wdio before adding enhancements to the current ones. 

For instance, allowing this command to take options seems like overkill here. This is a very specific command and an options object adds an additional layer of complexity to its use that I don't think is needed here. Plus it lesses the readability of the command. 

As per @mhemesath suggestion regarding a global option, I also think that would be valuable, but I'd again want to scope out the use-case and future commands we wish to add. 

I think there are a lot of commands we can add to be useful which would then need a similar setup to this if we make changes now without thinking them through. 

@bjankord @mjhenkes @mhemesath  Thoughts??

mhemesath: > This is a very specific command and an options object adds an additional layer of complexity to its use that I don't think is needed here

I tend to agree. I think we could get by with `Terra.should.themeEachCustomProperty('.my-selector', {})`; In this case its easy to check if the first param is a string, assume a custom selector is added, else use default.

JakeLaCombe: See if this works better. I do like the idea of the global setting. https://github.com/cerner/terra-toolkit/pull/48/commits/fadce63e19d9c69b398a1b8da014c3717a1f064b

src/wdio/services/TerraService.js
```diff
@@ -67,26 +67,37 @@ const beAccessible = (options) => {
 * @property {Object} customProperties - An object containing the CSS custom properties to assert.
 * @property {Object} options - Additional options for testing.
 */
-const themeEachCustomProperty = (customProperties, options = {}) => {
-  Object.entries(customProperties).forEach(([key, value]) => {
+const themeEachCustomProperty = (customSelector, customProperties) => {
+  let selector = global.browser.options.terra.selector;
+  let styleProperties = {};
```
mhemesath: you could technically make this always equal to the last argument
```js
// If more than 1 argument, selector is fist 
const selector = arguments.length > 1 ? arguments[0] : global.browser.options.terra.selector;
// Style properties are always last.
const styleProperties = arguments[arguments.length - 1];
```

mhemesath (approved):

src/wdio/services/TerraService.js
JakeLaCombe: When I run that with the wdio testing interface, I see that it's actually setting arguments to something else.

screen shot 2018-01-23 at 11 42 15 am

Do you know if there is away around that with the wdio command line? If not, we may not be be able to rely on the arguments object.

mhemesath: Sorry, arrow functions don't have arguments. Use rest params

```js
// Example
const x = (...args) => console.log(args)
```

JakeLaCombe: Forgot all about that myself! Thanks for the reminder! It works great, and even did a little extra cleanup too :) https://github.com/cerner/terra-toolkit/pull/48/commits/1ecb511eac01274dd31faa17e2eb9597c7665788

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Changelog
 Unreleased
 ----------
 * Removed React dependency
+* Add options argument to themeEachCustomProperty method
```
emilyrohrbough: Changelog needs updated to add both the addition of a global terra selector as well as the enhancement to  themeEachCustomProperty to allow for a selector

src/wdio/services/TerraService.js
```diff
@@ -64,28 +64,29 @@ const beAccessible = (options) => {
...
-    options = param1 || options;
-  }
+const matchScreenshot = (...args) => {
+  // If more than 1 argument, name is first
+  const name = args.length > 1 ? args[0] : 'default';
+  // test options are always last.
+  const options = args[args.length - 1] || {};
 
-  const selector = options.selector || '[data-reactroot]';
+  const selector = options.selector || global.browser.options.terra.selector;
```
emilyrohrbough, line 89: We need to add the selector key to the Terra object, with the default value of '[data-reactroot]'. We shouldn't need to add to our wdio.config file unless we'd want a default other than '[data-reactroot]'

wdio.conf.js
```diff
@@ -30,6 +30,9 @@ const config = {
       }],
     },
   },
+  terra: {
```
emilyrohrbough: This should be defaulted and no changes should be needed in this file. See comment above

JakeLaCombe: I was thinking this was the default configuration. If not, I can add it back to the service file.

emilyrohrbough: This file is the wdio.config for the Toolkit repo. Here is the [default config file](https://github.com/cerner/terra-toolkit/blob/1ecb511eac01274dd31faa17e2eb9597c7665788/src/wdio/conf.js) that we provide.

JakeLaCombe: See if this works for you. I added it to the default config file. https://github.com/cerner/terra-toolkit/pull/48/commits/09ee1978cfcf0f50f51ee1f85a3ed41308da06e9

emilyrohrbough (approved):

PR #56, opened 2018-02-13
Webdriver.io Visual Regression Enhancements

### LINKED ISSUE #51: matchReference assertion fails when the image is within the misMatchTolerance
# Issue Description
We have a screenshot that was 0.01% different than the reference screenshot (chrome changed how characters that went into the descender were underlined) so it passed the LocalCompare [1][2] check and no diff screenshot was generated. However, the matchReference [3] assertion failed because the images were not exactly the same. Instead of checking isExactSameImage [4], a check could be made against isWithinMisMatchTolerance [5].

[1] https://github.com/cerner/terra-toolkit/blob/v2.7.0/src/wdio/visualRegressionConf.js#L33
[2] https://github.com/zinserjan/wdio-visual-regression-service/blob/v0.8.0/src/methods/LocalCompare.js#L39
[3] https://github.com/cerner/terra-toolkit/blob/v2.7.0/src/wdio/services/TerraService.js#L40
[4] https://github.com/zinserjan/wdio-visual-regression-service/blob/v0.8.0/src/methods/BaseCompare.js#L39
[5] https://github.com/zinserjan/wdio-visual-regression-service/blob/v0.8.0/src/methods/BaseCompare.js#L37

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
The matchReference assertion should not fail for a screenshot that is within the misMatchTolerance.

## Current Behavior
The matchReference assertion fails.

### LINKED ISSUE #55: VR Fails Tests but Doesn't Render Diff
# Issue Description
At times, a screenshot comparison will fail, however a snapshot diff is not generated.

### What is happening?
In the Terra.matchesScreenshot() method, we use the visual regression service to generate and compare the screenshots. 

There are four types of information returned by the visual regression comparison:
1. `isExactMatch`
2. `isWithinMisMatchTolerance`
3. `isSameDimensions`
4. `misMatchPercentage`

Currently, we use our custom chai assertion to check if the generated screenshot `isExactMatch`. This should be updated to check if the image `isWithinMisMatchTolerance`. 

Our tests are failing because they are not exact, however it seems the screenshot diff is not generated because our screenshot is likely within this mismatch tolerance.

emilyrohrbough:
### Summary
Update file structure for readability and future test helpers. Additionally: 
#### Added
- Terra.matchScreenshotExactly mocha-chai helper to take screenshots and check for exact matches.
- Add global should reference to chai should
- Expanded Terra.matchScreenshot & Terra.matchScreenshotExactly to include the wdio-visual-regression-service caputure screenshot options of misMatchTolerance and viewportChangePause to allow for custom test step values.
#### Changed
- Enhanced the mocha-chai helper to be more descriptive and to iterate each viewport option for better test failure information
- For the chai assertion method matchReference
    - Enhanced error message to be more descriptive for fail assertion
    - Include screenshot compare results for fail assertion
    - Accepts matchType to make assertion on comparison results to be withinTolerance or to be the exact image. Defaults to 'withinTolerance'. Previous behavior only checked for exact image and did not account for globally defined misMatchTolerance value. 
        - Resolves #51. 
        - Resolves #55 - this issue was caused  by unexpected behavior by checking for an exact image (which failed) when the image was indeed within the mismatch tolerance. 

### Additional Details
More documentation is needed to document these new options.

@cerner/terra


src/wdio/services/TerraCommands/visual-regression.js
```diff
@@ -0,0 +1,140 @@
...
+* @property {String} matchType - Specifies the type of matchReference assertion. Either 'withinTolerance'
+*    or 'exactly'.
+*/
+const matchScreenshot = (testArguments, matchType) => {
+  const { name, selector, viewports, options } = determineScreenshotOptions(...testArguments);
+
+  const testDescription = getTestDescription(matchType);
+
+  if (viewports.length) {
+    // Create assertion for each viewport so know which screenshots-viewport combinations pass/fail comparison
```
mjhenkes: nit: grammer

mjhenkes (approved):

src/wdio/services/TerraCommands/chai-methods.js
```diff
@@ -0,0 +1,86 @@
...
+
+  let imagesMatch;
+  if (matchExactly) {
+    imagesMatch = screenshots.every(screenshot => (screenshot && screenshot.isSameDimensions && screenshot.isExactSameImage));
+  } else {
+    imagesMatch = screenshots.every(screenshot => (screenshot && screenshot.isSameDimensions && screenshot.isWithinMisMatchTolerance));
+  }
+
+  this.assert(imagesMatch === true,
+    `expected to ${testDescription}, but recieved the following comparison results \n${comparisonResults}`,
```
JakeLaCombe: *received

src/wdio/services/TerraCommands/visual-regression.js
```diff
@@ -0,0 +1,140 @@
...
+/**
+* Mocha-chai wrapper method to capture screenshots of a specified element and assert the
+* screenshot comparision results are an exact match.
+* @property {Array} args - The list of test arguments to parse. Accepted Arguments:
+*    - String (optional): the test case name. Default name is 'default'
+*    - Object (optional): the test options. Options include selector, viewports,
+*        misMatchTolerance and viewportChangePause.
+*     Note: args list order should be: name, then options when using both.
+*/
+const matchScreenshotExactly = (...args) => {
```
JakeLaCombe: This method does scare me. We have had issues with being too pixel perfect with projects I've worked on for other teams, and it makes me a little concerned this method could be used way too much. Should we have some kind of documentation saying this should rarely be used?

emilyrohrbough: I'm not sure to be honest. @mjhenkes thoughts?

From a standalone component standpoint, it makes sense to want a closer to perfect image, but IMO from an application standpoint consumer terra-components, it would be best to use `matchScreenshot()` method given we might a small style change that they wouldn't need to be concerned with.

All of our test will be using `matchScreenshot()`, which I imagine most people will be referencing when starting to write tests.

JakeLaCombe (approved):

Issue #36, opened 2017-12-15
Fallback to webpack.config.js devServer entry if config.webpackDevServerConfig is not provided


mschile:
# Issue Description
The [WebpackDevService](https://github.com/cerner/terra-toolkit/blob/v2.4.0/src/wdio/services/WebpackDevServerService.js#L14) does not support the webpack.config.js [devServer](https://webpack.github.io/docs/configuration.html#devserver) entry.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The webpack.config.js devServer entry is used as a fallback if config.webpackDevServerConfig is not populated.

## Current Behavior
The webpack.config.js devServer entry is ignored.

emilyrohrbough:
Conversation held on https://github.com/cerner/terra-site/pull/1 which may link.
<img width="671" alt="screen shot 2018-01-30 at 8 35 26 am" src="https://user-images.githubusercontent.com/14099737/35571751-9fd2063c-0598-11e8-9f5d-54be39924d41.png">
Link in comment https://webpack.js.org/configuration/dev-server/#devserver

emilyrohrbough:
The WebpackDevServerService is being deprecated in #81

noahbenham:
Since #81 is merged should this close @emilyrohrbough?

PR #68, opened 2018-03-09
Add a new Express Dev server service.

mjhenkes:
### Summary
Add a new simple express dev server as a drop in replacement for the webpack dev sever. 
This server webpacks the assets into a memory-fs and serves the files from there to avoid building artifacts. It does not offer hot reloading by design as that was causing some test instability.

### DIFF
--- a/docs/ExpressDevServerService.md
+++ b/docs/ExpressDevServerService.md
@@ -1,0 +1,36 @@
+# ExpressDevServer Service
+ExpressDevServer Service is provided as a convenience to start an express server and return a promise when the webpack compiler is complete. This enables test setup to be complete before testing is began. This service must be provided as a service to webdriver.io to host your test pages.
+
+## Options
+
+In the wdio.conf.js you can pass a configuration object with the following structure:
+
+* **webpackConfig** - the webpack configuration used to start the express server. Must be provided to use this service.
+* **expressDevServer** - the webpack-dev-server configuration to be passed to the webpack-dev-server. Defaults to { index: 'index.html', port: '8080' }.
+
+```js
+// wdio.conf.js
+const wdioConf = require('terra-toolkit/wdio/conf');
+const webpackConfig = require('./webpack.config.js');
+const ExpressDevService = require('terra-toolkit/wdio/services/index').ExpressDevService;
+const localIP = require('ip');
+
+const port = 8080;
+
+const config = {
+  ...wdioConf.config,
+
+  // Point base URL at the site to be tested for correct webdriver.io setup
+  baseUrl: `http://${localIP.address()}:${port}`,
+
+  // Configuration for WebpackDev service
+  webpackConfig,
+  expressDevServer: {
+    port,
+  },
+
+  service = wdioConf.config.services.concat([ExpressDevService]);
+};
+
+exports.config = config;
+```
--- a/docs/ExpressDevServerService.md
+++ b/docs/ExpressDevServerService.md
@@ -3,10 +3,10 @@ ExpressDevServer Service is provided as a convenience to start an express server
 
 ## Options
 
-In the wdio.conf.js you can pass a configuration object with the following structure:
-
-* **webpackConfig** - the webpack configuration used to start the express server. Must be provided to use this service.
-* **expressDevServer** - the webpack-dev-server configuration to be passed to the webpack-dev-server. Defaults to { index: 'index.html', port: '8080' }.
+| Name  | Required | Default Value | Description |
+| ------------- | ------------- | ------------- | ------------- |
+| **webpackConfig**  | true | `undefined` | The webpack configuration used to start the express server. |
+| **expressDevServer**  | false | `{ index: 'index.html', port: '8080' }` | The webpack-dev-server configuration to be passed to the webpack-dev-server. |
 
 ```js
 // wdio.conf.js
--- a/package.json
+++ b/package.json
@@ -46,9 +46,12 @@
     "async": "^2.5.0",
     "axe-core": "^2.6.1",
     "chai": "^4.1.2",
+    "express": "^4.16.2",
     "glob": "^7.1.1",
     "ip": "^1.1.5",
-    "load-json-file": "^2.0.0"
+    "load-json-file": "^2.0.0",
+    "memory-fs": "^0.4.1",
+    "mime-types": "^2.1.18"
   },
   "peerDependencies": {
     "nightwatch": "^0.9.12",
--- a/src/wdio/services/ExpressDevServerService.js
+++ b/src/wdio/services/ExpressDevServerService.js
@@ -1,0 +1,109 @@
+import express from 'express';
+import webpack from 'webpack';
+import MemoryFS from 'memory-fs';
+import mime from 'mime-types';
+import path from 'path';
+
+
+export default class WebDevServerService {
+  async onPrepare(config) {
+    if (!config.webpackConfig) {
+      // eslint-disable-next-line no-console
+      console.log('[WebDevService] No webpack configuration provided');
+      return;
+    }
+
+    const webpackConfig = config.webpackConfig;
+    const port = config.webpackPort || 8080;
+
+    await WebDevServerService.startExpressDevServer(webpackConfig, port).then((server) => {
+      this.server = server;
+    });
+  }
+
+  async onComplete() {
+    await this.stop();
+  }
+
+  static startExpressDevServer(webpackConfig, port) {
+    return WebDevServerService.compile(webpackConfig).then((fs) => {
+      const app = express();
+
+      // Setup a catch all route, we can't use 'static' because we need to use a vfs
+      app.get('*', (req, res, next) => {
+        let filename = req.url;
+        // Setup a default to route / to the index.html file.
+        // Should this be configurable in the future?
+        if (filename === '/') {
+          filename = '/index.html';
+        }
+
+        const filepath = `${webpackConfig.output.path}${filename}`;
+
+        if (fs.existsSync(filepath)) {
+          res.setHeader('content-type', mime.contentType(path.extname(filename)));
+          res.send(fs.readFileSync(filepath));
+        } else {
+          next();
+        }
+      });
+
+      app.use((req, res, next) => {
+        const err = new Error(`Not Found: ${req.originalUrl}`);
+        err.status = 404;
+        next(err);
+      });
+
+      // error handler
+      app.use((err, req, res) => {
+        // set locals, only providing error in development
+        res.locals.message = err.message;
+        res.locals.error = err;
+
+        // render the error page
+        res.status(err.status || 500);
+        res.render('error');
+      });
+
+      const server = app.listen(port);
+      // eslint-disable-next-line no-console
+      console.log('[WebDevService] Express server started');
+
+      return server;
+    });
+  }
+
+  static compile(webpackConfig) {
+    return new Promise((resolve, reject) => {
+      const compiler = webpack(webpackConfig);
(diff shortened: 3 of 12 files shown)


src/wdio/services/ExpressDevServerService.js
```diff
@@ -0,0 +1,109 @@
...
+
+  async onComplete() {
+    await this.stop();
+  }
+
+  static startExpressDevServer(webpackConfig, port) {
+    return WebDevServerService.compile(webpackConfig).then((fs) => {
+      const app = express();
+
+      // Setup a catch all route, we can't use 'static' because we need to use a vfs
```
emilyrohrbough: > because we need to use a vfs

vfs?

src/wdio/services/ExpressDevServerService.js
```diff
@@ -0,0 +1,109 @@
...
+      app.get('*', (req, res, next) => {
+        let filename = req.url;
+        // Setup a default to route / to the index.html file.
+        // Should this be configurable in the future?
```
emilyrohrbough: Can we make this configurable now? We can access global webdriver configuration within this service. We can add/ the `express` key to pass object of information. I am thinking:

```
// wdio 
express: {
    defaultRoute: // defaults to '/index.hml'
}
```

src/wdio/services/ExpressDevServerService.js
```diff
@@ -0,0 +1,109 @@
...
+  static compile(webpackConfig) {
+    return new Promise((resolve, reject) => {
+      const compiler = webpack(webpackConfig);
+      // setup a virtual file system to write webpack files to.
+      compiler.outputFileSystem = new MemoryFS();
+      // eslint-disable-next-line no-console
+      console.log('[WebDevService] Webpack compilation started');
+      compiler.run((err, stats) => {
+        if (err || stats.hasErrors()) {
+          // console.log(err);
```
emilyrohrbough: whoops removes comments

src/wdio/services/ExpressDevServerService.js
mjhenkes: We could, but do we need to? index.html is what all our sites are set to.

src/wdio/services/ExpressDevServerService.js
mjhenkes: Yeah, those are helpful if webpack is failing, but we can remove them.

src/wdio/services/ExpressDevServerService.js
mjhenkes: Virtual file system.

src/wdio/services/ExpressDevServerService.js
```diff
@@ -0,0 +1,109 @@
...
+          // console.log(stats);
+          // eslint-disable-next-line no-console
+          console.log('[WebDevService] Webpack compiled unsuccessfully');
+          reject();
+        }
+        // eslint-disable-next-line no-console
+        console.log('[WebDevService] Webpack compiled successfully');
```
emilyrohrbough: These should be updated to `ExpresssService`

emilyrohrbough:
I think it would be valuable to consider this logged issue with introducing this service: https://github.com/cerner/terra-toolkit/issues/36

src/wdio/services/ExpressDevServerService.js
```diff
@@ -0,0 +1,109 @@
+import express from 'express';
+import webpack from 'webpack';
+import MemoryFS from 'memory-fs';
+import mime from 'mime-types';
+import path from 'path';
+
+
```
emilyrohrbough: nit: extra line

src/wdio/services/ExpressDevServerService.js
emilyrohrbough: It would open it up for consumers use

mjhenkes: Fair enough, i'll look into it.

src/wdio/services/ExpressDevServerService.js
mjhenkes: It's still a Web Dev Service, just not a Web**pack** Dev Service. :)

emilyrohrbough: Aww I see, should we be naming it ExpressDevService to clearly articulate which service provided this ?

mjhenkes:
@emilyrohrbough, issue #36 will be negated by the use of this express service. I don't think it would make sense to use webpack dev server config for this service.

src/wdio/services/ExpressDevServerService.js
mjhenkes: I went with:
`//wdio
expressDevServer: {
    index: // defaults to '/index.hml'
}
`
I used index because that more closely aligns with the webpackdev server api. https://webpack.js.org/configuration/dev-server/#devserver-index

mjhenkes:
I need to add some doc.
Edit: Done!

package.json
```diff
@@ -46,9 +46,12 @@
     "async": "^2.5.0",
     "axe-core": "^2.6.1",
     "chai": "^4.1.2",
+    "express": "^4.16.2",
```
emilyrohrbough, line 49: Should these be peer dependencies since some people will not need/want to use the Express server?

src/wdio/services/ExpressDevServerService.js
```diff
@@ -0,0 +1,106 @@
...
+      const compiler = webpack(webpackConfig);
+      // setup a virtual file system to write webpack files to.
+      compiler.outputFileSystem = new MemoryFS();
+      // eslint-disable-next-line no-console
+      console.log('[ExpresDevService] Webpack compilation started');
+      compiler.run((err, stats) => {
+        if (err || stats.hasErrors()) {
+          // eslint-disable-next-line no-console
+          console.log('[ExpresDevService] Webpack compiled unsuccessfully');
+          reject();
```
emilyrohrbough, line 86: Should we return the errors for easier debugging?

src/wdio/services/WebpackDevServerService.js
```diff
@@ -25,6 +25,8 @@ export default class WebDevServerService {
   // eslint-disable-next-line class-methods-use-this
   startWebpackDevServer() {
     return new Promise((resolve, reject) => {
+      // eslint-disable-next-line no-console
+      console.error('[WebDevService] WARNING: ExpressDevServer service should be used instead of the Terra provided WebpackDevServerService. See https://github.com/cerner/terra-toolkit/blob/master/docs/ExpressDevServerService.md');
```
emilyrohrbough: Does throwing an error make prevent the service from starting?

emilyrohrbough (approved):

src/wdio/services/ExpressDevServerService.js
```diff
@@ -0,0 +1,106 @@
...
+export default class ExpressDevServerService {
+  async onPrepare(config) {
+    if (!config.webpackConfig) {
+      // eslint-disable-next-line no-console
+      console.log('[ExpresDevService] No webpack configuration provided');
+      return;
+    }
+
+    const webpackConfig = config.webpackConfig;
+    const port = config.expressDevServer.port || 8080;
```
emilyrohrbough: It looks like the terra-core build failed due to not specifying the expressDevServer key within the wdio config.
We should update to be something like
```
const expressConfig = config.expressDevServer.port || 8080;
const port = expressConfig.port || 8080;
```

emilyrohrbough: Also, with the nightwatch tests passing and the wdio test passing it made realized we also need to update our nightwatch configuration. It uses the [WebpackDevServer service](https://github.com/cerner/terra-toolkit/blob/master/src/nightwatch/nightwatch.config.js#L24).

emilyrohrbough (changes requested):

docs/ExpressDevServerService.md
```diff
@@ -0,0 +1,36 @@
+# ExpressDevServer Service
+ExpressDevServer Service is provided as a convenience to start an express server and return a promise when the webpack compiler is complete. This enables test setup to be complete before testing is began. This service must be provided as a service to webdriver.io to host your test pages.
+
+## Options
+
+In the wdio.conf.js you can pass a configuration object with the following structure:
+
+* **webpackConfig** - the webpack configuration used to start the express server. Must be provided to use this service.
```
bjankord: Displaying this in a table may be easier for consumers to read through.

| Name  | Required | Default Value | Description |
| ------------- | ------------- | ------------- | ------------- |
| **webpackConfig**  | true | `undefined` | The webpack configuration used to start the express server. |
| **expressDevServer**  | false | `{ index: 'index.html', port: '8080' }` | The webpack-dev-server configuration to be passed to the webpack-dev-server. |

package.json
bjankord: If these are moved to peerDeps, they should also be added as devDeps so we can still test toolkit with them.

src/wdio/services/WebpackDevServerService.js
bjankord: Potentially, I'd say we change this `console.warn` or even `console.log`.

package.json
emilyrohrbough: The current toolkit setup is listing the Service dependencies under the peer dependencies and then describing the dependency requirements in the read-me sections. see https://github.com/cerner/terra-toolkit/blob/master/docs/Nightwatch_Utility.md#getting-started

src/wdio/services/ExpressDevServerService.js
bjankord: I'd say yes, debugging with minimal to no error output is always painful. The could be opt-in via a verbose config but I think we just default verbose error output.

bjankord (changes requested):
Left some requested changes on the review

src/wdio/services/WebpackDevServerService.js
mjhenkes: I based the decision for error based on this already added depreciation warning. https://github.com/cerner/terra-toolkit/blob/master/src/nightwatch/nightwatch.config.js#L14
Do we want to change that to a `warn` too?

ryanthemanuel: I think one might actually be an alias of the other: https://nodejs.org/api/console.html#console_console_warn_data_args

package.json
mjhenkes: I think the current setup is incorrect. We should not be using peer dependencies as 'optional' dependancies. The peer dependency warnings produced do not indicate that this is an 'optional' dependency. terra-toolkit should be added as a dev dependency to any project consuming it. The only harm that could be caused by including express would be a larger node_modules folder in the dev env. The delivered package size to the client won't be increased, nor the production server size. For terra-toolkit to work properly we should add any dependancies that are needed for it's components to work as dependencies. If we feel strongly about not having this service available by default we should split it out into it's own package. I don't think that is necessary.

src/wdio/services/ExpressDevServerService.js
mjhenkes: The error is corrected, however I do not want to update the nightwatch test runner. We're deprecating nightwatch and would rather see tests converted to wdio, than switch nightwatch to the new test server.

package.json
emilyrohrbough: Fair enough. +1

src/wdio/services/ExpressDevServerService.js
emilyrohrbough: True. The more warning with the test setup will motive people to switch testing strategies.

emilyrohrbough (approved):

bjankord (approved):

ryanthemanuel (approved):

PR #72, opened 2018-03-30
Prepare for web pack 4

mjhenkes:
### Summary
* Reduced testing webpack config to the bare minimum needed.
* Nightwatch now uses the Express Dev Server Service.
* Express Dev Server Service now fails silently for '/favicon.ico' routes when no favicon exists.

@cerner/terra


mjhenkes:
Seriously Danger? Theres 7 files.

mjhenkes:
The webpack config ends up being kinda silly. There's no javascript or css to pack, it's just putting html files in a vfs so that our express dev server service can read them.

package.json
```diff
@@ -62,33 +62,21 @@
...
-    "node-sass": "^4.5.2",
-    "postcss-loader": "^2.0.6",
-    "raw-loader": "^0.5.1",
     "rimraf": "^2.6.1",
-    "sass-loader": "^6.0.6",
     "shelljs": "^0.7.6",
-    "style-loader": "^0.19.0",
     "webdriverio": "^4.9.8",
     "webpack": "^3.6.0",
     "webpack-dev-server": "2.7.1",
```
bjankord, line 82: Looks like this can be removed too since we're removing `devServer` config from `test.config`

bjankord:
@mjhenkes DangerJS display's that message [based on if there are more than 1000 combined additions or removals](https://github.com/cerner/terra-toolkit/blob/master/dangerfile.js#L28-L31).

bjankord (approved):

package.json
mjhenkes: It's still referenced here in the webpack dev server service and people using toolkit 2.0 and wdio will still be using this because the config to switch to the express dev server service is in their packages :( . https://github.com/cerner/terra-toolkit/blob/master/src/wdio/services/WebpackDevServerService.js

tests/test.config.js
```diff
@@ -1,70 +1,12 @@
-// By default eslint assumes packages imported are supposed to be dependencies,
-// not devDependencies. Disabling this rule in webpack.conig.js
-/* eslint-disable import/no-extraneous-dependencies */
-const webpack = require('webpack');
-
```
emilyrohrbough, line 5: 😄 this is still funny to me because I'm the one who made this so complex

emilyrohrbough (approved):

PR #74, opened 2018-04-05
Migrate i18n aggregation

mjhenkes:
### Summary
This is a straight copy from terra-i18n to terra-toolkit. 
Do note that I intend to prepend all terra-toolkit exported commands with the 'tt:' namespace. In this case a namespace is needed to avoid a conflict with the aliased terra-i18n aggregate-translations command.

I will put out a corresponding terra-core pull request consuming this change so you can see how this is used.

@cerner/terra


tests/jest/aggregate-messages.test.js
```diff
@@ -6,7 +6,7 @@ global.console = { warn: jest.fn() };
 
 const translationDirectories = [
   path.resolve(__dirname, 'fixtures', 'translations'),
-  path.resolve(__dirname, 'fixtures', 'node_modules', 'fixtures1', 'translations'),
+  path.resolve(__dirname, 'fixtures', 'ned_modules', 'fixtures1', 'translations'),
```
bjankord: wat?

scripts/aggregate-translations/aggregate-translations.js
```diff
@@ -0,0 +1,50 @@
...
+  path.resolve(baseDirectory, 'node_modules', 'terra-*', 'translations'), // root level dependency translations
+  path.resolve(baseDirectory, 'packages', 'terra-*', 'translations'), // package level translations
+  path.resolve(baseDirectory, 'packages', 'terra-*', 'node_modules', 'terra-*', 'translations'), // package level dependency translations
+]);
+
+const customDirectories = (baseDirectory, directories) =>
+  (directories.map(dir => path.resolve(baseDirectory, dir)));
+
+const aggregatedTranslations = (options) => {
+  const baseDir = (options || {}).baseDir || process.cwd();
```
JakeLaCombe, line 21: would it be better to do a store options in a temp variable at the beginning of the method instead of doing options || {} for all these variables?

tests/jest/aggregate-messages.test.js
JakeLaCombe: Let's kick it off with a nice White Wine Spriter then!

JakeLaCombe (approved):

ryanthemanuel (approved):

scripts/aggregate-translations/aggregate-translations.js
mjhenkes: This options stuff will be updated with the forthcoming webpack4 update. :)

bjankord (approved):

PR #90, opened 2018-04-24
Fail fast when running wdio tests

mjhenkes:
### Summary
This pr will fail the wdio run if one test in one test set fails.

@cerner/terra


emilyrohrbough:
Do we want to do this? It takes a long time to run WDIO and this would require multiple runs to find each screenshot that needs to be updated or fixed.

mjhenkes:
Specifically I think that our travis CI jobs should fail fast at a minimum, one failed test will prevent merging, we might as well save the resources. I do see value in what you're describing, but I think the alternate is true too. Because our tests take so long, I'd like to stop when there is a failure and fix that failure instead of waiting for the entire run.

mjhenkes:
We could potentially add an env flag to not fail fast

emilyrohrbough:
> We could potentially add an env flag to not fail fast

I was just thinking this. I see value in both sides - fail right away so we aren't waiting forever for the build to pass and then run a full sweep if you want to know each possible failure.

What if we updated the config to be like:
```js
const bail = process.env.WDIO_NO_BAIL ? 0 : 1;
config = {
...
bail,
...
}
```
Then we can add inform people they can run test without bail locally like
```bash
WDIO_NO_BAIL=true npm run wdio
```

emilyrohrbough:
Or we we default to not bailing and we can add `WDIO_BAIL` for Travis builds so people receive the same test setup locally without changing anything

mjhenkes:
I will change it to only bail in travis ci or ci

bjankord (approved):

CHANGELOG.md
```diff
@@ -3,10 +3,12 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* WDIO bails at the first failure.
```
emilyrohrbough: Lets update to reflect this option is open for process.env.TRAVIS || process.env.CI

emilyrohrbough (approved):

PR #91, opened 2018-04-24
Set the default browser size explicitly before tests.

mjhenkes:
### Summary
Set the default browser size explicitly before tests.

@cerner/terra


mjhenkes:
I also removed some screenshots that got added previously at the wrong screen size.

config/wdio/wdio.conf.js
```diff
@@ -48,4 +48,9 @@ exports.config = {
     ui: 'bdd',
     timeout: 20000,
   },
+
+  before: () => {
+    /* global browser */
+    browser.setViewportSize({ width: 1050, height: 895, name: 'default' });
```
emilyrohrbough: Would we want to define this as a terra-size ? Or leave for now till MVB?

JakeLaCombe (approved):

config/wdio/wdio.conf.js
mjhenkes: the Terra variable isn't available here for some reason.

mjhenkes:
3 min 37 sec for before hook vs 3 min 3 sec with docker-compose.yml
But thats just one run of each, hard to say if it is statistically significant.

emilyrohrbough:
I imagine it won't have a huge impact given our current setup coninuously adject the viewport size when multiple viewports are supplied to `terra.should.beAccessible` or `terra.should.matchScreenshot`I would be curious to verify this against the core repo for tests not defining a viewport explicitly

bjankord:
Let's move forward with the docker-compose approach.

bjankord (approved):

PR #94, opened 2018-04-27
remove 'no-multi-assign' eslint rule

emilyrohrbough:
### Summary
Resolves https://github.com/cerner/terra-core/issues/1265#issuecomment-369707164


mjhenkes (approved):
Maybe we should wrap our eslint config up into it's own repo

PR #100, opened 2018-05-14
Enable parallel testing during builds

ryanthemanuel:
### Summary

In order to speed up builds and avoid the travis time limit, this pull request intends to enable parallel testing during the build process. To accomplish this, the build will now use docker to run npm install during a build phase. Then during the test phase, it will run different wdio test suites in parallel. 

We are also going to publish a docker container that can be used throughout all of our monorepos for testing purposes.

The goal here is also to be able to run tests locally the exact same way as previously.


config/wdio/wdio.conf.js
```diff
@@ -6,12 +6,14 @@ const SeleniumDockerService = require('../../lib/wdio/services').SeleniumDocker;
 const visualRegressionConfig = require('../../lib/wdio/visualRegressionConf');
 const ServeStaticService = require('../../lib/wdio/services/index').ServeStaticService;
 const path = require('path');
+const PackageUtilities = require('lerna/lib/PackageUtilities');
+const Repository = require('lerna/lib/Repository');
```
emilyrohrbough, line 10: Should we include a `lerna` dependency for test setup?

ryanthemanuel: It should be included transitively.

config/docker/ci-docker-compose.yml
```diff
@@ -0,0 +1,24 @@
+version: '3'
+
+services:
+  base:
+    build: ./
```
emilyrohrbough: So based on this, 

screen shot 2018-05-14 at 5 07 51 pm


I think we can just specify `build: .` here and in test. I almost feel like we could drop the base service.

emilyrohrbough: I was thinking about this more, and we should be able to specify `build: .` in the test service but we'd want to keep what we have for base

.travis.yml
```diff
@@ -1,15 +1,46 @@
...
+      script:
+        - echo "$DOCKER_PASSWORD" | docker login -u "ryanthemanuel" --password-stdin
+        - 'docker build -t ryanthemanuel/terra-node -f ./docker/terra-node/Dockerfile ./docker/terra-node'
+        - 'docker push ryanthemanuel/terra-node'
+      if: type != pull_request
+    - stage: release snapshot of terra-node
+      script:
+        - echo "$DOCKER_PASSWORD" | docker login -u "ryanthemanuel" --password-stdin
+        - "docker build -t ryanthemanuel/terra-node:$TRAVIS_PULL_REQUEST_BRANCH -f ./docker/terra-node/Dockerfile ./docker/terra-node"
+        - "docker push ryanthemanuel/terra-node:$TRAVIS_PULL_REQUEST_BRANCH"
```
ryanthemanuel: Do we want the snapshot versions to be associated to the branch?  Or just a generic snapshot?  Also, note to self, I'll need this to be cerner rather than ryanthemanuel

config/wdio/wdio.conf.js
emilyrohrbough: Would this have an adverse affect on non-mono-repo projects? And/or is this a dependency we want included in toolkit?

.dockerignore
```diff
@@ -0,0 +1 @@
+node_modules
```
emilyrohrbough, line 7: I have seen other git repos that have added things like `.git` to this file. For example, [docker/compose's .dockerignore](https://github.com/docker/compose/blob/master/.dockerignore)

docker/terra-node/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+WORKDIR $APP_HOME
+
+# Mount Source into Container
+ONBUILD ADD . /app
+
+# expose the port
+EXPOSE 8080
+
+ONBUILD RUN npm config set unsafe-perm true
+ONBUILD RUN npm install
```
mjhenkes: You can just set the flag for unsafe perm. It's also a touch faster if you do it silently
```
npm install --unsafe-perm --silent
```

config/wdio/wdio.conf.js
ryanthemanuel: I think the only adverse effect (since its a dev dependency) would be a longer install time.  I talked with @bjankord offline about this and he was ok with this added install time.

ryanthemanuel:
Note that this *will* warrant a major version bump

.dockerignore
ryanthemanuel: Done

config/docker/ci-docker-compose.yml
ryanthemanuel: I actually simplified this greatly.  Dropped the test service entirely.

emilyrohrbough: Nice. So I am wondering, should we change the `base` service to be chrome specific? like `test-chrome`?

Dockerfile
```diff
@@ -0,0 +1,5 @@
+# use node as base image
+FROM ryanthemanuel/terra-node:test_parallel
```
emilyrohrbough: Where can I find this?

emilyrohrbough:
We need a migration guide for v3 to v4

.dockerignore
mjhenkes: I suppose we'll have to duplicate this docker ignore in each repo. lib and build probably don't matter much here.

mjhenkes (approved):

Dockerfile
ryanthemanuel: It gets created as part of the terra-toolkit build.  It's in the docker directory

config/docker/ci-docker-compose.yml
ryanthemanuel: We could do that.

config/wdio/wdio.conf.js
```diff
@@ -51,3 +53,28 @@ exports.config = {
...
   },
 };
+
+if (ci) {
+  config.host = 'standalone-chrome';
+}
+
+// This code only executes for monorepos.  It will create a set of suites that can then be executed
+// independently and/or in parallel via 'wdio --suite suite1' for example
+const isRepoTest = !process.cwd().includes('/packages/');
```
emilyrohrbough: We need to update this to be `const isRepoTest = !process.cwd().includes('packages');` for windows 

Windows has `\\` in their paths

mjhenkes: Good catch

emilyrohrbough (approved):

PR #133, opened 2018-07-16
Default host to '0.0.0.0' for docker development

ryanthemanuel:
### Summary

In order to access a running docker container, it will be helpful to default the host for our servers to '0.0.0.0'.  This can still be overridden by consumers.

### DIFF
--- a/scripts/serve/serve.js
+++ b/scripts/serve/serve.js
@@ -10,7 +10,7 @@ const webpackServe = (options) => {
   const serveConfig = {
     config,
     ...(port) && { port },
-    ...(host) && { host },
+    ...(host) && { host || '0.0.0.0' },
     dev: {
       stats: {
         colors: true,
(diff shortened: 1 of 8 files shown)


scripts/serve/serve.js
```diff
@@ -10,7 +10,7 @@ const webpackServe = (options) => {
   const serveConfig = {
     config,
     ...(port) && { port },
-    ...(host) && { host },
+    ...(host) && { host || '0.0.0.0' },
```
emilyrohrbough: This fallback will only be hit if called serve as a function. We we want to add a fallback for this case? We don't for the port value, which is defaulted to `8080` in the cli. If we do, would it be worth adding a server/serves-static defaults config file so we only need to update fallback values in one place?

dkasper-was-taken (approved):

mjhenkes (approved):

StephenEsser (approved):

emilyrohrbough (approved):

emilyrohrbough:
Looks good after we add a changelog regarding this new default.

nfaltermeier (approved):

bjankord (approved):

PR #149, opened 2018-08-09
Update doc for aggregate-translations CLI

dv297:
### Summary
I think there is some discrepancy in how Commander handles arguments across Windows and Macs. When using the aggregate-translations command, if you have a space between the locales,
```
tt-aggregate-translations -b ./ -d ./translations -l ['en', 'es'] -o ./aggregated-translations
```
...Commander (at least on Windows), won't group the argument together, which was causing an incorrect aggregated-translations folder to be generated. I've included a screenshot.

with-space

Even if I remove the space
`['en','es']`
...it doesn't quite parse the list correctly
capture

So I added a small change to the Regex to strip the extra quotes. I also removed the .map(String), though maybe I'm missing an edge case since I'm assuming everything after the split will be a string.

But I think the more important change here is just to change the docs so that you take in a comma-separated list rather than some pseudo-array structure. I think the pseudo-array structure implies that you can use any valid JS-array syntax (with spaces included) and that it'd work, but it's not quite the case, at least on Windows. The regex change at least maintains backwards compatibility with the older syntax as long as you remove spaces.

### Additional Details

@cerner/terra

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -12,6 +12,7 @@ Cerner Corporation
 - Alex Bezek [@alex-bezek]
 - Noah Benham [@noahbenham]
 - Ben Boersma [@BenBoersma]
+- Daniel Vu [@dv297]
 
 [@emilyrohrbough]: https://github.com/emilyrohrbough
 [@brettjankord]: https://github.com/bjankord
@@ -25,3 +26,4 @@ Cerner Corporation
 [@alex-bezek]: https://github.com/alex-bezek
 [@noahbenham]: https://github.com/noahbenham
 [@BenBoersma]: https://github.com/BenBoersma
+[@dvv297]: https://github.com/dv297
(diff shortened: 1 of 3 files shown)


emilyrohrbough:
Hey @dv297 you were correct that the documentation on how to use the aggreagated-translation-cli tool was incorrect and changing how cli values can be passed is a no-passive change.

We use [this same logic in the tt-wdio-cli script](https://github.com/cerner/terra-toolkit/blob/master/scripts/wdio/wdio-runner-cli.js#L18) and we actually use these cli options in toolkit itself for testing [here](https://github.com/cerner/terra-toolkit/blob/master/package.json#L60). So this setup doesn't need a change, only docs.

Would you mind updating the documentation to read:
```diff
-"aggregate-translations": "tt-aggregate-translations -b ./ -d ./src/**/translations -d ./translations -l ['en', 'es'] -o ./aggregated-translations",
+"aggregate-translations": "tt-aggregate-translations -b ./ -d ./src/**/translations -d ./translations -l=['en','es'] -o ./aggregated-translations",
```

bjankord:
Talked with @emilyrohrbough about this, let's go with this example @dv297 for the docs 
```
-l ['en','es']
```

bjankord (approved):

PR #153, opened 2018-08-17
Bump node-sass version to latest release

bjankord:
### Summary
Ran into some issues with using node-sass 4.9.2 on Mac OS High Sierra 10.13. Bumping the node-sass version to 4.9.3 seems to resolve the issues we've seen on High Sierra.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Bumped node-sass dependency to v4.9.3. This version should work better on latest High Sierra OS on mac.
 
 4.9.0 - (August 14, 2018)
 ----------
--- a/package-lock.json
+++ b/package-lock.json
@@ -787,11 +787,6 @@
         }
       }
     },
-    "assert-plus": {
-      "version": "0.2.0",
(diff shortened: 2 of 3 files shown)


bjankord:
Testing this on @amichaelparker's new mac now.

emilyrohrbough:
For a clean install on his device, it should have been pulling this version anyways. Is the project he is pulling also defining this dependency?

emilyrohrbough (approved):

StephenEsser (approved):

bjankord:
@emilyrohrbough  He was working in terra-core. It looks like we don't explicitly set a devDepedency on node-sass in that repo and instead rely on the version in terra-toolkit, which I think is fine. I'm not 100% sure, but I think the package-lock.json plays into what version is installed in this scenario. Even though we should get the latest version of 4.x using [the caret](https://github.com/cerner/terra-toolkit/blob/master/package.json#L85), the [package-lock file for this repo uses 4.9.2](https://github.com/cerner/terra-toolkit/blob/master/package-lock.json#L11033) so I'm guessing that's why we're not seeing the latest 4.x version being used in terra-core.

yuderekyu (approved):

mjhenkes (approved):

PR #168, opened 2018-09-25
Lock down webpack version.

jmsv6d:
### Summary
Lock down webpack version to prevent errors in webpack-cli while releasing.

@cerner/terra

Thanks for contributing to Terra! -->

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Unreleased
 ----------
 ### Changed
 * Updated linter and linter plugins
+* Locked webpack version to 4.19.0
 
 4.11.1 - (August 29, 2018)
 ----------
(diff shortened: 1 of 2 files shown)


bjankord (approved):

ryanthemanuel (approved):

pmoeller91:
Was there a particular reason for locking webpack as opposed to uplifting webpack-cli? Will it be changed in the future you think?

pmoeller91:
Thanks for the response. I figured it was something like that, and I trust the judgment. 
I resolved the issue on my end by bumping up the version of webpack-cli and it did not cause me any issue with terra-toolkit, at least that I noticed, but an anecdote is definitely not enough to justify it as the go-to solution. Thanks for the really fast issue resolution time!

PR #177, opened 2018-10-03
Fix server file path extension

emilyrohrbough:
### Summary
The serve-static script was incorrect the header content type which can result in incorrect loading on the page.

What happened was the serve-static server would remove any hashing following a file extension when it determined the file path. However, when serve-static would parse the extension name, it would return the non-cleaned up extension. This resulted in ab invalid content type returned by mime which was then set as the header content type. 

##### Before:
screen shot 2018-10-03 at 2 01 28 pm

##### After:
screen shot 2018-10-03 at 2 03 12 pm

I have been working with @nramamurth on this issue. Thank you for exposing this issue. Can you verify this has been resolved for you?


dkasper-was-taken (approved):

nramamurth:
This did resolve the issue that I was facing. This is a good find! 🎉

tbiethman (approved):

mjhenkes (approved):

PR #183, opened 2018-10-09
Pause for Selenium IE Browsers

emilyrohrbough:
### Summary
Add pause to the Terra service setup for IE browsers when setting the viewport. IE browsers need more time before being interacted with.

### DIFF
--- a/src/wdio/services/TerraService.js
+++ b/src/wdio/services/TerraService.js
@@ -61,7 +61,9 @@ export default class TerraService {
     chai.Assertion.addMethod('matchReference', chaiMethods.matchReference);
     console.log(global.browser.timeouts());
     global.browser.timeouts({
-        "script": 3000,
+      "implicit": 0,
+      "pageLoad": 300000,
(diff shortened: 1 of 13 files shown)


src/wdio/services/TerraService.js
```diff
@@ -58,5 +57,10 @@ export default class TerraService {
     };
     chai.Assertion.addMethod('accessible', chaiMethods.accessible);
     chai.Assertion.addMethod('matchReference', chaiMethods.matchReference);
+    // IE driver takes a longer to be ready for browser interactions
```
yuderekyu, line 60: just curious, have an idea how much longer on average?

emilyrohrbough: Searching online says give it 10 seconds. I don't have a set time. Sometimes this pause isn't needed because the browser is ready, sometimes its slower.

yuderekyu (approved):

mjhenkes (approved):

PR #188, opened 2018-10-25
Fix Serve CLI Port Arg

emilyrohrbough:
### Summary
When using the serve-cli, the port gets passed as string but the commandar module assumes the passed cli port value should be a number due to the default being 8080. This caused NaN to be passed for the port value, thus the server would fail to start. 

### Test
Pull branch & run `npm run start -- --port 9090`

### DIFF
--- a/scripts/serve/serve-cli.js
+++ b/scripts/serve/serve-cli.js
@@ -9,7 +9,7 @@ commander
   .version(packageJson.version)
   .option('--config <path>', 'The webpack config to serve.', undefined)
   .option('--host <s>', 'Sets the host that the server will listen on. eg. \'10.10.10.1\'', '0.0.0.0')
-  .option('--port <n>', 'The port the server should listen on.', parseInt, 8080)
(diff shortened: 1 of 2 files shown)


scripts/serve/serve-cli.js
```diff
@@ -9,7 +9,7 @@ commander
   .version(packageJson.version)
   .option('--config <path>', 'The webpack config to serve.', undefined)
   .option('--host <s>', 'Sets the host that the server will listen on. eg. \'10.10.10.1\'', '0.0.0.0')
-  .option('--port <n>', 'The port the server should listen on.', parseInt, 8080)
+  .option('--port <n>', 'The port the server should listen on.', Number(parseInt), 8080)
```
emilyrohrbough, line 12: This is not an issue for serve-static because we do not define a default value, however I could make this change there as well.

mjhenkes (approved):

tbiethman (approved):

dkasper-was-taken (approved):

PR #198, opened 2018-11-06
Added a tt-pack script to pack packages into a tar archive file

StephenEsser:
### Summary
Added a reusable script to pack a package into a tar archive file.

### Additional Details
There are some [naming concerns](https://github.com/cerner/terra-clinical/pull/417#discussion_r231282413) with the new script. Many projects implement an existing pack command as a common pattern for packing webpack assets.  

The new command is a wrapper for running [npm pack](https://docs.npmjs.com/cli/pack).

@cerner/terra

### DIFF
--- a/bin/pack.js
+++ b/bin/pack.js
@@ -1,0 +1,3 @@
+#!/usr/bin/env node
+
+require('../scripts/pack/pack');
--- a/scripts/pack/README.md
+++ b/scripts/pack/README.md
@@ -14,13 +14,13 @@ If there is an existing tar file it will be overwritten.
 ```JSON
 {
   "scripts": {
-    "pack": "tt-pack"
+    "tt-pack": "tt-pack"
   }
 }
 ```
 ### Running the command
 ```
-npm run pack
+npm run tt-pack
 ```
 ### Output
 package-name.tgz
--- a/scripts/pack/README.md
+++ b/scripts/pack/README.md
@@ -5,7 +5,7 @@ Terra Toolkit offers a way to pack individual packages into a tar archive file.
 ## Description
 `tt-pack` utilizes [npm pack](https://docs.npmjs.com/cli/pack) to create a tarball from a package.
 
-The tar archive file will be written to the current working directory in the following format: `<package-name>.tgz`. Assets should be compiled before packing.
+The tar archive file will be written to the current working directory in the following format: `<package-name>.tgz`. Assets should be compiled before packing. Everything except the items declared in the npmignore will be included within the tar archive file.
 
 If there is an existing tar file it will be overwritten.
 
--- a/scripts/pack/README.md
+++ b/scripts/pack/README.md
@@ -5,7 +5,7 @@ Terra Toolkit offers a way to pack individual packages into a tar archive file.
 ## Description
 `tt-pack` utilizes [npm pack](https://docs.npmjs.com/cli/pack) to create a tarball from a package.
 
-The tar archive file will be written to the current working directory in the following format: `<package-name>.tgz`.
+The tar archive file will be written to the current working directory in the following format: `<package-name>.tgz`. Assets should be compiled before packing.
 
 If there is an existing tar file it will be overwritten.
 
--- a/scripts/pack/README.md
+++ b/scripts/pack/README.md
@@ -1,0 +1,46 @@
+# Terra Toolkit Pack
+
+Terra Toolkit offers a way to pack individual packages into a tar archive file.
+
+## Description
+`tt-pack` utilizes [npm pack](https://docs.npmjs.com/cli/pack) to create a tarball from a package.
+
+The tar archive file will be written to the current working directory in the following format: `<package-name>.tgz`.
+
+If there is an existing tar file it will be overwritten.
+
+## Usage
+### In your package.json
+```JSON
+{
+  "scripts": {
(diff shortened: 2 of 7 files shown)


bin/pack.js
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
```
emilyrohrbough, line 1: Because npm does provide its own pack script, is there a possibility for this `bin/pack` script to override it in the node modules dir or does npm's pack script live somewhere else?

StephenEsser: I believe all scripts defined in the bin folder are protected by the required `run` command to invoke them. 

`npm pack` vs `npm run pack`

Both commands are still working and don't appear to have any conflicts.

StephenEsser:
I only have 7 files terra-bot.

scripts/pack/README.md
```diff
@@ -0,0 +1,46 @@
...
+The tar archive file will be written to the current working directory in the following format: `<package-name>.tgz`.
+
+If there is an existing tar file it will be overwritten.
+
+## Usage
+### In your package.json
+```JSON
+{
+  "scripts": {
+    "pack": "tt-pack"
```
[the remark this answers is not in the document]
mjhenkes: I vote tt-pack

mjhenkes (approved):

scripts/pack/pack.js
```diff
@@ -0,0 +1,26 @@
+/* eslint import/no-extraneous-dependencies: ["error", {"devDependencies": true}] */
```
emilyrohrbough: Which dependency is throwing the error for being a dev-dependency? Any dependency used here should be shipped for others to use this script

scripts/pack/pack.js
```diff
@@ -0,0 +1,26 @@
...
+const fs = require('fs');
+const path = require('path');
+const shell = require('shelljs');
+
+const packageName = path.basename(process.cwd());
+const archiveName = `${packageName}.tgz`;
+
+// Remove the previous archive file if one exists.
+if (fs.existsSync(archiveName)) {
+  shell.rm(archiveName);
```
emilyrohrbough, line 10: I'm surprised we don't need to use `-rf` here? Is a zipped folder treated as a single file? It makes since that is is given the name "tar file", just didn't think of it 😅

scripts/pack/pack.js
```diff
@@ -0,0 +1,26 @@
...
+shell.exec('npm pack --ignore-scripts', { silent: true }, (code, stdout, stderr) => {
+  if (code !== 0) {
+    /* eslint-disable-next-line no-console */
+    console.log(stderr);
+    shell.exit(-1);
+  }
+
+  // By default npm pack will append the package version to the tar archive file name.
+  // Rename the tar achive file to exclude the appended version.
+  shell.mv(stdout, archiveName);
```
emilyrohrbough, line 23: @StephenEsser What happens if two banchs have different tarfiles checked in? `master` has the latest version and a 'test-enhancement' branch has a different version. How will this resolve?

emilyrohrbough: I think I take back my question, you would need to reference the branch here:
`"package-name": "https://raw.githubusercontent.com/org/project/blob/package-name.tgz"`

scripts/pack/README.md
```diff
@@ -0,0 +1,46 @@
+# Terra Toolkit Pack
+
+Terra Toolkit offers a way to pack individual packages into a tar archive file.
```
emilyrohrbough, line 3: In your prototype PR in terra-clinical, you updated the compile script to run 
```
"compile": "npm run compile:clean && npm run compile:build && tt-pack",
```
I know this is kinda a given but I wonder if its worth mentioning if this packs the whole package with src and lib, or just lib, if we need to compile first with babel etc.

scripts/pack/pack.js
StephenEsser: Shelljs, I'll update this to be a dependency. I don't think it'll cause any issues.

emilyrohrbough (approved):

scripts/pack/README.md
StephenEsser: I added a note that assets should be compiled before running this command. Pack works just like releasing your package. It will pack everything in the package except the items blacklisted in the npmignore. 

https://github.com/cerner/terra-toolkit/pull/198/commits/2481a6e49cead4eba3e9bbdec6344d1009f41be6

scripts/pack/pack.js
StephenEsser: Moved to dependency: https://github.com/cerner/terra-toolkit/pull/198/commits/2481a6e49cead4eba3e9bbdec6344d1009f41be6

scripts/pack/README.md
emilyrohrbough: > It will pack everything in the package except the items blacklisted in the npmignore.

Lets add that as well. This would be handy to know if people were curious what a new package would be released with.

StephenEsser: https://github.com/cerner/terra-toolkit/pull/198/commits/78bcb62cf71077a3e462da7e6cfbda93d45c274b

BenBoersma (approved):

benbcai (approved):

bjankord (approved):

PR #202, opened 2018-11-12
Fix Default Port Value

emilyrohrbough:
### Summary
Default port value is returning NaN.


tbiethman (approved):

StephenEsser (approved):

mjhenkes (approved):

bjankord (approved):

Dockerfile
```diff
@@ -2,4 +2,4 @@
 FROM cerner/terra-node
 
 # run the server
-CMD ["npm", "run", "start"]
+CMD ["npm", "run", "start", "--", "--port", "8080"]
```
mjhenkes: This is temporary until we get a clean release right?

emilyrohrbough: Yes should be. @ryanthemanuel When do we push the 'latest' tagged docker image?

mjhenkes: Travis ci does it. https://github.com/cerner/terra-toolkit/blob/master/.travis.yml#L58

mjhenkes: Though, i'd be remiss if i didn't point out that if we were testing against this version and not the previously released version, this issue wouldn't have been released.

tbiethman:
+1 for commit messages

bjankord:
Can we test this branch against terra-core?

yuderekyu (approved):

mjhenkes (approved):
+1 final changes

Dockerfile
mjhenkes: Also i'm pretty sure my previous comment was wrong. Please disregard. :)

emilyrohrbough:
After spending way too log going down a lovely rabbit hole with @mjhenkes , it was discovered `har-validator` project released v5.1.3 three days ago and must have [corrupted/removed the v5.1.2 version](https://github.com/ahmadnassri/node-har-validator/issues/113). This is why the builds were continuously failing on the install step.

screen shot 2018-11-13 at 10 01 53 am


Removing the package-lock did the trick to clean this up.

PR #217, opened 2018-12-13
Add local and network address display to tt-serve

cody-dot-js:
### Summary

Inspired by [Storybook.js](https://storybook.js.org/). Add network address display when using `tt-serve`

#### Storybook Example

image

#### Terra-Toolkit Example

screen shot 2018-12-13 at 3 44 26 pm

@cerner/terra

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changelog
 Unreleased
 ----------
 ### Added
+* Add local and network address display to tt-serve
 * Add exclude directories option to aggregateTranslations script and cli
 
 4.20.0 - (December 6, 2018)
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "babel-polyfill": "^6.23.0",
     "browserslist-config-terra": "^1.0.0",
     "chai": "^4.1.2",
+    "chalk": "^2.4.1",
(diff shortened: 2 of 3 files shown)


bjankord (approved):

emilyrohrbough (commented):
So... I have experienced running tests via a VPN and it will display your local IP but this IP is not the masked IP assigned via by the VPM that you should be the network IP. 

Also, what's the size of add the chalk module to terra-toolkit. Do we want that to be a dependency pulled in by all projects?

bjankord:
Size of chalk: https://bundlephobia.com/result?p=chalk@2.4.1
This dependency would not be included in webpack bundles, it is only used in NodeJS context.

bjankord:
Updated package-lock in this commit: https://github.com/cerner/terra-toolkit/commit/14eac6304465ccbdc5d84b95cdc632545d228468

cody-dot-js:
Hmm.. not sure why the package-lock wasn't ever committed.. I always do `git commit -am '...'` or `git add . && git commit -m '...'` 🤔

bjankord:
Logged https://github.com/cerner/terra-toolkit/issues/218 so we can better track that the package-lock file gets updated when we update dependencies / devDependencies

PR #232, opened 2019-01-07
Add babel-plugin-syntax-dynamic-import to terra-toolkit

### LINKED ISSUE #201: Add babel-plugin-syntax-dynamic-import
# Feature Request

## Description
To better support modern code splitting, we should add [babel-plugin-syntax-dynamic-import](https://www.npmjs.com/package/babel-plugin-syntax-dynamic-import). We'll want to use the plugin compatible with babel 6 until we're ready to upgrade to babel 7. When we upgrade to babel 7, we'll want to swap out the plugin to use the [babel 7 compatible plugin](https://www.npmjs.com/package/@babel/plugin-syntax-dynamic-import).

JakeLaCombe:
### Summary
Resolves #201

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,5 +5,5 @@
         "browsers": [ "extends browserslist-config-terra" ]
       }
     }]],
-  "plugins": ["transform-object-assign", "transform-object-rest-spread", "transform-runtime"],
+  "plugins": ["transform-object-assign", "transform-object-rest-spread", "transform-runtime", "syntax-dynamic-import"],
 }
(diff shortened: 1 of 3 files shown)


bjankord (approved):

emilyrohrbough:
Has this been verified in our repos? Are there an any downstream affects for adding this in projects that are using require.ensure instead of dynamic imports?

JakeLaCombe:
I've tested it in terra-core and terra-framework, and it appears to be building just fine. I ran it in IE11 as well, and I don't see any issues while accessing some of the packages that need translations.

emilyrohrbough (approved):

PR #234, opened 2019-01-08
Use Dynamic Imports for loading translations

### LINKED ISSUE #175: Enable format option with aggregated-translations pre-build tool.
# Feature Request

## Description
We've had some users run into issues where the generated files we produce via the aggregated-translations pre-build tool don't work in IE as it uses ES6+ features not supported in that browser natively. In our webpack config, we run those files through babel, though people who are not using our webpack config may miss running these generated files through babel resulting in code that doesn't work in one of our supported browsers.

To help with this, I propose we add a `format` option to the aggregated-translations pre-build tool.
By default, `format` would be `undefined` and would result in the generated files using an `ES5` format. Generated files would use `var` statements and functions instead of `const` statements and arrow functions. By making the default `undefined`, users wouldn't have to change anything in therir code to get a version of the generated file that will work in all browsers terra-ui supports.

By setting the `format` option to `modern`, we would generate a file that uses modern ES6+ features intended to be run through babel. I'm flexible on what name we use for value that enables this format, for now I went with `modern`.

This format option resolves [an issue we saw recently](https://github.com/cerner/terra-core/issues/1830). It also solves a compatibility issue we'd run into with trying to use aggregated-translations pre-build tool with create-react-app v2. CRA v2 prevents the use of `require.ensure`, which is what we use in our current generated files. When using the `modern` format, we would switch to use `dynamic imports`, which CRA v2 supports.

As we add this `format` option, we wouldn't need to add dynamic imports at the same time. We could enhance the `modern` format to use dynamic imports at a later date and just have the `modern` version generate files the same as we do currently with `const` and arrow functions.

JakeLaCombe:
### Summary
Resolves #175

Different development teams are having issues using the translation syntax that isn't ES5. In addition, Create React App 2 will not work with require.ensure syntax that our translations currently use. To resolve this, we added the ability to export translations into ES6 syntax through the use of a format flag. Setting this to modern will output the translations with ES6 syntax.


scripts/aggregate-translations/write-i18n-loaders.js
```diff
@@ -2,25 +2,50 @@ const path = require('path');
...
+  );
+};
 
-const createTranslationLoader = (loaderName, locale) => (
-  `var ${loaderName} = function ${loaderName}(callback, scope) {
+const createTranslationLoader = (loaderName, locale, format) => {
+  if (format === 'modern') {
+    return (
+      `const ${loaderName} = (callback, scope) =>
+   import( /* webpackChunkName: "${locale}-translations" */ '${locale}.js')
```
emilyrohrbough, line 29: Does this syntax require Jest configuration changes since this will use wepackChunkName?

scripts/aggregate-translations/write-i18n-loaders.js
```diff
@@ -2,25 +2,50 @@ const path = require('path');
...
 const supportedLocales = require('./i18nSupportedLocales');
 
-const createIntlLoader = (loaderName, locale) => (
-  `var ${loaderName} = function ${loaderName}() {
+const createIntlLoader = (loaderName, locale, format) => {
+  if (format === 'modern') {
+    return (
+      `const ${loaderName} = () =>
+   import('intl/locale-data/jsonp/${locale}.js')
+   .catch(error => console.log('An error occurred while loading intl/locale-data/jsonp/${locale}.js'));\n
```
emilyrohrbough: Lets throw add the error to the error message

scripts/aggregate-translations/write-i18n-loaders.js
```diff
@@ -2,25 +2,50 @@ const path = require('path');
...
 
-const createTranslationLoader = (loaderName, locale) => (
-  `var ${loaderName} = function ${loaderName}(callback, scope) {
+const createTranslationLoader = (loaderName, locale, format) => {
+  if (format === 'modern') {
+    return (
+      `const ${loaderName} = (callback, scope) =>
+   import( /* webpackChunkName: "${locale}-translations" */ '${locale}.js')
+     .then((module) => { callback.call(scope, module);})
+     .catch(error => console.log('An error occurred while loading ${locale}.js'));\n
```
emilyrohrbough: Lets add the error to the error message

docs/AggregateTranslations.md
```diff
@@ -23,6 +23,7 @@ Once all of the translation files are created for the specified locales, the scr
 | locales  | -l, --locales | Array of Strings | The list of locale codes to aggregate. **Note: 'en' is always added if not specified.** | [terra-supported locales](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/src/i18nSupportedLocales.js) |
 | outputDir | -o, --ouputDir | String | Output directory for the translation and loader files | ./aggregated-translations |
 | configPath | -c, --config | String | The path to the terra i18n configuration file | undefined |
+| format | -f, --format | String | The format of syntax to output the translations with. Setting to undefined outputs the syntax with ES5 format. Setting format to 'modern' outputs the translations with ES6 modern syntax | undefined |
```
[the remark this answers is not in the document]
emilyrohrbough: I am wondering if it would clearer to use an option called `esVersion` or `esSyntax` for when new es versions are released and if the syntax does change. Modern makes sense at the moment, but won't if something changes later on.

If we do that, we could accept `es5` or `es6` as the syntax value which would clarify the output of the loaders

emilyrohrbough:
Has this been tested downstream? 

I am curious the downstream affect these changes may have on Jest setup. Can we get a validation PR?

JakeLaCombe:
Are you referring to a PR on terra-core that uses this package references and builds with it?

JakeLaCombe:
I tested this with terra-core, and all the jests test passed without the need for any screenshot updates. I'll run the wdio to see what happens.

emilyrohrbough:
@JakeLaCombe 
Was that tested with the fallback? That looks good to me, I am wondering more about the 'modern' output. 

I am thinking about a jest setup like the following when using the `modern` format for testing translated components. I know a lot of consumers have this setup. 
```
/* eslint-disable import/no-extraneous-dependencies */
import Enzyme, { render, shallow, mount } from 'enzyme';
import Adapter from 'enzyme-adapter-react-16';
import { IntlProvider } from 'react-intl';
import aggregateTranslations from 'terra-toolkit/scripts/aggregate-translations/aggregate-translations';
import messages from './react/translations/en.json';

// Prevent intlLoaders error
// https://github.com/cerner/terra-core/issues/1964
aggregateTranslations();

Enzyme.configure({ adapter: new Adapter() });

// Create the IntlProvider to retrieve context, allowing translations to be resolved in tests
const intlProvider = new IntlProvider({ locale: 'en-US', messages }, {});
const { intl } = intlProvider.getChildContext();

global.shallow = Component => shallow(Component, { context: { intl } });
global.render = Component => render(Component, { context: { intl } });
global.mount = Component => mount(Component, { context: { intl } });

// Prevent errors in tests; some components expect tags that terra-base provides in production code
const htmlTag = document.getElementsByTagName('html')[0];
htmlTag.setAttribute('dir', 'ltr');

// Used by Terra
window.matchMedia = () => ({ matches: true });
```

Would this be compatible?

JakeLaCombe:
I added this line to the jest setup, and it worked appropriately.

```
aggregateTranslations({format: 'modern'});
```

The one thing I did notice with that setup is that the aggregate translation has console statements, so they get outputted in the jests test output. I think I also noticed a tiny bit of slow down from the console log statements, but not a large noticable amount.

docs/AggregateTranslations.md
```diff
@@ -88,3 +90,7 @@ To provide the aggregated-translations files and loaders as modules to the terra
 This `resolve.modules` configuration indicates module resolving occurs in this order:
 1. `./aggregated_translations` (or indicated output directory)
 2. `./node_modules`
+
+## Syntax Format
+
+The aggregated-translations script provides the ability of either writing the files with ES5 syntax that is IE11 compatible, and ES6 syntax more inline with modern JavaScript. Note that if you plan on using Create React App version 2, you will need to use ES6 syntax by setting format to 'modern'.
```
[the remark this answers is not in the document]
bjankord: We can drop the note about usage in Create React App.

docs/AggregateTranslations.md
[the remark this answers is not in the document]
bjankord: I like this, I'm thinking we can keep the `format` flag and instead of passing `modern` we pass `es6`. I also want to maintain the option to not pass a format flag and have it default to es5 for passivity sake.

scripts/aggregate-translations/write-i18n-loaders.js
JakeLaCombe: added

scripts/aggregate-translations/write-i18n-loaders.js
JakeLaCombe: added

scripts/aggregate-translations/write-i18n-loaders.js
JakeLaCombe: Shouldn't. Configuring the .babelrc should be enough for this syntax to work.

bjankord (approved):

scripts/aggregate-translations/write-i18n-loaders.js
emilyrohrbough: Is this something we should be providing documentation on or is this something we should expect consumers to understand how to setup and differentiate?

JakeLaCombe: We can do a documentation update to talk about this point.

emilyrohrbough (approved):
We need to add this option to our webpack env to pass to the aggregate-translations options 

EDIT: Nvm. Looks like we spread all aggregateOptions: https://github.com/cerner/terra-toolkit/blob/master/config/webpack/webpack.config.js#L147

emilyrohrbough (approved):

PR #235, opened 2019-01-10
[DO NOT MERGE] Refactoring screenshot comparisons

ryanthemanuel:


src/wdio/services/compare/TerraCompare.js
```diff
@@ -0,0 +1,54 @@
+import BaseCompare from 'wdio-visual-regression-service/lib/methods/BaseCompare';
+import resemble from 'node-resemble-js';
+import get from 'lodash.get';
+
+export default class TerraCompare extends BaseCompare {
+  constructor(screenshotProcessor) {
+    super();
+    this.screenshotProcessor = screenshotProcessor;
```
ryanthemanuel: `screenshotProcessor` provides a mechanism for different implementations of getting reference screenshots, saving latest screenshots, saving diff screenshots, and handling a "no reference" screenshot.

config/wdio/visualRegressionConf.js
```diff
@@ -1,94 +1,7 @@
...
-    const refDir = screenshotSetup[`${ref}Dir`];
-
-    return path.join(testPath, '__snapshots__', refDir, getScreenshotDir(context), getScreenshotName(context));
-  };
-}
+const TerraCompare = require('../../lib/wdio/services/compare/TerraCompare').default;
+const LocalScreenshotProcessor = require('../../lib/wdio/services/compare/LocalScreenshotProcessor').default;
 
 module.exports = {
-  compare: new LocalCompare({
```
emilyrohrbough, line 87: We may need to think about how we could accept custom screenshot naming paths. (or decide if we allow it---maybe we shouldn't).. I don't think many people are doing it but was offered previously. For now, they would have to override the compare key, so it will work out.

src/wdio/services/compare/LocalScreenshotProcessor.js
```diff
@@ -0,0 +1,106 @@
...
+
+  static async saveDiffFileData(context, diffData) {
+    const diffScreenshotPath = LocalScreenshotProcessor.getScreenshotPath('diff', context);
+    await fs.outputFile(diffScreenshotPath, diffData);
+  }
+
+  static async handleNoReferenceScreenshot(context, latestScreenshotData, createResultReport) {
+    const referenceScreenshotPath = LocalScreenshotProcessor.getScreenshotPath('reference', context);
+    await fs.outputFile(referenceScreenshotPath, latestScreenshotData);
+    return createResultReport(0, true, true);
```
emilyrohrbough: Talked offline:

- eventually want to fail for new reference if in `process.env.CI = true`
- could add console messages for now & warn when running in CI for visibility

src/wdio/services/compare/LocalScreenshotProcessor.js
```diff
@@ -0,0 +1,106 @@
...
+        testPath = testPath.split('node_modules')[1];
+      }
+      testPath = baseDir + testPath;
+    }
+    const refDir = screenshotSetup[`${ref}Dir`];
+
+    return path.join(testPath, '__snapshots__', refDir, LocalScreenshotProcessor.getScreenshotDir(context), LocalScreenshotProcessor.getScreenshotName(context));
+  }
+
+  static async getReferenceScreenshotData(context) {
```
emilyrohrbough, line 85: Talked offline but adding here to I/we don't forget.

Not sure with how this would be called from webdriver.io offhand, but it might be worth checking if we can return the reference directory, instead of the reference screenshot, and then filter on what screenshots were actually compared against to find 'stale' screenshots

src/wdio/services/compare/TerraCompare.js
```diff
@@ -0,0 +1,54 @@
...
+  async processScreenshot(context, base64Screenshot) {
+    const referenceScreenshotData = await this.screenshotProcessor.getReferenceScreenshotData(context);
+    const latestScreenshotData = Buffer.from(base64Screenshot, 'base64');
+    await this.screenshotProcessor.saveLatestScreenshotData(context, latestScreenshotData);
+    if (referenceScreenshotData) {
+      const comparisonData = await TerraCompare.compareImages(referenceScreenshotData, latestScreenshotData);
+      const { isSameDimensions } = comparisonData;
+      const misMatchPercentage = Number(comparisonData.misMatchPercentage);
+      const misMatchTolerance = get(context, 'options.misMatchTolerance', 0.01);
+
```
emilyrohrbough, line 21: nit: since we are calling back to the BaseCompare method, can we defined
```diff
if (referenceScreenshotData) {
  // stuff
+ const isWithinMisMatchTolerance = misMatchPercentage <= misMatchTolerance

+ if (!isWithinMisMatchTolerance) {
- if (misMatchPercentage > misMatchTolerance) {
 // stuff
 // don't reporter to do outside the if block
-     return this.createResultReport(misMatchPercentage, false, isSameDimensions);
 }

- return this.createResultReport(misMatchPercentage, true, isSameDimensions);
+ return this.createResultReport(misMatchPercentage, isWithinMisMatchTolerance, isSameDimensions);
}
return this.screenshotProcessor.handleNoReferenceScreenshot(context, latestScreenshotData, this.createResultReport);
```

src/wdio/services/compare/LocalScreenshotProcessor.js
ryanthemanuel: Or we could configure it for now as opt in and change it to the default in an MVB

src/wdio/services/compare/LocalScreenshotProcessor.js
```diff
@@ -0,0 +1,111 @@
...
+    return path.join(locale, testForm, testSuite);
+  }
+
+  static getScreenshotPath(ref, context) {
+    let testPath = path.dirname(context.test.file);
+
+    const baseDir = global.browser.options.baseScreenshotDir;
+    if (baseDir) {
+      testPath = testPath.split(process.cwd())[1];
+      // Added to allow for test reusablility from terra repositories
```
bjankord, line 74: 👍

src/wdio/services/compare/TerraCompare.js
```diff
@@ -0,0 +1,55 @@
...
+  async processScreenshot(context, base64Screenshot) {
+    const referenceScreenshotData = await this.screenshotProcessor.getReferenceScreenshotData(context);
+    const latestScreenshotData = Buffer.from(base64Screenshot, 'base64');
+    await this.screenshotProcessor.saveLatestScreenshotData(context, latestScreenshotData);
+    if (referenceScreenshotData) {
+      const comparisonData = await TerraCompare.compareImages(referenceScreenshotData, latestScreenshotData);
+      const { isSameDimensions } = comparisonData;
+      const misMatchPercentage = Number(comparisonData.misMatchPercentage);
+      const misMatchTolerance = get(context, 'options.misMatchTolerance', 0.01);
```
bjankord, line 20: I feel like our misMatchTolerance should be increased. It seems we've run into a lot of diffs that are related to anti-aliasing text or other subpixel diffs.

bjankord (approved):

src/wdio/services/compare/LocalScreenshotProcessor.js
ryanthemanuel: I'd like to do this at some point, but I think this may be out of scope for this right now.

PR #237, opened 2019-01-11
Add a docker development container.

mjhenkes:
### Summary
Updating our base container to be open for development.

The docker container should be passive. docker-compose files will have to be added to the sub projects to make each developable in docker.
See: https://github.com/cerner/terra-clinical/pull/452

@cerner/terra

### DIFF
--- a/docker/terra-node/Dockerfile
+++ b/docker/terra-node/Dockerfile
@@ -1,16 +1,26 @@
-# use node as base image
-FROM node:8.9.2
+FROM node:8
 
-# Prepare Directory for Source Code
-ENV APP_HOME /app
-RUN mkdir $APP_HOME
-WORKDIR $APP_HOME
-
-# Mount Source into Container
-ONBUILD ADD . /app
-
-# expose the port
 EXPOSE 8080
 
-ONBUILD RUN npm config set unsafe-perm true
-ONBUILD RUN npm install --silent
+COPY ./entrypoint /opt/entrypoint
+
+RUN npm config set unsafe-perm true -g
+
+#install zsh / oh-my-zsh
+RUN apt-get update \
+  && apt-get install -y zsh \
+  && apt-get clean \
+  && wget https://github.com/robbyrussell/oh-my-zsh/raw/master/tools/install.sh -O - | zsh || true
+
+# Do not set a default here or the building docker file will not be able to override the variable
+ONBUILD ARG project_name
+ONBUILD ARG npm_install
+
+# Copy all the things
+ONBUILD COPY . /opt/${project_name:-module}
+
+ONBUILD WORKDIR /opt/${project_name:-module}
+
+ONBUILD RUN if [ ${npm_install:-true} = "true" ]; then npm install --loglevel error ; else echo "Skipping npm install"; fi
+
+CMD ["sh", "/opt/entrypoint/start.sh"]
--- a/docker/terra-node/entrypoint/npm_install.sh
+++ b/docker/terra-node/entrypoint/npm_install.sh
@@ -1,0 +1,8 @@
+#/bin/sh -e
+# check if the mode modules folder exists and is populated.
+if [ -n "$(ls -A node_modules 2>/dev/null)" ]
+then
+  echo "node_modules installed"
+else
+  npm install
+fi
--- a/docker/terra-node/entrypoint/npm_install.sh
+++ b/docker/terra-node/entrypoint/npm_install.sh
@@ -2,7 +2,7 @@
 # check if the mode modules folder exists and is populated.
 if [ -n "$(ls -A node_modules 2>/dev/null)" ]
 then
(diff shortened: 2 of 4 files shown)


docker/terra-node/entrypoint/npm_install.sh
```diff
@@ -0,0 +1,8 @@
+#/bin/sh -e
+# check if the mode modules folder exists and is populated.
+if [ -n "$(ls -A node_modules 2>/dev/null)" ]
+then
+  echo "node_modules installed"
```
emilyrohrbough: Lets make a bit more explicit. Something that will call out the local node_modules were used in the build. Maybe "local node_modules exists. Copying locale node_modules to docker container"

docker/terra-node/entrypoint/test.sh
```diff
@@ -0,0 +1,5 @@
+#/bin/sh -e
+
+sh /opt/entrypoint/npm_install.sh
```
emilyrohrbough, line 3: Would the npm_intall script be called when the image was build?

docker/terra-node/Dockerfile
```diff
@@ -1,16 +1,26 @@
...
+  && apt-get install -y zsh \
+  && apt-get clean \
+  && wget https://github.com/robbyrussell/oh-my-zsh/raw/master/tools/install.sh -O - | zsh || true
+
+# Do not set a default here or the building docker file will not be able to override the variable
+ONBUILD ARG project_name
+ONBUILD ARG npm_install
+
+# Copy all the things
+ONBUILD COPY . /opt/${project_name:-module}
```
emilyrohrbough, line 20: should we change opt -> app for naming consistency with other docker files consumers use

mjhenkes: by default it'll be opt -> module. In some cases (terra-dev-site) changing the directory name will cause tests to fail. So the project_name env was added to allow docker containers to set then name as they see fit and minimize refactoring to consume this change.

docker/terra-node/entrypoint/test.sh
mjhenkes: Npm install needs to be run in a container when either opts out of running npm install (env npm_install=false) during the build phase or when a volume is mounted that overwrites node modules/lib. In the case of running tests i'd assume the latter where it can allow a dev to short circuit having to build then run. They can make test changes and just run  ```docker-compose run test```. With the volume mount too will allow for caching of node modules and the use of this npm install script will not re-install node modules if they are already installed.

docker/terra-node/Dockerfile
emilyrohrbough: I was referencing the `/opt/` part of this, not the `{project-name:-module}` naming. Most dockerfiles of our consumers have `app` as the home directory, which would end up with `/app/{project-name:-module}`

mjhenkes: I think thats a cerner convention. We don't have an app, we have a module.

docker/terra-node/entrypoint/test.sh
emilyrohrbough: hmm gottchya. Makes sense. Not sure if it would be worth a short blurb here since most people building docker containers wouldn't be copying local node_modules into the container?? I have no strong opinion. 👍

docker/terra-node/Dockerfile
emilyrohrbough: True, this isn't an app. I does seem like a cerner convention to use `app`

docker/terra-node/Dockerfile
```diff
@@ -1,16 +1,26 @@
-# use node as base image
-FROM node:8.9.2
+FROM node:8
 
-# Prepare Directory for Source Code
-ENV APP_HOME /app
-RUN mkdir $APP_HOME
-WORKDIR $APP_HOME
+EXPOSE 8080
```
emilyrohrbough, line 3: Should we open this up to to accept a port arg?

docker/terra-node/Dockerfile
```diff
@@ -1,16 +1,26 @@
...
 
-# Mount Source into Container
-ONBUILD ADD . /app
+COPY ./entrypoint /opt/entrypoint
 
-# expose the port
-EXPOSE 8080
+RUN npm config set unsafe-perm true -g
```
emilyrohrbough, line 7: I forget the reason, why do we need to set unsafe-perm?

emilyrohrbough (approved):

docker/terra-node/Dockerfile
mjhenkes: It allows you to install packages as root (it does not affect all packages)
https://docs.npmjs.com/misc/config#unsafe-perm

bjankord (approved):

ryanthemanuel (approved):

PR #238, opened 2019-01-18
Add css custom property scripts

bjankord:
### Summary
* Add scripts to generate JSON file for css custom properties
* Add scripts to spell check css custom properties

Resolves https://github.com/cerner/terra-ui/issues/140

EDIT: Not sure why the file permissions changed on the bin scripts. Do we need them to be `100644`?


emilyrohrbough:
Is it possible to push the spell check script to the stylelint-config-terra module to check spelling while we verify correct css-variable names?

emilyrohrbough (commented):
Lets add jest tests for these scripts

bin/custom-properties-spell-check.js
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
+
+require('../scripts/custom-properties/spell-check');
```
emilyrohrbough, line 3: Did we change the read-write privs on the bin files `pack.js`, `serve-static.js`, etc? I notice its changed from `100644` to `100755`

emilyrohrbough:
Is it possible to create a custom stylelint reporter that will output these key-value pairs for each component?

bin/custom-properties-spell-check.js
bjankord: Yeah, not sure why these changed, need to revert the read-write privs on these files.

emilyrohrbough:
If we don't think these could be pushed to our stylelint-configuration / rules, should we push these to our webpack configuration ?

bjankord:
> Is it possible to push the spell check script to the stylelint-config-terra module to check spelling while we verify correct css-variable names?

I'll take a look at adding the spell check part to stylelint-config-terra

scripts/custom-properties/collect-custom-properties.js
```diff
@@ -0,0 +1,68 @@
...
+          // Remove everything from var( to beginning of string
+          // eslint-disable-next-line prefer-destructuring
+          trimmedCssValue = trimmedCssValue.split('var(')[1];
+
+          // Split property and value into an array
+          const cleanedCssValue = trimmedCssValue.split(/,(.+)/);
+
+          // If theme variable has no defined value
+          if (cleanedCssValue[1] === undefined) {
+            customProperties.push([cleanedCssValue[0], 'No defined value']);
```
JakeLaCombe, line 56: would it be better to use empty string, null, or undefined here? I think that would make outputting it to the file easier to work with, as it would play with browsers and default settings in an easier fashion.

JakeLaCombe (approved):
Looks good from my end.

JakeLaCombe (approved):

bjankord:
Closing this based on team sync we had today. More info here: https://github.com/cerner/terra-ui/issues/140

PR #241, opened 2019-01-28
Add Logger for Formatted / Colored Console Output

emilyrohrbough:
### Summary
Add toolkit logger such that console output  is formatted & colored for toolkit services and scripts to make reading messages easier. My terminal has custom colors, but here is the output:

screen shot 2019-01-23 at 10 30 21 am


bjankord (dismissed):

src/logger.js
```diff
@@ -0,0 +1,39 @@
...
+  console.log(getMessage(details, 'grey'))
+);
+
+const consoleWarn = details => (
+  /* eslint-disable-next-line no-console */
+  console.warn(getMessage(details, 'magenta'))
+);
+
+module.exports = {
+  consoleError, consoleLog, consoleWarn, chalk,
```
StephenEsser: What are your thoughts on converting this logger to a static class?

#### Class
```
class Logger {
  static warn() {

  }

  static error() {

  }
}
```

#### Usage
```
const Logger = require('logger');
.
.
.
.
Logger.warn('This is a warning message');
.
.
.
Logger.log('This is a log message');

```

StephenEsser: It may also be beneficial to abstract chalk in some way. Importing and exporting a dependency directly feels a little strange. If we want to keep chalk in this file we may be able to wrap it.

#### Class
```
const chalk = require('chalk');

class Logger {
  // Code
}

Logger.chalk = chalk;

module.exports = Logger;
```

#### Usage
```
const Logger = require('logger');
.
.
.
.
Logger.chalk.cyan('This is a warning message');
```

StephenEsser: This is overkill but if you want to be really fancy you can proxy the chalk methods onto the Logger. ( No IE Support but I don't think any of these are run in the browser )

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy

```
const chalk = require('chalk');

class Logger {
  static log(message) {
    console.log(message);
  }
}

const handler = {
  get: (object, property) => {
    // May want to reverse the order of lookup so Logger methods take precedence
    if (chalk[property]) {
      return chalk[property];
    }
    return Logger[property];
  },
};

module.exports = new Proxy(Logger, handler);
```

This applies all the chalk methods onto the logger.

```
const Logger = require('./logger');
.
.
.
Logger.cyan('This is a log message');
```

It's a little dangerous to have an unknown/pass through API, we could whitelist methods to be safer, but I thought this was really interesting either way.

emilyrohrbough: I like all of these suggestions. will look into them!

emilyrohrbough:
@StephenEsser Made some changes based on your suggestions. Let me know if this aligns with what you were thinking.

StephenEsser:
The new syntax for using the Logger looks really clean. One thing we might look into is a consistent API. It looks like the Logger flexes between `Logger.log(string)` and `Logger.log(object)`. It might be cleaner to make the API take options as a second parameter. Something like `Logger.log(message<String>, options<Object>)`

```js
Logger.warn('This is a warning', { context: 'Context' });
```

Or if we're confident there won't ever be additional options other than context it can just be the second parameter.

```js
Logger.warn('This is a warning', 'Context');
``` 

I learn towards object so the API can expand passively is we're unsure of whether or not more options will be added.

bjankord:
+1 to `(string, object)` for Logger aurguments API

tbiethman (dismissed):

src/logger.js
```diff
@@ -0,0 +1,48 @@
+const chalk = require('chalk');
```
mjhenkes: I'm assuming this file is only ever run through node. if thats the case we could move it to the scripts folder fo clean up most includes.

CHANGELOG.md
```diff
@@ -5,10 +5,11 @@ Unreleased
 ----------
 ### Added
 * Added format option to aggregate-translations to output modern syntax
-* Added the Duplicate Package Checker Plugin to the default webpack config. The duplicate package checker plugin will warn if duplicate packages (different versions) are included in your webpack bundle.
+* Added formatted & colored console output for toolkit services and scripts
 
 ### Changed
 * Update the `connectionRetryCount` to 1 in the default wdio configuraiton
+* Added the Duplicate Package Checker Plugin to the default webpack config. The duplicate package checker plugin will warn if duplicate packages (different versions) are included in your webpack bundle.
```
StephenEsser: Why did this entry move sections?

scripts/utils/logger.js
```diff
@@ -0,0 +1,48 @@
+const chalk = require('chalk');
```
StephenEsser, line 1: I had a few ideas that were a little hard to explain. I've made a gist with what I had in mind. 

I think the formatting of the logger can expand to allow all of the normal chalk options. Example: `chalk.bold.underline.red('Message')` 

The main changes are combining getMessage, setContext, getDetails into a single API.

```js
/**
 * Formats a message.
 * @param {string} message - The message to format.
 * @param {Object} options - The message options.
 * @param {Object} decorations - The decoration options (bold, underline, color, etc...).
 * @returns - A formatted message.
 */
static format(message, options, decorations) {
  // Code
}
```

And adding a decorate function:

```js
/**
 * Decorates a message with the provided options.
 * @param {string} message - The message to decorate.
 * @param {Object} decorations - The decoration options.
 */
static decorate(message, decorations) {
  // Code
}
```
https://gist.github.com/StephenEsser/bc6b7ff1d2f9b890797030528048410e

StephenEsser: This is missing the emphasis function, but let me know what your thoughts are.

emilyrohrbough: For anyone curious, there is conversation on the gist Stephen provided & on here:  https://gist.github.com/emilyrohrbough/665d93aff112323110fda2fbc255d019

@bjankord @tbiethman  @mjhenkes  I would be curious which direction you would lean towards.

emilyrohrbough:
Adding DO_NOT_MERGE label for now till we land on final API

emilyrohrbough:
This is ready for final review @cerner/terra

bjankord (approved):

scripts/utils/logger.js
```diff
@@ -0,0 +1,95 @@
...
+   * @param {string[]} options.decorations - The chalk decoration options (bold, underline, color, etc...).
+   */
+  static decorateText(text, decorations = []) {
+    if (!text) {
+      return '';
+    }
+
+    // Apply all chalk options to the message. Bold, Color, Underline, etc...
+    const decoratedText = decorations.reduce((acc, option) => {
+      if (chalk[option]) {
```
StephenEsser: Because this is now a direct pass through I think we should apply all options without the check or log a message for the invalid option. A bad option shouldn't be silent.

scripts/utils/logger.js
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Terra logger for consistently formatted and decorated log, warning and error messages.
+ * Uses `chalk` module for decoration and gives direct access to chalk for complex message
+ * styling. See chalk's docs for decoration options: https://github.com/chalk/chalk.
+ */
+class Logger {
+  /**
+   * Decorates a message with the provided options.
+   * @param {string} message - The message to decorate.
```
StephenEsser: Parameter name doesn't match the name used below.

scripts/utils/logger.js
```diff
@@ -0,0 +1,95 @@
...
+  static format(message, options) {
+    const { context, decorations } = options;
+
+    const decoratedMessage = Logger.decorateText(message, decorations);
+
+    if (context) {
+      const decoratedContext = Logger.decorateText(context, ['bold', 'dim']);
+      return `${decoratedContext} ${decoratedMessage}`.trim();
+    }
+    return decoratedMessage.trim();
```
StephenEsser: The trim is no longer necessary with the split logic block. 

```js
const decoratedMessage = Logger.decorateText(message, decorations);
const decoratedContext = Logger.decorateText(context, ['bold', 'dim']);

// Trim was being used to remove the empty space if decoratedContext was empty
return `${decoratedContext} ${decoratedMessage}`.trim();
```

scripts/utils/logger.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /**
+   * Logs a message with grey decoration by default.
```
StephenEsser: I would avoid specifying the colors in the documentation. The colors should be abstracted.

scripts/utils/logger.js
emilyrohrbough: Ah I didn't realize it trim the beginning and end. I thought it was just the end.

StephenEsser (approved):

PR #252, opened 2019-02-18
Theme Aggregation

StephenEsser:
### Summary
This pull request introduces an opt-in concept of theme aggregation. The theme aggregation script recursively searches all nested dependency theme files and aggregates them into a single file that automatically gets included in the webpack bundle. 

By default, terra-toolkit's webpack config will automatically run the aggregation script if a `theme.config.js` file is found.

Theme files are expected to be in the following directory format:

```
project
└── themes
    ├── terra-dark-theme
    │   ├── component.scss
    │   ├── root-theme.scss
    │   └── scoped-theme.scss
    └── terra-light-theme
        ├── component.scss
        ├── root-theme.scss
        └── scoped-theme.scss
```

If a consumer opts into theme aggregation the script will output an `aggregated-themes.scss` file. This file will automatically be included into their application if they are using the terra-toolkit webpack config. This file is autogenerated and should not be checked into source control. 

### Additional Details
- Scoped theme aggregation will be added on this pull request.
- A setup guide for teams wanting to use their own webpack config will be added on this pull request or a future pull request. 

@cerner/terra


config/webpack/webpack.config.js
```diff
@@ -152,7 +154,9 @@ const defaultWebpackConfig = (env = {}, argv = {}) => {
     resolveModules.unshift(path.resolve(rootPath, 'aggregated-translations'));
   }
 
-  const options = { rootPath, resolveModules };
+  const themeFile = aggregateThemes();
```
emilyrohrbough: We should provide an option to prevent automatic theme aggregation. It seems like this should be opt-in by default. Especially if this requires changes to repos to ignore the generated files

scripts/aggregate-themes/disclaimer.txt
```diff
@@ -0,0 +1,7 @@
+/*
+ * Warning - Do not modify this file.
+ *
+ * This file is autogenerated and should not be checked into source control.
+ *
+ * It is recommended to add this file to your .gitignore and .eslintignore.
```
emilyrohrbough: also to `.dockerignore` and `.npmignore`

scripts/aggregate-themes/aggregate-themes.js
```diff
@@ -0,0 +1,150 @@
...
+ */
+class ThemeAggregator {
+  /**
+   * Aggregates theme assets.
+   * @returns {string|null} - The output path of the aggregated theme file. Null if not generated.
+   */
+  static aggregate() {
+    const defaultConfig = path.resolve(process.cwd(), CONFIG);
+
+    if (fs.existsSync(defaultConfig)) {
```
emilyrohrbough: Do we need to double-check this is not a directory? This is what we've done for loading i18n config and translation files

scripts/aggregate-themes/aggregate-themes.js
```diff
@@ -0,0 +1,150 @@
...
+  static aggregateTheme(options = {}) {
+    ThemeAggregator.validate(options);
+
+    const { theme } = options;
+
+    const dirs = ThemeAggregator.find(`**/themes/${theme}/`, options);
+    const assets = ThemeAggregator.filter(dirs, options);
+
+    if (fs.existsSync(`${NODE_MODULES}${theme}/`)) {
+      assets.unshift(theme);
```
emilyrohrbough: why are we removing the `node_modules/theme` from the assets list?

scripts/aggregate-themes/aggregate-themes.js
```diff
@@ -0,0 +1,150 @@
...
+   * Aggregates theme assets.
+   * @param {Object} options - The aggregation options.
+   * @returns {string} - The output path of the aggregated theme file.
+   */
+  static aggregateTheme(options = {}) {
+    ThemeAggregator.validate(options);
+
+    const { theme } = options;
+
+    const dirs = ThemeAggregator.find(`**/themes/${theme}/`, options);
```
emilyrohrbough: I think I'm a little lost. From what I can tell, we are filtering for files and directories it looks like initially, then we are filtering each returned dir/file to find `.css` files? We can pass `noDir` to glob to ensure we never return a directory.

It might be easier to match on these patters so you wouldn't need to filter on directories 
```
const rootStyles = ThemeAggregator.find(`**/themes/${theme}/root-theme.scss`, options);
const otherStyles = ThemeAggregator.find(`**/themes/${theme}/*.scss`, { exclude: options.exclude.push('root-theme.scss') });
```
I think you'd be able to drop the dirFilter then

Then in filter method, for each  path/root-theme.scss could remove path/*.scss options from otherStyles list

scripts/aggregate-themes/aggregate-themes.js
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * Finds files and directories matching a pattern.
+   * @param {string} pattern - A regex pattern.
+   * @param {Object} options - The aggregation options.
+   * @returns {string[]} - An array of matching files and directories.
+   */
+  static find(pattern, options) {
+    const { exclude = [] } = options;
+
+    return glob.sync(pattern, { ignore: exclude });
```
emilyrohrbough: It'd be worth considering adding `follow` key to the glob options. This will allow the search patterns to follow symlinked folders

emilyrohrbough (commented):
IMO we should add all supporting documentation in this PR. It seems like when we wait its easy to push off. Also are there any tests for this?

config/webpack/webpack.config.js
StephenEsser: This functionality is opt in by default. The script will exit if there is not a `theme.config.js` file. By adding the file you are opting into theme aggregation.

scripts/aggregate-themes/aggregate-themes.js
StephenEsser: The config file is required to be at the same level as the process being run. With the current implementation it can't be a directory. 

```
project
├── theme.config.js
├── webpack.config.js
```

scripts/aggregate-themes/aggregate-themes.js
StephenEsser: We're not removing it from the list. Unshift appends it to the beginning of the list.

scripts/aggregate-themes/aggregate-themes.js
StephenEsser: We are finding all theme directories and then filtering the contents of the directories. Not all theme directories will have a root-theme.scss file. If there is not a root-theme we add all the files from the directory. 

```
project
└── themes
    ├── terra-dark-theme
    │   ├── component.scss
    │   ├── root-theme.scss
    │   └── scoped-theme.scss
    └── terra-light-theme
        ├── component.scss // There is only a single file, there isn't a root-theme.scss
```

We only use the root-theme if it exists, otherwise we aggregate all available scss files in that directory.

docs/ThemeAggregation.md
```diff
@@ -0,0 +1,64 @@
+# Theme Aggregation
+
+Terra Toolkit provides a built-in mechanism for aggregating themes.
+
+By default, the terra-toolkit webpack configuration enables theme aggregation when it detects the presence of a `theme.config.js` file.
```
JakeLaCombe: It still feels like there should be a flag you set in your webpack configuration that determines if themes should be aggregated or not. What if for some reason they already have a theme.config.js which does something else?

If you do need to do the approach of having a file present, could we namespace the file (terra_theme.config.js)?

scripts/aggregate-themes/aggregate-themes.js
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Filters theme assets.
+   * @param {string[]} patterns - An array of directory and file paths.
+   * @param {Object} options - The aggregation options.
+   * @returns {string[]} - An array of file names.
+   */
+  static filter(patterns, options) {
+    const assets = [];
+    patterns.forEach((asset) => {
```
JakeLaCombe: Just out of curiousity. It feels like we could just do a patterns.map here?

docs/ThemeAggregation.md
StephenEsser: Is there a scenario you can think of where a user would have an intentional theme aggregation config file and not want to aggregate themes?

We could definitely look into name-spacing the file. It just feels a little redundant to have a configuration file and then also be required to opt into using it.

JakeLaCombe: We wouldn't be the first one to do that. I know there are several PostCSS plugins that either allow you to define their configuration inside the webpack file or inside a separate file itself.

Nevertherless, I think having a namespaced config file is appropriate.

scripts/aggregate-themes/aggregate-themes.js
StephenEsser: The directory filter returns an array that we need to flatten. Map would potentially result in an array of arrays.

JakeLaCombe: Thanks! I was looking at this, and felt like I was missing something I couldn't think of.

docs/ThemeAggregation.md
StephenEsser: Do those configurations require you to do both? Declare the file and also provide a webpack configuration option to use that file? 

I'm not sure we should enforce a `terra` specific namespace on project files, but we could do something like `theme-aggregation.config.js` or something similar. It's a little wordy but I wouldn't expect there to be any collisions.

JakeLaCombe: I think the PostCssAssetsPlugin either wants the config inside the webpack file, or a configuration file in your repo, and then in the webpack config, you provide a path to the configuration file.

JakeLaCombe: theme-aggregation is fine as well.

scripts/aggregate-themes/aggregate-themes.js
emilyrohrbough: 🤦‍♀️

JakeLaCombe (approved):

mjhenkes (approved):

ryanthemanuel (approved):

docs/ThemeAggregation.md
StephenEsser: The configuration file has been updated to be `terra-theme.config.js` https://github.com/cerner/terra-toolkit/pull/252/commits/6ae2c94b5d48b535ead8ee6753bc7ea47f107578

scripts/aggregate-themes/disclaimer.txt
StephenEsser: These have been added: https://github.com/cerner/terra-toolkit/pull/252/commits/6ae2c94b5d48b535ead8ee6753bc7ea47f107578

docs/ThemeAggregation.md
```diff
@@ -0,0 +1,82 @@
...
+@import 'component-1.scss';
+@import 'component-2.scss';
+```
+
+### Scoped Theme Example
+
+The `scoped-theme.scss` file should include a selector to enable scoping.
+
+```scss
+$selector: '.terra-dark-theme';
```
bjankord: The `$selector` variable is an implementation detail and assumes you have a Sass mixin within the component theme file that either outputs `:root` or the string defined in the `$selector`.

Adding a selector variable in this file won't make this a scoped theme unless the imported theme code uses a mixin to swap its contents between `:root` selector and `$selector` as a selector.

Should we document this? I'm thinking we should but this doesn't seem like the right place as it seems more that we need documentation on how to create a custom theme and that would be more fitting.

scripts/aggregate-themes/disclaimer.txt
```diff
@@ -0,0 +1,11 @@
+/*
```
bjankord, line 1: What is this file intended for?

bjankord (approved):

scripts/aggregate-themes/disclaimer.txt
StephenEsser: This is a disclaimer to the generated aggregation file to help communicate the file is auto generated and shouldn't be check into git. I added it in a separate txt file so it was easier to read and reason about. 

The file that gets outputted will have this prefixed to it.

docs/ThemeAggregation.md
StephenEsser: I agree, we need some documentation on how to property create theme files. Any ideas where that should live? 

I'll remove this documentation for now so it doesn't confuse anyone.

bjankord: I think we should add something to terra-ui.

PR #253, opened 2019-02-18
Offer screenshots and accessibility in one command

### LINKED ISSUE #249: Perform accessibility tests when taking screenshots
# Feature Request

## Description

In order to better support and require accessibility, we will create a mechanism to perform accessibility testing whenever a screenshot is captured.

## Design
After discussing these changes with a few people there are a few assumptions I want to make about how this will work.

- This will be a new helper method that wraps both screenshot comparisons and accessibility.  Unfortunately this will mean that teams won't get this for free but this will allow us to make the change non-passively and without any additional configuration.
- Multiple form factors *will not* be supported with a single call to this new method helper.  This is the direction that we want to go with our tests.  It is more efficient and in the cases where we can parallelize all of our form factors, this way of executing tests is not compatible.  
- Selectors *will* be supported.  Any selector being used for screenshots will also be used for accessibility.
- Other configurations will be supported as well.  The method signature will be the same as the existing screenshot method except axe options will now be accepted in the options hash.

ryanthemanuel:
### Summary

In order to better support and require accessibility, we will create a mechanism to perform accessibility testing whenever a screenshot is captured.

Fixes #249


ryanthemanuel:
I still have to write more tests and doc, but wanted to get people's thoughts.

tests/wdio/matchScreenshotAndTestStandards-spec.js
```diff
@@ -0,0 +1,38 @@
...
+
+    Terra.should.matchScreenshotAndTestStandards('test name');
+
+    expect(accessibilityCalled).to.equal(true);
+    expect(matchScreenshotImplementationName).to.equal('test name');
+
+    accessibility.beAccessible = oldAccessible;
+    visualRegressions.matchScreenshotImplementation = oldMatchScreenshotImplementation;
+  });
+});
```
emilyrohrbough: Should we add a check for passing in selector?

src/wdio/services/TerraCommands/match-screenshot-and-test-standards.js
```diff
@@ -0,0 +1,43 @@
...
+
+  return {
+    name,
+    selector,
+    misMatchTolerance,
+    axeOptions: { ...options.axeRules, context: selector },
+  };
+};
+
+const matchScreenshotAndTestStandards = (...args) => {
```
emilyrohrbough: We should add inline comment on the possible options

ryanthemanuel: Yup.  Was going to add some docs in a bit.

tests/wdio/matchScreenshotAndTestStandards-spec.js
ryanthemanuel: I need to beef up these tests.

src/wdio/services/TerraCommands/visual-regression.js
```diff
@@ -96,24 +96,13 @@ const getTestDescription = matchType => (
...
-* either within the mismatch tolerance or are an exact match.
-* @property {Array} testArguments - The test arguments passed to the `matchScreenshotWithinTolerance`
-*    or `matchScreenshotExactly` methods.
-* @property {String} matchType - Specifies the type of matchReference assertion. Either 'withinTolerance'
-*    or 'exactly'.
-*/
-const matchScreenshot = (testArguments, matchType) => {
-  const { name, selector, options } = determineScreenshotOptions(...testArguments);
-
+const matchScreenshotImplementation = (name, matchType, selector, options) => {
```
emilyrohrbough: Some of this is getting a little layered. I might be easier to understand if we can someone associate this name with the `it` block or the fact its the VR / screenshot comparison test block

emilyrohrbough: Also we should document this method

tests/wdio/matchScreenshotAndTestStandards-spec.js
```diff
@@ -0,0 +1,38 @@
+/* global describe, before, browser, it, expect, Terra */
```
emilyrohrbough: If we are using our eslint-configuration, we shouldn't need to add these

src/wdio/services/TerraCommands/visual-regression.js
ryanthemanuel: Good call.  I wasn't sure what to call this.  I like referencing that this is the actual it block.

tests/wdio/matchScreenshotAndTestStandards-spec.js
ryanthemanuel: Yeah it's on my list to figure out why that's not working.

tests/wdio/matchScreenshotAndTestStandards-spec.js
```diff
@@ -0,0 +1,38 @@
...
+const visualRegressions = require('../../lib/wdio/services/TerraCommands/visual-regression').default;
+
+describe('matchScreenshotAndTestStandards', () => {
+  describe('full implementation', () => {
+    before(() => {
+      browser.url('/compare.html');
+      browser.setViewportSize(Terra.viewports('tiny')[0]);
+    });
+
+    Terra.should.matchScreenshotAndTestStandards();
```
emilyrohrbough: I don't have a good suggestion right now, but I think it would be worth thinking of a shorter name for this....

emilyrohrbough (commented):
We will want to add documentation to deprecate or suggest against using the matchScreenshot's direct usage. We also needs docs to explain this new Terra service helper and options

tests/wdio/matchScreenshotAndTestStandards-spec.js
emilyrohrbough: `Terra.should`....
- `passAudit()`
- `auditElement()`
- `meetCompliance()`
- `meetStandards()`
- `testStandards()`

 not sure if I'm the biggest fan of standards or any of these words because this implies more than we currently are doing, but then it leaves it open for more later on

src/wdio/services/TerraCommands/match-screenshot-and-test-standards.js
```diff
@@ -0,0 +1,43 @@
...
+    options = typeof param2 === 'object' && !Array.isArray(param2) ? param2 : options;
+  } else {
+    options = typeof param1 === 'object' && !Array.isArray(param1) ? param1 : options;
+  }
+
+  // Check if custom selector should be used, otherwise use the global value.
+  const selector = options.selector || global.browser.options.terra.selector;
+
+  // Check if custom misMatchTolerance should be used, otherwise use the global value.
+  const misMatchTolerance = options.misMatchTolerance || global.browser.options.visualRegression.compare.misMatchTolerance;
```
emilyrohrbough: This is a new piece. Since this is the direction we are pushing people, should we place a misMatchTolerance maximum on screenshot comparison?

tests/wdio/matchScreenshotAndTestStandards-spec.js
ryanthemanuel: Fixed.

src/wdio/services/TerraCommands/match-screenshot-and-test-standards.js
```diff
@@ -0,0 +1,43 @@
+import accessibilityMethods from './accessiblity';
+import visualRegressionMethods from './visual-regression';
+
+const determineOptions = (...args) => {
+  const param1 = args.length ? args[0] : undefined;
+  const param2 = args.length > 1 ? args[1] : undefined;
```
JakeLaCombe: Can we add documentation on what we expect the parameters to be before the function declaration?

tests/wdio/matchScreenshotAndTestStandards-spec.js
JakeLaCombe: matchAccessibleScreenshot maybe?

JakeLaCombe:
Overall, I think this feature is a great addition to toolkit. Hopefully we can get other teams around Cerner to adopt it quickly.

package.json
```diff
@@ -31,9 +31,7 @@
     "tt-serve-static": "bin/serve-static.js",
     "tt-wdio": "bin/wdio-runner.js"
   },
-  "browserslist": [
-    "extends browserslist-config-terra"
-  ],
+  "browserslist": [],
```
emilyrohrbough: why did we drop the browserslist-config?

src/wdio/services/ServeStaticService.js
```diff
@@ -7,8 +7,8 @@ const context = '[Terra-Toolkit:serve-static-service]';
 
 export default class ServeStaticService {
   async onPrepare(config = {}) {
-    const site = config.site;
-    const webpackConfig = config.webpackConfig;
+    const { site } = config;
```
emilyrohrbough: This could be reduced to 
const { site, webpackConfig } = config;

package.json
ryanthemanuel: eslint was complaining about promises.  Rather than adding the polyfill config I preferred removing the browser list since this is a pure node project.

src/wdio/services/ServeStaticService.js
emilyrohrbough: same with locale further down

tests/wdio/matchScreenshotAndTestStandards-spec.js
ryanthemanuel: I like validateElement maybe?  Means that it's validating a given element in terms of screenshot and anything else we want like accessibility.

emilyrohrbough: I'm good with validateElement!

package.json
```diff
@@ -31,9 +31,16 @@
     "tt-serve-static": "bin/serve-static.js",
     "tt-wdio": "bin/wdio-runner.js"
   },
-  "browserslist": [],
+  "browserslist": [
+    "extends browserslist-config-terra"
+  ],
```
emilyrohrbough: Hmm you know, we might not even need this configuration here. Everything within this module is not specific to web development, but are just tools for developing web components. We pass browserslist to the webpack for testing

src/wdio/services/TerraCommands/match-screenshot-and-test-standards.js
ryanthemanuel: What are people's thoughts on this?  I'm not sure what we'd want the max to be.  Any thoughts?

ryanthemanuel:
> We will want to add documentation to deprecate or suggest against using the matchScreenshot's direct usage. We also needs docs to explain this new Terra service helper and options

Done.

ryanthemanuel:
For reference, here are the changes that would need to be made in terra-clinical to consume this: https://github.com/cerner/terra-clinical/pull/469

src/wdio/services/TerraCommands/validate-element.js
```diff
@@ -0,0 +1,59 @@
...
+*/
+const determineOptions = (...args) => {
+  const param1 = args.length ? args[0] : undefined;
+  const param2 = args.length > 1 ? args[1] : undefined;
+
+  let name = 'default';
+  let options = {};
+  if (typeof param1 === 'string') {
+    name = param1;
+    options = typeof param2 === 'object' && !Array.isArray(param2) ? param2 : options;
```
yuderekyu, line 16: i was going to say we could dry this logic up, but if we plan on deprecating the visual regression service this is fine

config/wdio/visualRegressionConf.js
```diff
@@ -70,10 +70,10 @@ function getScreenshotPath(ref) {
 
     const baseDir = global.browser.options.baseScreenshotDir;
     if (baseDir) {
-      testPath = testPath.split(process.cwd())[1];
+      [, testPath] = testPath.split(process.cwd());
```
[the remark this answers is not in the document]
ryanthemanuel: Linter wanted this :)

src/wdio/services/TerraCommands/validate-element.js
ryanthemanuel: I'd like to deprecate it at some point.

src/wdio/services/TerraCommands/match-screenshot-and-test-standards.js
bjankord: A diff more the 1.0 is usually noticeable where I've found diffs below 1 to not be noticeable unless you specifically know what you're looking for.

bjankord (approved):
+1, nice work on this @ryanthemanuel!

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Unreleased
 * Updated aggregate translations script fallback to base locale if translation is missing for regional locale on a string-by-string basis.
 
 ### Added
+* Added a new method `validateElement` to both do a screenshot comparison and test accessibility
```
emilyrohrbough: Whoops, this needs to be moved to unreleased

config/wdio/visualRegressionConf.js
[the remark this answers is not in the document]
emilyrohrbough: yeah, I had disabled this eslint rule previously since we do so much with objects in the repo. https://github.com/cerner/terra-toolkit/pull/253/files#diff-1dc6ee56b778cd91e0327b52aaeaa1b9L13

docs/Wdio_Utility.md
```diff
@@ -70,14 +70,21 @@ Then, to assist with testing, the TerraService provides the Terra global helper
     - By default returns all viewports if not name key are provided.
 - `Terra.should.beAccessible()` mocha-chai convenience method that runs an axe test for the page. Takes the same arguments as the `axe()` utility.
     - See [beAccessible-spec.js](https://github.com/cerner/terra-toolkit/blob/master/tests/wdio/beAccessible-spec.js) for examples.
-- `Terra.should.matchScreenshot()` mocha-chai convenience method that takes a screenshot for the specified viewports and verifies the images are within the specified mis-match tolerance. Note: this method provides its own mocha it test case. The methods accepts these arguments (in this order):
+- `Terra.should.matchScreenshot()` **Note - It is preferred to use Terra.should.validateElement().  Terra.should.matchScreenshot() may eventually be deprecated** mocha-chai convenience method that takes a screenshot for the specified viewports and verifies the images are within the specified mis-match tolerance. Note: this method provides its own mocha it test case. The methods accepts these arguments (in this order):
```
emilyrohrbough, line 73: > Terra.should.matchScreenshot() may eventually be deprecated

We talked about not being certain if this will still be the case. Should we keep this message?

src/wdio/services/TerraCommands/validate-element.js
```diff
@@ -0,0 +1,59 @@
...
+  const misMatchTolerance = options.misMatchTolerance || global.browser.options.visualRegression.compare.misMatchTolerance;
+
+  return {
+    name,
+    selector,
+    misMatchTolerance,
+    axeOptions: {
+      ...options.axeRules && { rules: options.axeRules },
+      restoreScroll: true,
+      ...selector && { context: selector },
```
emilyrohrbough: this should alway be true since we fallback to the global selector above

tests/wdio/validateElement-spec.js
```diff
@@ -0,0 +1,77 @@
...
+  describe('full implementation', () => {
+    before(() => {
+      browser.url('/compare.html');
+      browser.setViewportSize(Terra.viewports('tiny')[0]);
+    });
+
+    Terra.should.validateElement();
+  });
+
+  it('calls the appropriate methods downstream with all arguments', () => {
```
emilyrohrbough: I feel like these last two tests would be better suited as jest tests that mock the accessibility and visual regression methods

emilyrohrbough (approved):

JakeLaCombe (approved):

BenBoersma (approved):

mjhenkes (approved):

docs/Wdio_Utility.md
```diff
@@ -144,7 +151,7 @@ describe('Advanced Test', () => {
 ```
 
 ### Testing multiple viewports.
-Sometimes its necessary to rerun the test steps in each viewport. To do this, `Terra.viewports` can be used to wrap the `describe` block. Example:
+Sometimes its necessary to rerun the test steps in each viewport. Some build systems support running viewports in parallel and control the viewport via the FORM_FACTOR environment variable. In the open source world which uses travis and doesn't support that method, `Terra.viewports` can be used to wrap the `describe` block. Example:
```
tbiethman: Can we change 'In the open source world which uses travis and doesn't support that method,' to 'For build systems that don't support parallelization,'

PR #256, opened 2019-02-25
Regional locale translation fallback

### LINKED ISSUE #219: Add capability to help detect missing translations
# Feature Request

## Description
With the plan to drop Arabic from being a supported Terra UI locale, it's been requested we help provide the ability to easily detect when translations are missing. I think this would be beneficial for any teams adding locales, e.g. your app only supports english and your adding spanish locales.

I'd like to use this issue to collect ideas on what we want this feature to do.

## @ Mentions
@ryanthemanuel @emilyrohrbough @mhemesath @nickpith

jmsv6d:
### Summary
Allows regional locale translations to fallback to base locale translation on a string-by-string basis. Will also log warning when we do this. Fixes #219 

### Additional Details

@cerner/terra


scripts/aggregate-translations/write-aggregated-translations.js
```diff
@@ -12,7 +12,23 @@ const writeAggregatedTranslations = (aggregatedMessages, locales, fileSystem, ou
       }
 
       const translationFilePath = path.resolve(outputDir, `${locale}.js`);
-      fileSystem.writeFileSync(translationFilePath, generateTranslationFile(locale, aggregatedMessages[locale]));
+      const messages = aggregatedMessages[locale];
+      const localeRegionSplit = locale.split('-');
```
emilyrohrbough: It might make this a bit clean if we did something like:

```js
const [baseLocale, region] = locale.split('-');
if (region) {
...
}
```

scripts/aggregate-translations/write-aggregated-translations.js
```diff
@@ -12,7 +12,23 @@ const writeAggregatedTranslations = (aggregatedMessages, locales, fileSystem, ou
...
+      let mergedMessages = messages;
+
+      // Check if this is a regional locale
+      if (localeRegionSplit.length > 1) {
+        const localeMessages = aggregatedMessages[localeRegionSplit[0]];
+        if (localeMessages) {
+          Object.keys(localeMessages).forEach((key) => {
+            if (!messages[key]) {
+              Logger.warn(`${locale} translations missing for ${key}, ${localeRegionSplit[0]} translation string will be used instead.`);
```
emilyrohrbough: translations should be singular

scripts/aggregate-translations/write-aggregated-translations.js
```diff
@@ -12,7 +12,23 @@ const writeAggregatedTranslations = (aggregatedMessages, locales, fileSystem, ou
...
 
       const translationFilePath = path.resolve(outputDir, `${locale}.js`);
-      fileSystem.writeFileSync(translationFilePath, generateTranslationFile(locale, aggregatedMessages[locale]));
+      const messages = aggregatedMessages[locale];
+      const localeRegionSplit = locale.split('-');
+      let mergedMessages = messages;
+
+      // Check if this is a regional locale
+      if (localeRegionSplit.length > 1) {
+        const localeMessages = aggregatedMessages[localeRegionSplit[0]];
```
emilyrohrbough: nit: lets call this baseLocaleMessages for clarity

Then can we add an inline comment specifying we are merging on the base locale messages?

scripts/aggregate-translations/write-aggregated-translations.js
```diff
@@ -12,7 +12,23 @@ const writeAggregatedTranslations = (aggregatedMessages, locales, fileSystem, ou
...
+        if (localeMessages) {
```
emilyrohrbough: If this searches `en-AU`, then `en` will this provided `en` fallback or do we need to sort the locale list before aggregating to ensure the base locale messages have been found?

jmsv6d: All the translation aggregating happens before this function is called, so there shouldn't be an issue.

emilyrohrbough (approved):

bjankord (approved):

ryanthemanuel (approved):

PR #266, opened 2019-03-21
Updated the devtool to use eval-source-map

### LINKED ISSUE #258: devtool in development should produce original source sourcemaps
# Feature Request

## Description

Currently, devtool is set to 'cheap-source-map' which produces a SourceMap without column-mappings ignoring loader SourceMaps. According to the documentation, this devtool option is only used in special cases [1]. Webpack states the best quality SourceMap for dev is 'eval-source-map' which produces original source SourceMaps.

[1] https://webpack.js.org/configuration/devtool/#special-cases
[2] https://webpack.js.org/configuration/devtool/#development

lalitha94:
Changed the devtool SourceMap from 'cheap-source-map' to 'eval-source-map' to address the [issue](https://github.com/cerner/terra-toolkit/issues/258)

Resolves #258


mjhenkes:
Does this cause any increase in build time? Could you provide a comparison? With something like terra-core.

lalitha94:
> Does this cause any increase in build time? Could you provide a comparison? With something like terra-core.

The [documentation](cheap-source-map) says cheap-source-map provides better performance/less build time

mjhenkes:
Could you time the terra-core build with this enabled before and after so we can get a real metric?

lalitha94:
> Could you time the terra-core build with this enabled before and after so we can get a real metric?

It gives a difference of around 2 seconds.

mjhenkes:
Hi @lalitha94, 
Let me give you some background on what I'm looking for. This webpack config is shared amongst many projects at cerner most of which are built using our CI system. Changes to this config will have an effect on our capacity and how long teams wait for builds both in the CI system and locally developing. Before we pull in this change (which looks great) I'd like to have an idea of what to expect for users's builds. I know we didn't include this data in my original pull request and change to webpack config, but we probably should have. We're growing as a platform and affecting more teams all the time.

That said I'm looking for timing numbers on one of our larger repos to get a baseline for what users can expect. Looking at the documentation [here](https://webpack.js.org/configuration/devtool/). It looks like eval-source-map is slow on initial build but fast on subsequent builds where as cheap-source-map is pretty fast on initial build but no faster on subsequent builds.

Could you use the time command (https://linuxize.com/post/linux-time-command/) to wrap this command ```webpack --config webpack.config.js``` in the terra core repo both with and without your change for an initial and subsequent build.

mjhenkes:
Also, you'll have to update this test:
https://github.com/cerner/terra-toolkit/blob/master/tests/jest/webpack.config.test.js#L136

lalitha94:
> Also, you'll have to update this test:
> https://github.com/cerner/terra-toolkit/blob/master/tests/jest/webpack.config.test.js#L136

I already updated the test before the raising the PR :)

mjhenkes:
So you did! I guessed that was why the build was failing. On further investigation, it looks like you just need to add a comment to the change log to get the build to succeed (probably).

bjankord:
I'd be interested to know the compile times on terra-core with these changes as well.

lalitha94:
> Hi @lalitha94,
> Let me give you some background on what I'm looking for. This webpack config is shared amongst many projects at cerner most of which are built using our CI system. Changes to this config will have an effect on our capacity and how long teams wait for builds both in the CI system and locally developing. Before we pull in this change (which looks great) I'd like to have an idea of what to expect for users's builds. I know we didn't include this data in my original pull request and change to webpack config, but we probably should have. We're growing as a platform and affecting more teams all the time.
> 
> That said I'm looking for timing numbers on one of our larger repos to get a baseline for what users can expect. Looking at the documentation [here](https://webpack.js.org/configuration/devtool/). It looks like eval-source-map is slow on initial build but fast on subsequent builds where as cheap-source-map is pretty fast on initial build but no faster on subsequent builds.
> 
> Could you use the time command (https://linuxize.com/post/linux-time-command/) to wrap this command `webpack --config webpack.config.js` in the terra core repo both with and without your change for an initial and subsequent build.

Thanks, for the detailed explanation.
I ran the [compile:build](https://github.com/cerner/terra-core/blob/master/package.json#L36). It took
Eval-source-map:  
1. real    0m52.471s
2. user    1m3.150s
3. sys     0m7.485s 

Cheap-source-map:
1. real    0m50.970s
2. user    1m1.416s
3. sys     0m7.266s

For, re-building I tried to add the watch script(`"watch": "webpack --config webpack.config --watch"`) in terra-core/package.json but it doesn't seem to take it.

mjhenkes:
@lalitha94 thanks for looking into the timings. 
I think all you have left is to add a comment to the change log. Otherwise it looks good.

mjhenkes (approved):

bjankord (approved):

emilyrohrbough (approved):

PR #280, opened 2019-05-06
Script / Dev Dependency Clean Up

emilyrohrbough:
### Summary
Remove `rimraf` and `shelljs` for scripts that can be written and run from the package.json


bjankord (approved):

package.json
```diff
@@ -125,12 +126,12 @@
     "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-env": "^1.7.0",
     "babel-runtime": "^6.26.0",
+    "check-installed-dependencies": "^1.0.0",
```
StephenEsser, line 126: What's this for?

StephenEsser (approved):

package.json
emilyrohrbough: This module will check that the current installed packages match the versions defined in the package.json. This is to ensure the project isn't tested and then released against dependency versions different than defined. We already use this in the mono-repos: https://github.com/cerner/terra-core/blob/master/package.json#L53

The repos description it states this will prevent publish if these don't match: https://www.npmjs.com/package/check-installed-dependencies#usage-in-npm-scripts

PR #298, opened 2019-06-12
Add Firefox and IE Capabilities 

emilyrohrbough:
### Summary
Made changes to enable firefox testing and selenium gird tests for chrome, firefox, and IE. This increased timeouts to account for longer new session requests when a selenium grid queue is large. 

Fixes an issue where serve-static servers an empty build dir and not urls are valid.

Adds default selector to remove the need to sharing terra-dev-site's wdio configuration, making terra-toolkit the only place for wdio config.


scripts/wdio/README.md
```diff
@@ -1,25 +1,14 @@
...
-| **--formFactors**  | `undefined` | The list of viewport sizes to test. |
-| **-locales** | `['en']` | The list of locales to test. |
-| **--continueOnFail** | `undefined` | Whether or not to execute all test runs when a run fails. |
-| **--updateReference** | `undefined` | Whether or not to remove reference screenshots during screenshot cleanup. CLI use only. |
-| **--port**  | `undefined` | [wdio option] The selenium server port. |
-| **--host** | `undefined` | [wdio option] The selenium server host address. |
-| **--baseUrl** | `undefined` | [wdio option] The base URL. |
-| **--suite** | `undefined  ` | [wdio option] The suite to run. |
-| **--spec** | `undefined` | [wdio option] The spec file to run. |
+Run ```tt-wdio --help``` to see the configuration options.
```
yuderekyu, line 11: 👍

yuderekyu (approved):

config/wdio/wdio.conf.js
```diff
@@ -71,30 +82,25 @@ const config = {
...
-
   // Ignore deprecation warnings. When chrome supports /actions API we'll update to use those.
   deprecationWarnings: false,
 
   axe: {
     inject: true,
   },
 
+  terra: {
+    selector: '[data-terra-dev-site-content] *:first-child',
```
mjhenkes, line 93: Part of me wants to change this selector name, but i think that would be counter productive.

scripts/serve/serve-static.js
```diff
@@ -37,10 +37,10 @@ const staticApp = (site, index) => {
 const serve = (options) => {
   const {
     // Setting defaults here instead of in cli.
-    site = '.build/', port = 8080, host = '0.0.0.0', index,
+    site = './build', port = 8080, host = '0.0.0.0', index,
   } = options;
 
-  if (!dirExists(site)) {
+  if (!dirExistsWithContent(site)) {
     return Logger.warn(`Cannot serves content from ${site} because it does not exist.`, { context });
```
mjhenkes: We should update the message to add  ' or is empty'

scripts/serve/serve-static.js
```diff
@@ -9,7 +9,7 @@ const displayServer = (localAddress) => {
   Logger.log(`* Local:            ${Logger.emphasis(localAddress)}`);
 };
 
-const dirExists = site => (fs.existsSync(site) && fs.lstatSync(site).isDirectory());
+const dirExistsWithContent = site => glob.sync(`${site}/**/*`).length > 0;
```
mjhenkes: I think this would be cheaper just to read the directory with fs.readdirsync, the only case it wouldn't handle is if that directory contains empty directories. Is that a case we're worried about?

https://nodejs.org/api/fs.html#fs_fs_readdirsync_path_options

scripts/wdio/getWdioConfigPath.js
```diff
@@ -1,16 +1,14 @@
...
+const { isFile } = require('../../config/configUtils');
 
 const getWdioConfigPath = (configPath) => {
   if (configPath) {
     return path.resolve(configPath);
   }
 
   // Try to find the local to process.cwd webpack config
-  const localConfig = path.resolve(process.cwd(), 'wdio.conf.js');
+  const localConfig = path.resolve(process.cwd(), './wdio.conf.js');
```
mjhenkes: was this working previously?

config/wdio/wdio.conf.js
emilyrohrbough: Yeah, I don't really love this selector name referencing terra-dev-site, but wasn't sure what the best option would be. I think its okay enough since anyone can change it 🤷‍♀

scripts/serve/serve-static.js
emilyrohrbough: I hadn't thought about empty nested dir. I don't think its a much of a concern though since the clean plugin will blow away the build dir. I honestly just didn't spend too much time looking into the issue. Will look into the options/link you suggested

emilyrohrbough: Yes, I think reading the dir content and checking the length is a better option

config/configUtils.js
```diff
@@ -0,0 +1,16 @@
+const fs = require('fs');
+
+const isFile = filePath => (fs.existsSync(filePath) && !fs.lstatSync(filePath).isDirectory());
+
+const dynamicRequire = (filePath) => {
+  if (isFile(filePath)) {
+    // eslint-disable-next-line global-require, import/no-dynamic-require
+    return require(filePath);
+  }
+  return undefined;
```
bjankord, line 10: Should we log something to the console if this condition is hit, more so to make the consumer aware the file they tried to require is not registering as a file.

bjankord (approved):

config/configUtils.js
emilyrohrbough: I could see value in it, but we have created this to attempt to load configuration for people from the root level, if it doesn't exist, they will need to add it themselves by extending the terra config for via the command line but this dynamic require may still run to attempt to find this file which would caused error console message that aren't helpful. IMO keeping this silent is better. Other's thoughts?

scripts/serve/serve-static.js
emilyrohrbough: updated: https://github.com/cerner/terra-toolkit/pull/298/commits/9ff9152b60a139e62cc5c4e31afaddf077668ccf#diff-2d0e39e2b5431dc8d9df655f5fc376fbR44

scripts/serve/serve-static.js
emilyrohrbough: updated: https://github.com/cerner/terra-toolkit/pull/298/commits/9ff9152b60a139e62cc5c4e31afaddf077668ccf#diff-2d0e39e2b5431dc8d9df655f5fc376fbR12

scripts/wdio/getWdioConfigPath.js
emilyrohrbough: idk why I changed this. Reverted: https://github.com/cerner/terra-toolkit/pull/298/commits/9ff9152b60a139e62cc5c4e31afaddf077668ccf#diff-2b14bdc794a09b6c5703f768f8877711R11

mjhenkes (approved):

PR #301, opened 2019-06-18
Basepath config

jessica-bailey:
### Summary

Modify webpack config to define a BASEPATH environment variable. This will be used in the pathHelper function in orion-path-helper-js project to modify hard coded paths in node modules to include the correct base path.

### Additional Details

@cerner/terra

### DIFF
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -9,8 +9,9 @@ const TerserPlugin = require('terser-webpack-plugin');
 const merge = require('webpack-merge');
 const DuplicatePackageCheckerPlugin = require('duplicate-package-checker-webpack-plugin');
 const aggregateTranslations = require('terra-aggregate-translations');
-const ThemeAggregator = require('../../scripts/aggregate-themes/theme-aggregator');
 const webpack = require('webpack');
+const ThemeAggregator = require('../../scripts/aggregate-themes/theme-aggregator');
+
 const base = process.env.BASEPATH;
 
 const webpackConfig = (options, env, argv) => {
@@ -209,8 +210,8 @@ const defaultWebpackConfig = (env = {}, argv = {}) => {
     staticOptions,
   };
   return merge.strategy({
-    entry:'append',
-  })(webpackConfig(options, env, argv), basepathConfig)
+    entry: 'append',
+  })(webpackConfig(options, env, argv), basepathConfig);
 };
 
 module.exports = defaultWebpackConfig;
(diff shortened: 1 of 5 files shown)


config/webpack/webpack.config.js
```diff
@@ -170,6 +172,14 @@ const webpackConfig = (options, env, argv) => {
...
+const basepathConfig = {
+  plugins: [
+    new webpack.DefinePlugin({
+      BASEPATH: JSON.stringify(base),
+    }),
+  ],
+};
+
 const defaultWebpackConfig = (env = {}, argv = {}) => {
   const { disableAggregateTranslations, disableHotReloading } = env;
```
emilyrohrbough, line 185: I'm thinking we should make this change by adding an additional webpack env to follow syntax we've done with other webpack envs to toggle configuration. Something like:

```
const { basepath, disableAggregateTranslations, disableHotReloading } = env;

...

 return webpackConfig(options, env, argv);
```

Then we can compile webpack like:

```bash
webpack --config webpack.config.js -p --env.basepath='mybaseroot/'
```

We should then have access to add this directly to the plugins listed without merging the default configuration

config/webpack/webpack.config.js
```diff
@@ -170,6 +173,14 @@ const webpackConfig = (options, env, argv) => {
   });
 };
 
+const basepathConfig = {
```
mjhenkes, line 176: You can just add the plugin to the devConfig plugin block instead of merging again.

config/webpack/webpack.config.js
```diff
@@ -170,6 +172,14 @@ const webpackConfig = (options, env, argv) => {
...
+  plugins: [
+    new webpack.DefinePlugin({
+      BASEPATH: JSON.stringify(base),
```
emilyrohrbough, line 179: If base is not defined, do we want to define it with the DefinePlugin?

config/webpack/webpack.config.js
```diff
@@ -9,8 +9,11 @@ const TerserPlugin = require('terser-webpack-plugin');
 const merge = require('webpack-merge');
 const DuplicatePackageCheckerPlugin = require('duplicate-package-checker-webpack-plugin');
 const aggregateTranslations = require('terra-aggregate-translations');
+const webpack = require('webpack');
 const ThemeAggregator = require('../../scripts/aggregate-themes/theme-aggregator');
 
+const base = process.env.BASEPATH;
```
mjhenkes, line 15: Does this default to undefined?

mjhenkes: If so, lets conditionally add the plugin based on the presence of this variable.

jessica-bailey:
Closing to make updates in orion-toolkit

PR #306, opened 2019-06-21
Consume updated duplicate package plugin

tbiethman:
### Summary

This PR replaces the `duplicate-package-checker-webpack-plugin` with the Cerner maintained fork. It also adds configuration to flag duplicates of 'react', 'react-dom', and 'react-intl' as errors rather than warnings at build time.

Given those errors, we no longer need to alias the react, react-dom, or react-intl packages. However, removing the aliases could result in a non-passive minor release if a consumer is absolutely reliant on those aliases. Personally, I feel better about making the issues known rather than masking a potentially serious problem, but it will probably burn a few folks. Thoughts?


config/webpack/webpack.config.js
```diff
@@ -108,17 +108,12 @@ const webpackConfig = (options, env, argv) => {
       }),
       new DuplicatePackageCheckerPlugin({
         showHelp: false,
+        alwaysEmitErrorsFor: ['react', 'react-dom', 'react-intl'],
```
emilyrohrbough: > react-intl

Oh, this might be an issue for any project not yet uplifted to terra-base v5 but is using terra-dev-site v4...

mjhenkes: Shouldn’t this also include disclosure manager and any of our other packages with context?

emilyrohrbough:
We should add documentation on the Webpack page for this plugin and how one should resolve any issues they see with this output

mjhenkes (approved):

StephenEsser (approved):

config/webpack/webpack.config.js
```diff
@@ -108,17 +108,23 @@ const webpackConfig = (options, env, argv) => {
...
+          'react',
+          'react-dom',
+          'react-intl',
+          'terra-base',
+          'terra-breakpoints',
+          'terra-application',
+          'terra-disclosure-manager',
+          'terra-modal-manager',
+          'terra-navigation-prompt',
+          'terra-slide-panel-manager',
```
ryanthemanuel: Can we add react-on-rails?

bjankord (approved):

config/webpack/webpack.config.js
```diff
@@ -108,17 +108,23 @@ const webpackConfig = (options, env, argv) => {
       }),
       new DuplicatePackageCheckerPlugin({
         showHelp: false,
+        alwaysEmitErrorsFor: [
+          'react',
+          'react-dom',
+          'react-intl',
+          'terra-base',
```
mjhenkes: Should terra-base be here? I'm wondering if it will break people more than we'd like (as of right now, us in terra-framework)

StephenEsser: image

config/webpack/webpack.config.js
tbiethman: Added the whole gang

config/webpack/webpack.config.js
tbiethman: Sure

config/webpack/webpack.config.js
tbiethman: We could remove terra-base for the time being and only enforce usage of it through terra-application. 🤷‍♂

tbiethman: Using the same logic here to remove terra-modal-manager and terra-slide-panel-manager from the list. terra-disclosure-manager is the package that really needs to be controlled, and those others have peer dependencies on it, so everything should work out.

emilyrohrbough:
We should open a validation PR in terra-framework to check this doesn't impact.

tbiethman:
PR in terra-framework: https://github.com/cerner/terra-framework/pull/754

emilyrohrbough (approved):

PR #309, opened 2019-06-27
Adding after command helper to help keep dev-site code splitting passive.

mjhenkes:
### Summary
With code splitting the js tests may now load after initial load of the page, this may cause tests to fail because they're trying to run too soon.

To address this I've added an after command hook that will
1. After every Url or refresh command, look for a specific selector that will be on the dom only for not yet loaded dev site pages
2. If the selector is found wait up till 5 seconds while looking for a selector that will be present only on fully loaded dev site pages.

This should be a no-op for all existing tests until dev-site v6 is released.

I will added a terra-framework test link confirming the passivity of these changes.
terra-framework test branch: https://github.com/cerner/terra-framework/pull/762
### Additional Details

@cerner/terra


src/wdio/services/TerraService.js
```diff
@@ -64,4 +64,21 @@ export default class TerraService {
...
     viewportHelpers.setViewport(global.browser.options.formFactor);
   }
+
+  // To more passively support code splitting in terra dev site, wait for data to load before progressing with the test.
+  // eslint-disable-next-line class-methods-use-this
+  afterCommand(commandName) {
+    if (commandName === 'refresh' || (commandName === 'url')) {
+      if (global.browser.isExisting('[data-terra-dev-site-loading]')) {
+        try {
+          global.browser.waitUntil(() => (
```
emilyrohrbough, line 74: I have a feeling this is could  break in IE since we are passing a function. We double check this.

src/wdio/services/TerraService.js
```diff
@@ -64,4 +64,21 @@ export default class TerraService {
...
+            global.browser.isExisting('[data-terra-dev-site-content]')
+          ), 5000, '', 100);
```
emilyrohrbough: Thoughts on updating this to be `global.browser.options. waitforTimeout`. We currently set this value to 3000

src/wdio/services/TerraService.js
mjhenkes: This isn't a browser.execute function, It's borrowed from here:
https://github.com/webdriverio-boneyard/v4/blob/master/lib/commands/waitForExist.js
but we don't need all the reversing logic.

src/wdio/services/TerraService.js
mjhenkes: I had it set to that but saw some random timeouts and bumped the value up. There isn't harm in increasing the timeout as this only will affect dev site code split examples.

src/wdio/services/TerraService.js
emilyrohrbough: Ah so we didn't want to do:
`browser.waitForExist('[data-terra-dev-site-content]', 5000);`

bjankord (approved):
Ship it 
SHIP IT

emilyrohrbough (approved):

PR #322, opened 2019-08-22
Add functionality to support theming with code splitting in IE

### LINKED ISSUE #200: Code splitting bundle segregation for theming
# Future Functionality Discussion

## Description
When working on a code splitting project, an issue was run across where dynamically imported bundles were not themed correctly in certain browsers. This is likely due to older browsers not being able to dynamically apply CSS variables when imported. As such, some items do not receive correct theming if they are included in a bundle that got imported after static variable application.

This is avoidable by creating a wrapping entrypoint which specifically pulls in the theme code. However, this is not scalable when multiple themes are involved (one entrypoint and container logic per theme).

There was some discussion of a webpack plugin injecting theme code into each bundle. Not sure the specifics on that, but as a starting point, it makes sense.

## @ Mentions
@mjhenkes @tbiethman @bjankord @ryanthemanuel

ryanthemanuel:
### Summary

Add functionality to support theming with code splitting in IE

Basically the idea will be, if we have a theme file, use a webpack promise to webpack it.  This promise will resolve to an object with themeable variables and values. This will then be used by the PostCSSCustomProperties plugin to update the end state CSS so that it is populated with values if variables aren't supported (e.g. IE10). This dance is necessary when code splitting to ensure the variables and values are applied across all code split css files.

In testing, this additional webpacking of the single file takes about a second.  It's much more streamlined than the entire tree of JS/CSS/etc.

Fixes #200


emilyrohrbough:
@ryanthemanuel Can you add some info on this?

ryanthemanuel:
@emilyrohrbough I added some additional context to the PR.

config/webpack/getThemeWebpackPromise.js
```diff
@@ -0,0 +1,120 @@
...
+const PostCSSAssetsPlugin = require('postcss-assets-webpack-plugin');
+const PostCSSCustomProperties = require('postcss-custom-properties');
+const path = require('path');
+const Logger = require('../../scripts/utils/logger');
+
+/**
+ * This is a basic configuration to webpack just the theme file and grab the resulting css.  Thus,
+ * there are no JS, raw, etc. loaders. The resulting themeable variables and values will be populated in
+ * the cachedObject
+ * @param {String} rootPath the root path from where the webpack is being run
```
emilyrohrbough, line 16: I think it'd be worth mentioning this does not support hot reloading and then will only compile the default theme defined int he theme configuration.

config/webpack/getThemeWebpackPromise.js
```diff
@@ -0,0 +1,120 @@
...
+ * @param {String} themeFile the file to be webpacked
+ * @param {Object} cachedObject this object will be populated with the resulting themeable variables and values
+ */
+const themeConfig = (rootPath, themeFile, cachedObject) => (
+  {
+    mode: 'production',
+    entry: {
```
emilyrohrbough, line 23: Do we need an entry point?

config/webpack/getThemeWebpackPromise.js
```diff
@@ -0,0 +1,120 @@
...
+            // Here is where the cachedObject is populated
+            exportTo: [
+              cachedObject,
+            ],
+          }),
+        ],
+      }),
+    ],
+    resolve: {
+      extensions: ['.js'],
```
emilyrohrbough, line 81: I don't think we need this resolve config since we are only concerned with styles

config/webpack/webpack.config.js
```diff
@@ -106,7 +107,15 @@ const webpackConfig = (options, env, argv) => {
...
         plugins: [
-          PostCSSCustomProperties({ preserve: true }),
+          PostCSSCustomProperties({
+            preserve: true,
+            // If we have a theme file, use the webpack promise to webpack it.  This promise will resolve to
+            // an object with themeable variables and values. This will then be used to update the end state CSS
+            // so that they are populated with values if variables aren't supported (e.g. IE10). This dance is
+            // necessary when code splitting to ensure the variables and values are applied across all code split
+            // css files
+            ...themeFile && { importFrom: [getThemeWebpackPromise(rootPath, themeFile)] },
```
emilyrohrbough, line 117: Can we run this wepback promise as a "prebuild" step instead of running webpack-within-webpack?  something like:

```diff
const webpackConfig = (options, env, argv) => {
  const {
    rootPath,
    resolveModules,
    staticOptions,
-   themeFile,
+  theme,
  } = options;

....
   entry: {
      ...
-      ...themeFile && { theme: themeFile },
+      ...theme && { theme: theme.themeFile },
    },
....
          PostCSSCustomProperties({
               preserve: true
-              ...themeFile && { importFrom: [getThemeWebpackPromise(rootPath, themeFile)] },
+              ....theme && { importFrom: [ theme.chunkedThemeOutput ]
}

const defaultWebpackConfig = (env = {}, argv = {}) => {
....
const themeFile = ThemeAggregator.aggregate();
+   if (themeFile) {
+       const chunkedThemeOutput = getThemeWebpackPromise(rootPath, themeFile);
+   }

    const options = {
        rootPath,
        resolveModules,
        staticOptions,
-       themeFile,
+      theme { themeFile,  chunkedThemeOutput },
      };

    return webpackConfig(options, env, argv);
}

```

We also need to manage what we do if the project errors out and providing meaningful error messaging so consumers know how to resolve this. For instance, if we dependency is missing, does this throw the same error messages as our config or what will that look like for consumers to start debugging?

config/webpack/getThemeWebpackPromise.js
```diff
@@ -0,0 +1,120 @@
...
+        aggregatedCustomPropertiesJSON[key] = customProperties[key].map(customPropertyComponent => String(customPropertyComponent)).join(' ');
+        return aggregatedCustomPropertiesJSON;
+      }, {})
+    ),
+    customProperties: {},
+  };
+  return new Promise((resolve, reject) => {
+    const compiler = webpack(themeConfig(rootPath, themeFile, cachedObject));
+    // Set the output file system to MemoryFS so that this all runs in memory
+    compiler.outputFileSystem = new MemoryFS();
```
emilyrohrbough, line 110: Do we need to run this in memory? mostly because I really don't want to add another dependency to toolkit

config/webpack/getThemeWebpackPromise.js
ryanthemanuel: Yes.  Webpack requires one as the entry point to the js then sass then css.

config/webpack/webpack.config.js
ryanthemanuel: Unfortunately we can't do it like that.  the promise isn't synchronous and it needs to be guaranteed to happen before we run the main webpack.

Good call as far as errors, it should display the same errors as what webpack does by using the recommendation here: https://webpack.js.org/api/node/#error-handling

config/webpack/getThemeWebpackPromise.js
ryanthemanuel: I don't want to have anything generated.  Otherwise a new js and css file will be generated.

config/webpack/webpack.config.js
ryanthemanuel: https://github.com/cerner/terra-toolkit/pull/322/commits/b56514b05e791c048ac505b946d93f9c540ae8b7

config/webpack/getThemeWebpackPromise.js
ryanthemanuel: https://github.com/cerner/terra-toolkit/pull/322/commits/3555c6105d7c7120831210210fde31f130e4c0d8

config/webpack/webpack.config.js
emilyrohrbough: Could we wrap it in a promise chain to get around async? I saw @StephenEsser do something with auth recently. something like:

```
const themeFile = ThemeAggregator.aggregate();
const themePromise = themeFile ? getThemeWebpackPromise(rootPath, themeFile) : Promise.resolve(undefined);

return themePromise.then(themeOutput => webpackConfig({ ...options, themeOutput }, env, argv))

```

config/webpack/getThemeWebpackPromise.js
mjhenkes: and MemoryFS is already a transitive dependency due to webpack.

mjhenkes (approved):

yuderekyu (approved):

bjankord (approved):

StephenEsser (approved):

emilyrohrbough (approved):

PR #326, opened 2019-09-03
Add post-install warning if terra-toolkit is listed as a hard dependency

StephenEsser:
### Summary

This PR adds a post-install script that logs a warning message if terra-toolkit is installed as a hard dependency.

### Additional Details

It is possible for this warning to be printed multiple times if multiple versions of terra-toolkit are installed.

@cerner/terra


scripts/post-install/post-install.js
```diff
@@ -0,0 +1,20 @@
...
+  const { dependencies } = JSON.parse(file);
+
+  if (dependencies['terra-toolkit']) {
+    Logger.warn('+-------------------------------------------------------------------------------------------------------------+');
+    Logger.warn('|                                                   WARNING                                                   |');
+    Logger.warn('+-------------------------------------------------------------------------------------------------------------+');
+    Logger.warn('|                            terra-toolkit should be installed as a devDependency.                            |');
+    Logger.warn('|                                                                                                             |');
+    Logger.warn('| Modify the package.json to remove terra-toolkit from dependencies and add terra-toolkit to devDependencies. |');
+    Logger.warn('+-------------------------------------------------------------------------------------------------------------+');
```
StephenEsser, line 21: Formatting and warning message suggestions are welcome. The table can format strangely on small viewports, but it does help it stand out a little more in the terminal output.

scripts/post-install/post-install.js
```diff
@@ -0,0 +1,20 @@
...
+    throw Logger.error(error);
+  }
+
+  const { dependencies } = JSON.parse(file);
+
+  if (dependencies['terra-toolkit']) {
+    Logger.warn('+-------------------------------------------------------------------------------------------------------------+');
+    Logger.warn('|                                                   WARNING                                                   |');
+    Logger.warn('+-------------------------------------------------------------------------------------------------------------+');
+    Logger.warn('|                            terra-toolkit should be installed as a devDependency.                            |');
```
tbiethman: `should` -> `must`

StephenEsser: https://github.com/cerner/terra-toolkit/pull/326/commits/bfb2643f266a8596dbfeeccdbf18d7c05969087b

scripts/post-install/post-install.js
```diff
@@ -0,0 +1,20 @@
+const fs = require('fs');
+const Logger = require('../utils/logger');
+
+fs.readFile(`${process.env.INIT_CWD}/package.json`, (error, file) => {
```
mjhenkes, line 7: what is the difference between `process.env.INIT_CWD` and `process.cwd()`?

StephenEsser: This is a tricky thing with npm. `process.cwd()` doesn't return the initial process directory, it returns the package being installed as the current working directory. `INIT_CWD` was added as a work around for this issue. I found a short discussion about it here: 

https://github.com/npm/npm/issues/16990

bjankord: Might be good to leave a code comment about this and link to the npm issue.

bjankord (approved):

scripts/post-install/post-install.js
```diff
@@ -0,0 +1,20 @@
...
+  if (error) {
+    throw Logger.error(error);
```
emilyrohrbough, line 9: Should we silently return null if theres an error?

StephenEsser: We can, but we might start having issues we never know about

StephenEsser:
Here's a sample build output for anyone who is curious how the logging looks: https://travis-ci.com/cerner/terra-clinical/jobs/230813450

scripts/post-install/post-install.js
emilyrohrbough: won't it come from reading the file though? I'd hate to error on people's installs

emilyrohrbough (approved):
I feel like this will be ignored by people in the console but changes look good

scripts/post-install/post-install.js
StephenEsser: We can grab a few more opinions on this before merging. I don't mind removing the throw, but the trade off is that we will never be alerted this isn't working correctly.

StephenEsser:
> I feel like this will be ignored by people in the console but changes look good

Hopefully at least a few people take notice of the message. We can discuss returning a non-zero exit code to fail the install on the next major version bump of terra-toolkit.

scripts/post-install/post-install.js
StephenEsser: Added a short comment and reference: https://github.com/cerner/terra-toolkit/pull/326/commits/6b9636e5d7d61cd6afeab5b9fc973e7b12b1fb03

tbiethman (approved):

scripts/post-install/post-install.js
mjhenkes: Should we state that this will fail install in the future? Even if it's just a threat? 😈

mjhenkes (approved):

ryanthemanuel (approved):

nramamurth (approved):

PR #329, opened 2019-09-10
Add post-install warning if terra-toolkit is listed as a hard dependency

StephenEsser:
### Summary

This PR adds a post-install script that logs a warning message if terra-toolkit is installed as a hard dependency.

### Additional Details

There is a packed `.tgz` file being used for testing, this will be removed before merging. 

This feature had previously been [reverted](https://github.com/cerner/terra-toolkit/pull/328) because a team was noticing their Jenkins build was failing. Upon investigation it was discovered they were using an old docker container that was running node v8.4.0 and npm 5.3.0. `INIT_CWD` was introduced in npm 5.4.0. A presence check was added that gracefully returns if `INIT_CWD` is undefined.

Previous PR: https://github.com/cerner/terra-toolkit/pull/326

@cerner/terra


scripts/post-install/post-install.js
```diff
@@ -0,0 +1,40 @@
...
+   * Note: This feature was introduced in npm 5.4.0.
+   */
+  if (!process.env.INIT_CWD) {
+    Logger.warn('WARNING: [terra-toolkit] - INIT_CWD was undefined. This is likely due to an outdated version of npm. Please consider upgrading.');
+    return;
+  }
+
+  fs.readFile(`${process.env.INIT_CWD}/package.json`, (error, file) => {
+    if (error) {
+      Logger.warn('WARNING: [terra-toolkit] - Unable to read package.json in post-install script.');
```
emilyrohrbough:
```suggestion
      Logger.warn('WARNING: [terra-toolkit] - Unable to validate if terra-toolkit was defined as a dev-dependency. Please ensure terra-toolkit has only added as a dependency. .');
```

scripts/post-install/post-install.js
```diff
@@ -0,0 +1,40 @@
...
+ * terra-toolkit is expected to always be installed as a devDependency.
+ */
+function validateInstall() {
+  /**
+   * INIT_CWD returns the initial working directory of the npm install.
+   * See the following issue for more information: https://github.com/npm/npm/issues/16990
+   * Note: This feature was introduced in npm 5.4.0.
+   */
+  if (!process.env.INIT_CWD) {
+    Logger.warn('WARNING: [terra-toolkit] - INIT_CWD was undefined. This is likely due to an outdated version of npm. Please consider upgrading.');
```
emilyrohrbough, line 16: I would like to add toolkit 
```suggestion
    Logger.warn('WARNING: [terra-toolkit] - INIT_CWD was undefined. This is likely due to an outdated version of npm. Please consider upgrading; terra-toolkit does not support this npm version.');
```

emilyrohrbough (approved):

scripts/post-install/post-install.js
StephenEsser: I don't know if that description is entirely accurate, but it is more persuasive to upgrade. I appended the additional message: https://github.com/cerner/terra-toolkit/pull/329/commits/9ee6b3d77128f8aeca16e74417b45a53ea3399c9

scripts/post-install/post-install.js
StephenEsser: I think there were a few mistakes in the suggestion, I added something very similar: 9ee6b3d
```
WARNING: [terra-toolkit] - Unable to validate if terra-toolkit was installed as a dev-dependency. Please ensure terra-toolkit is defined as a devDependency and not a dependency in the package.json
```

scripts/post-install/post-install.js
emilyrohrbough: didn't you say the node version determined the npm version? or did I misunderstand? We define the minimum supported version as [node 8.9.2](https://github.com/cerner/terra-toolkit/blob/master/package.json#L7)

scripts/post-install/post-install.js
emilyrohrbough: lol yeah my description wasn't very accurate

scripts/post-install/post-install.js
StephenEsser: When you use nvm it installs an npm version with the node version, but I think you can technically set npm to be any version of your choosing

scripts/post-install/post-install.js
```diff
@@ -0,0 +1,40 @@
...
+ * terra-toolkit is expected to always be installed as a devDependency.
+ */
+function validateInstall() {
+  /**
+   * INIT_CWD returns the initial working directory of the npm install.
+   * See the following issue for more information: https://github.com/npm/npm/issues/16990
+   * Note: This feature was introduced in npm 5.4.0.
+   */
+  if (!process.env.INIT_CWD) {
+    Logger.warn('WARNING: [terra-toolkit] - INIT_CWD was undefined. This is likely due to an outdated version of npm. Please consider upgrading. terra-toolkit does not support this version of npm.');
```
mjhenkes: We just now decided with this file change that we don't support npm prior to version 5.4.0. I'm not sure this checker is a good enough reason to claim we don't support an npm version.

StephenEsser: There's a discussion here regarding the change: https://github.com/cerner/terra-toolkit/pull/329#discussion_r322816362

I'm fine with removing it or keeping it. The definition of our support is a little blurry with our declaration here: https://github.com/cerner/terra-toolkit/blob/master/package.json#L7

StephenEsser: Node and NPM versions are tightly coupled: https://nodejs.org/en/download/releases/

StephenEsser: Removed the unsupported message: https://github.com/cerner/terra-toolkit/pull/329/commits/43e0f8877f28d110e342f0caa1f8d1fc4d0bbbf1

scripts/post-install/post-install.js
emilyrohrbough: ahh

StephenEsser: Removed the unsupported message after the discussion here: https://github.com/cerner/terra-toolkit/pull/329#discussion_r322837728

emilyrohrbough: I still approve 😄 its a pretty blurry line

mjhenkes (approved):

bjankord (approved):

tbiethman (approved):

ryanthemanuel (approved):

PR #341, opened 2019-10-21
Using DefinePlugin to create global constants for build date/time and package version

tbiethman:
### Summary

Some applications may need to reference the webpack execution time or application version to comply with regulatory standards. I updated the default webpack configuration to define constants for these data points so that consuming applications can access them easily.


config/webpack/webpack.config.js
```diff
@@ -133,6 +134,10 @@ const webpackConfig = (options, env, argv) => {
           'terra-navigation-prompt',
         ],
       }),
+      new webpack.DefinePlugin({
+        PACKAGE_VERSION: JSON.stringify(process.env.npm_package_version),
+        WEBPACK_BUILD_TIMESTAMP: JSON.stringify(new Date(Date.now()).toISOString()),
+      }),
     ],
```
tbiethman, line 140: If you're wondering what's up with the `stringify`s, from the [docs](https://webpack.js.org/plugins/define-plugin/):

> Note that because the plugin does a direct text replacement, the value given to it must include actual quotes inside of the string itself. Typically, this is done either with alternate quotes, such as '"production"', or by using JSON.stringify('production').

emilyrohrbough:
should these changes be in the webpack config we provide for servers?

config/webpack/webpack.config.js
```diff
@@ -133,6 +134,10 @@ const webpackConfig = (options, env, argv) => {
           'terra-navigation-prompt',
         ],
       }),
+      new webpack.DefinePlugin({
+        PACKAGE_VERSION: JSON.stringify(process.env.npm_package_version),
```
emilyrohrbough: is process.env.npm_package_version available when running in yarn?

emilyrohrbough: will this require use to release and/or bump the client package-json in out servers for this version to change?

tbiethman: 1. I think so? I see package.json vars referenced within their documentation.

2. Yes, that version would need to be updated. Given that, I don't think the version we're providing here would be all useful with how we're deploying applications today. I kept it in because it _could_ be helpful in some ways, and just exposing it isn't hurting anything, but I also wouldn't argue over just removing it in the meantime. The build date is much more necessary.

emilyrohrbough (approved):
Looks good. The changelog needs a small tweak to remove the versioning bit though

mjhenkes (approved):

config/webpack/webpack.config.js
tbiethman: Went ahead and removed version for the time being

yuderekyu (approved):

PR #356, opened 2019-11-18
Fixed custom global axe rules specified in wdio config to be applied/implemented

### LINKED ISSUE #351: Custom axe rules specified in wdio.conf.js are not being reflected/implemented
# Feature Request 

## Description

Axe rules specified in wdio.conf.js are not being reflected/implemented as expected.

## Possible Solution
Merging the global axe rules with the custom axe rules could resolve this issue.

## @ Mentions

@ryanthemanuel @neilpfeiffer

avinashg1994:
### Summary

Merged global axe rules with default/normal axe rules to allow axe customizations from custom wdio.conf.js

Fixes #351


src/wdio/services/TerraCommands/determine-test-options.js
```diff
@@ -49,7 +49,9 @@ const axeOptions = (args) => {
     options, viewports,
   } = determineArgs(args);
 
-  const axeRules = options.rules || options.axeRules;
+  const globalAxeRules = (global.browser.options.axe.options) ? global.browser.options.axe.options.rules : undefined;
+  const customAxeRules = options.rules || options.axeRules;
+  const axeRules = { ...globalAxeRules, ...customAxeRules };
```
emilyrohrbough: Does this work as expected? If I remember correctly, the glabal Axe rules will be passes as an array of rules config like:
```
axe: {
   options: {
        rules: [
            'rule': { enabled: false }
        ],
     }
}
```
 where as the custom axe rules are passes as an object of rules:
```
Terra.it.isAccessible({ rules: {
      'rule': { enabled: false },
 });
```

emilyrohrbough (commented):
There should be a jest test added for this change.

src/wdio/services/TerraCommands/determine-test-options.js
avinashg1994: @emilyrohrbough ,
Yes, you are right, the global axe rules and custom axe rules are different and the final merged rules contain both these formats with different key value pairs.
I might be wrong but i guess both the formats are supported, that is the reason why even after merging the rules with different formats the rules still apply.

Either way, I could only think of adding something similar to the following code to convert them into a single format:
```
const globalConfig = [ {id: 'color-contrast', enabled: false}, {id: 'aria-alt', enabled: true}];
const specConfig = { 'color-contrast': {enabled: true, any: 'hello'} };
const result = {};
globalConfig.forEach((entry, index) => {
  const {id, ...rest} = entry;
  result[id] = rest;
});
console.log({...specConfig, ...result});
```
So what are your thoughts or suggestions on this?

emilyrohrbough: Can we use array.reduce to accomplish the conversion from an array to an object?
We will also want to spread the test specific rules after the global config.

Have you tried a few options out and validated with tests since you've posted this question?

avinashg1994: I have implemented the same using `array.reduce` and spread the test specific rules after the global config.

> Have you tried a few options out and validated with tests

Yes, i have tried with a few options and the logic works but I have yet to write tests for this.

avinashg1994:
> There should be a jest test added for this change.

All changes have been addressed and Jest tests have been added for the merge logic.

emilyrohrbough:
@avinashg1994  Can you add a wdio test and open a PR in terra-core to confirm these changes work as expect?

avinashg1994:
> @avinashg1994 Can you add a wdio test and open a PR in terra-core to confirm these changes work as expect?

@emilyrohrbough 
There is already a PR out in terra-core which can be used to verify this : https://github.com/cerner/terra-core/pull/2758/

And as for wdio tests in toolkit : https://github.com/cerner/terra-toolkit/blob/master/tests/wdio/beAccessible-spec.js
Does this suffice? 'or' Do we need to add wdio tests?

emilyrohrbough:
@avinashg1994 
> There is already a PR out in terra-core which can be used to verify this : cerner/terra-core#2758

Ah didn't know you had it out. Thanks for the link. That works for me 😄 

I would expect these changes to work for both of these wdio tests, where both global rules are applied. I think more tests are better since these are so heavily used by consumers:
- https://github.com/cerner/terra-toolkit/blob/master/tests/wdio/axe-spec.js
-  https://github.com/cerner/terra-toolkit/blob/master/tests/wdio/beAccessible-spec.js

avinashg1994:
I have added wdio tests as well to verify this change : https://github.com/cerner/terra-toolkit/pull/356/commits/effa83e0f92361d6226ba3de17da015195bc1051

emilyrohrbough:
@avinashg1994 I have added a few more test to validate the axe commands, and it isn't taking your changed and honoring the global configuration: 

https://github.com/cerner/terra-toolkit/pull/356/commits/d9704a809fa115cf6c92cc9b49126e2f6b8a6a65

avinashg1994:
> I have added a few more test to validate the axe commands, and it isn't taking your changed and honoring the global configuration:

Yes, the tests added are expected to fail as per the logic.
Since the logic for merging the global and custom axe rules was added in "axeOptions" helper method in "determine-test-options.js". This method is only invoked when we use our Terra helpers and since the tests written in "axe-spec.js" are invoking/testing using the axe native methods, the merge logic is never invoked, so only the custom rules will be applied and not the global rules.

So, now do we merge the rules at [runAxeTest](https://github.com/cerner/carbon-graphs/blob/master/docs/contributing/README.md) in "axe-command.js", which makes it so that the rules are applied always irrespective of the method.
'or' 
Do we merge at [axeOptions](https://github.com/cerner/terra-toolkit/blob/master/src/wdio/services/TerraCommands/determine-test-options.js#L47) in "determine-test-options", so that it is specific for our Terra helpers.

emilyrohrbough:
We want to merge the rules for both commands. They should behave the same.
We [add the axe-commmand to the browser object in the Terra Service](https://github.com/cerner/terra-toolkit/blob/master/src/wdio/services/TerraService.js#L26). Then, if you look at the Terra helpers, it is calling the [alias we used when adding the axe-command to the browser to run axe tests](https://github.com/cerner/terra-toolkit/blob/master/src/wdio/services/TerraCommands/accessibility.js#L13).

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ Changelog
 Unreleased
 ----------
 ### Changed
-* Update the Firefox test browser to use the `selenium/node-firefox:3.141.59-titanium` docker image
+* Merged global axe rules with default/normal axe rules to allow axe customizations from custom wdio.conf.js
```
nramamurth: I'd like to see a note here or in the Wiki that talks about which amongst the two takes precedence - the rules sent as part of a particular spec or the rules configured in the _wdio_ config?

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ Unreleased
 * Add support for onBeforeUnload in firefox selenium testing
 * Adding git to terra-node-base to be able to reference git branches while testing.
 
+### Changed
+* Update the Firefox test browser to use the `selenium/node-firefox:3.141.59-titanium` docker image
```
nramamurth: Please move this to "unreleased" section.

src/wdio/services/TerraCommands/axe-command.js
```diff
@@ -30,7 +30,18 @@ const injectAxe = (axeOptions) => {
 *
 * @returns {Object} axeResults - the axe results as list of passes, violations, incomplete and inapplicable.
 */
-const runAxeTest = (rules) => {
+const runAxeTest = (axeRules) => {
+  const globalAxeRules = (browser.options.axe && browser.options.axe.options) ? browser.options.axe.options.rules : undefined;
+  // Converting `globalAxeRules` from array of rules object to object of rules.
```
nramamurth, line 35: I am a bit confused — what stops us from providing the rules in the format below in _wdio_ config? Since toolkit dictates how the rules should look like (imo) and is responsible for massaging them and running the `axe` command against them, I'd expect it to work and further avoid the overhead of maintaining two different formats and having to do this for every accessibility test run. 

```
axe: {
   inject: true, 
   options: {
       rules: {
         'color-contrast': { enabled: true },
         'valid-lang': { enabled: false }
      }
   }
}
```

src/wdio/services/TerraCommands/axe-command.js
```diff
@@ -30,7 +30,18 @@ const injectAxe = (axeOptions) => {
...
+  const formattedGlobalRules = globalAxeRules && globalAxeRules.reduce((formattedRules, rule) => {
+    const { id, ...rest } = rule;
+    const rulesObject = formattedRules;
```
nramamurth: What is happening here? You were looking to clone it?

nramamurth: 🙊super nit - I'd avoid the 'object' suffix.

avinashg1994: Addressed here : https://github.com/cerner/terra-toolkit/pull/356/commits/50ac3e447dfe3961f6e15fdd57bf106335fd8844

avinashg1994: Yeah, I did not want to mutate the parameter, so I did want to clone it
I see your concern and have addressed here : https://github.com/cerner/terra-toolkit/pull/356/commits/50ac3e447dfe3961f6e15fdd57bf106335fd8844

CHANGELOG.md
avinashg1994: Changed here : https://github.com/cerner/terra-toolkit/pull/356/commits/50ac3e447dfe3961f6e15fdd57bf106335fd8844

avinashg1994:
> We want to merge the rules for both commands. They should behave the same.

I have moved the merge logic into the `runAxeTest` method in "axe-command.js" file.
So now all the wdio tests are passing and the merge of global and custom axe rules occurs for any axe command used to test accessibility.

src/wdio/services/TerraCommands/axe-command.js
avinashg1994: All global axe rules specified in the custom wdio config are injected using the `axe.configure` as seen [here](https://github.com/cerner/terra-toolkit/blob/master/src/wdio/services/TerraCommands/axe-command.js#L64).
So, the above mentioned format will not be supported as per the [documentation](https://www.deque.com/axe/axe-for-web/documentation/api-documentation/#api-name-axeconfigure) and only accepts rules as an array of objects in the specified format.

CHANGELOG.md
avinashg1994: @emilyrohrbough, your thoughts on this?

emilyrohrbough: anything defined at the test level will be used over the global configuration. I think adding documentation on this behavior would be good. 👍

emilyrohrbough (approved):

emilyrohrbough:
@avinashg1994  your latest changes look good! Good catch at finding this bug.

CHANGELOG.md
avinashg1994: Updated documentation : https://github.com/cerner/terra-toolkit/pull/356/commits/d7d2f237bb503854958d8aa3b9ed8ad76ea95e07

benbcai (approved):

mjhenkes (approved):

PR #361, opened 2019-12-05
Add new docker containers

mjhenkes:
### Summary
Add new docker containers based on alpine for development and testing.

### Additional Details
Validated here, I pushed snapshots of these docker containers to my personal docker hub account to test against.
https://github.com/cerner/terra-framework/pull/981
https://github.com/cerner/terra-dev-site/pull/223

There will be a follow on pull request after this to consume these new docker containers in terra-toolkit because we have a chicken and the egg problem with releasing and using the docker containers here.

@cerner/terra

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,14 +27,14 @@ jobs:
         - travis_wait docker build -t cerner/terra-node -f ./docker/terra-node/Dockerfile ./docker/terra-node/
         - travis_wait docker build -t cerner/terra-node-base:latest -f ./docker/terra-node-base/Dockerfile ./docker/terra-node-base/
         - travis_wait docker build -t cerner/terra-node-ci:latest -f ./docker/terra-node-ci/Dockerfile ./docker/terra-node-ci/
-        - travis_wait docker build -t cerner/terra-node-term:latest -f ./docker/terra-node-term/Dockerfile ./docker/terra-node-term/
+        - travis_wait docker build -t cerner/terra-node-zsh:latest -f ./docker/terra-node-zsh/Dockerfile ./docker/terra-node-zsh/
         - travis_wait docker push cerner/terra-node
         - travis_wait docker push cerner/terra-node-base:latest
         - travis_wait docker push cerner/terra-node-base:1
         - travis_wait docker push cerner/terra-node-ci:latest
         - travis_wait docker push cerner/terra-node-ci:1
-        - travis_wait docker push cerner/terra-node-term:latest
-        - travis_wait docker push cerner/terra-node-term:1
+        - travis_wait docker push cerner/terra-node-zsh:latest
+        - travis_wait docker push cerner/terra-node-zsh:1
     - stage: build
       before_install: true
       script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,14 +27,14 @@ jobs:
         - travis_wait docker build -t cerner/terra-node -f ./docker/terra-node/Dockerfile ./docker/terra-node/
         - travis_wait docker build -t cerner/terra-node-base:latest -f ./docker/terra-node-base/Dockerfile ./docker/terra-node-base/
         - travis_wait docker build -t cerner/terra-node-ci:latest -f ./docker/terra-node-ci/Dockerfile ./docker/terra-node-ci/
-        - travis_wait docker build -t cerner/terra-node-zsh:latest -f ./docker/terra-node-zsh/Dockerfile ./docker/terra-node-zsh/
+        - travis_wait docker build -t cerner/terra-node-dev:latest -f ./docker/terra-node-dev/Dockerfile ./docker/terra-node-dev/
         - travis_wait docker push cerner/terra-node
         - travis_wait docker push cerner/terra-node-base:latest
         - travis_wait docker push cerner/terra-node-base:1
         - travis_wait docker push cerner/terra-node-ci:latest
         - travis_wait docker push cerner/terra-node-ci:1
-        - travis_wait docker push cerner/terra-node-zsh:latest
-        - travis_wait docker push cerner/terra-node-zsh:1
+        - travis_wait docker push cerner/terra-node-dev:latest
+        - travis_wait docker push cerner/terra-node-dev:1
     - stage: build
       before_install: true
       script:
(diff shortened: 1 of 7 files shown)


config/wdio/wdio.conf.js
```diff
@@ -22,7 +22,7 @@ const internalPort = process.env.WDIO_INTERNAL_PORT || 8080;
 const ci = process.env.CI;
 
 /* Use to bail fast while running locally. */
-const bail = process.env.WDIO_BAIL || ci;
+const bail = process.env.WDIO_BAIL !== 'FALSE' && (process.env.WDIO_BAIL || ci);
```
mjhenkes: This change allows me to override the bail flag if the ci flag is set. Previously in my development docker container i'd always have to go and manually edit this config to run all my tests instead of just failing on the first one.

Anything other than FALSE for the WDIO_BAIL env will still set bail to true. But now setting the WDIO_BAIL env to 'FALSE' will not set bail to true too :)

.travis.yml
```diff
@@ -25,7 +25,16 @@ jobs:
...
+        - travis_wait docker build -t cerner/terra-node-base:latest -f ./docker/terra-node-base/Dockerfile ./docker/terra-node-base/
+        - travis_wait docker build -t cerner/terra-node-ci:latest -f ./docker/terra-node-ci/Dockerfile ./docker/terra-node-ci/
+        - travis_wait docker build -t cerner/terra-node-zsh:latest -f ./docker/terra-node-zsh/Dockerfile ./docker/terra-node-zsh/
         - travis_wait docker push cerner/terra-node
+        - travis_wait docker push cerner/terra-node-base:latest
+        - travis_wait docker push cerner/terra-node-base:1
+        - travis_wait docker push cerner/terra-node-ci:latest
+        - travis_wait docker push cerner/terra-node-ci:1
+        - travis_wait docker push cerner/terra-node-zsh:latest
+        - travis_wait docker push cerner/terra-node-zsh:1
```
mjhenkes: I'm adding versions to these images to allow us to upgrade them more easily in the future.

docker/terra-node-base/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:dubnium-alpine
+
+EXPOSE 8080
+
+ENV CI true
+
+RUN npm config set unsafe-perm true -g
+
+WORKDIR /opt/module
```
mjhenkes, line 11: Variable work directories is no longer available or needed. (This was originally added for terra dev site, but is no longer needed).

.travis.yml
mjhenkes: Also I'm adding new docker images to keep this passive. We could remove terra-node in the future.

docker/terra-node-ci/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM cerner/terra-node-base:latest
+
+# Copy all the things
+ONBUILD COPY . /opt/module
+
+ONBUILD RUN npm install --loglevel error
```
emilyrohrbough, line 7: Do we need this Dockerfile? Why not use the base image directly?

docker/terra-node-zsh/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM cerner/terra-node-base:latest
+
+#install zsh / oh-my-zsh
+RUN apk add zsh curl wget git \
+  && wget https://github.com/robbyrussell/oh-my-zsh/raw/master/tools/install.sh -O - | zsh || true
+
+ENV WDIO_BAIL FALSE
```
emilyrohrbough: why not `ENV WDIO_BAIL false`?

docker/terra-node-zsh/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM cerner/terra-node-base:latest
```
emilyrohrbough: Can we just declare this as our dev env instead of zsh in the case we ever want to swap this?

or if you are against my suggestion here: https://github.com/cerner/terra-dev-site/pull/223/files#r354510784

can we call this image terra-node-term?

docker/terra-node-ci/Dockerfile
emilyrohrbough: Did we drop the command: https://github.com/cerner/terra-toolkit/blob/master/docker/terra-node/Dockerfile#L26

docker/terra-node-zsh/Dockerfile
mjhenkes: seems like all caps env variables are in vogue, also i only check for all caps false in the config.

docker/terra-node-zsh/Dockerfile
mjhenkes: so rename this container `terra-node-dev`?
I'd be fine with that.

docker/terra-node-ci/Dockerfile
mjhenkes: Yes we need this docker container, instead of combining dev and ci into one container (dev doesn't need npm install or to copy the files, ci doesn't need zsh) i split them out into child docker containers.

yes I removed the npm install command, it's not needed with the separate docker containers.

docker/terra-node-zsh/Dockerfile
emilyrohrbough: I'm not sure I follow... If we set 
```
ENV WDIO_BAIL false
```
can't we remove [this change?](https://github.com/cerner/terra-toolkit/pull/361/files#diff-cbe4c848b6c22e669052e0ba331eeca6R25)

emilyrohrbough: oh I see.... nevermind... we want false even though CI is true....

mjhenkes: i switched it to lowercase false, that seems more normal? i dunno

emilyrohrbough (approved):

emilyrohrbough (commented):
Can we add docs on each image, what they support and when they should be used?

config/wdio/wdio.conf.js
```diff
@@ -22,7 +22,7 @@ const internalPort = process.env.WDIO_INTERNAL_PORT || 8080;
 const ci = process.env.CI;
 
 /* Use to bail fast while running locally. */
-const bail = process.env.WDIO_BAIL || ci;
+const bail = process.env.WDIO_BAIL !== 'false' && (process.env.WDIO_BAIL || ci);
```
ryanthemanuel, line 25: yay

ryanthemanuel (approved):

PR #376, opened 2020-01-14
[theme-aggregator] Always execute theme generation

### LINKED ISSUE #368: [theme-aggregator] Do not fallback to theme generation. Always execute theme generation instead.
# Update theme generation to not be a fall back
<a id="org9ff85c6"></a>

## Overview
When we aggregate themes, we search for `theme-name/root.scss` or `theme-name/scope.scss` files. If we fail to find files for aggregation, we fallback to generating themes; we then search for `theme-name/theme-name.scss` in order to generate `root.scss` or `scope.scss` files. This naming convention was purposefully built as a passive theme generation fallback strategy for instances when a project did not pull in an internal theme. But in hindsight, only triggering generation as a fallback leaves a hole for themed applications that pull in different versions of components.

The result - if an application has any current Terra components but also pulls in a matching internal theme, only the internal theme is imported.

If we open the theme-aggregator to do both, it will result in chance collisions or odd visual behavior. We then need a mechanism to force open source themes to trump internal themes. Luckily, if we allow both aggregation and generation, we get that import order for free.

### Order of Execution Per Theme
1. Attempt to aggregate themes first: https://github.com/cerner/terra-toolkit/blob/master/scripts/aggregate-themes/theme-aggregator.js#L59-L62
2. If no theme files found, attempt to generate themes: https://github.com/cerner/terra-toolkit/blob/master/scripts/aggregate-themes/theme-aggregator.js#L64-L83

## Potential Solution

Alter `aggregateTheme` to return an asset array containing both aggregated and generated themes. We remove the fallback line [[1](https://github.com/cerner/terra-toolkit/blob/master/scripts/aggregate-themes/theme-aggregator.js#L64)] and add the generated scss to the asset array, rather than returning it [[2](https://github.com/cerner/terra-toolkit/blob/master/scripts/aggregate-themes/theme-aggregator.js#L78)]. If we keep this same execution order, the generated themes will be imported after aggregated ones.

yuderekyu:
### Summary

Closes https://github.com/cerner/terra-toolkit/issues/368

theme-aggregator will now *always* generate root or scope theme files after aggregating an internal theme. Previously, theme-aggregator would only generate root or scope themes if no internal theme existed.

I moved the functions into a top-down format and re-wrote some logic to increase cohesion. I realize some of that is out of scope - if the changes are an eyesore, I can move the re-ordering changes into a separate branch and open a separate PR after this merges.

### Verification Pull Requests

* https://github.com/cerner/terra-core/pull/2792
* Internal worfklow_server pull request

@cerner/terra


scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -33,127 +34,121 @@ class ThemeAggregator {
...
+   * Aggregates theme assets into a js file.
    * @param {Object} options - The aggregation options.
-   * @param {boolean} defaultFlag - Whether the theme to be generated is a root or scope theme. Guards against default theme and scope theme being equivalent.
-   * @returns {array} - An array of file names.
+   * @returns {array} - An array of aggregated theme files
    */
-  static aggregateTheme(themeName, options = {}, defaultFlag) {
-    const { theme, scoped = [] } = options;
+  static aggregateThemes(options) {
+    if (!ThemeAggregator.validate(options)) return null;
```
emilyrohrbough: Can we add braces around these ifs

yuderekyu: https://github.com/cerner/terra-toolkit/pull/376/commits/3ad5c5a087f95ed4e09c6caff52af29ad9a4a250

I am a little torn - I want to be succinct since these can be one-liners, but I understand the argument that brackets improve maintainability and extensibility. I am curious what others think?

StephenEsser: My personal preference is to always include brackets for if statements. I think they're easier to read and understand.

tbiethman:
FWIW I validated this change within my orion-application branch, works like a charm.

scripts/aggregate-themes/README.md
```diff
@@ -37,7 +37,7 @@ project
         ├── component-2.scss
         └── terra-dark-theme.scss
 ```
-*Note*: To support previous functionality, if a project contains existing `root` or `scoped` theme files within namespaced theme directories, terra-toolkit will aggregate these and skip generation. Toolkit will default to theme generation on the 6.0 release.
+*Note*: To support previous functionality, if a project contains existing `root.scss` or `scoped.scss` theme files within namespaced theme directories, terra-toolkit will aggregate those theme files first, then generate theme files second. Based on this import order, generated theme files take precedence over aggregated theme files. Toolkit will default to theme generation on the 6.0 release.
```
emilyrohrbough: > those theme files first, then generate theme files second.
This is a little confusing to me.  Why aggregate a project's theme first, but then the genearted files take precedence over the defined theme?

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -113,21 +48,18 @@ class ThemeAggregator {
    * @returns {array} - An array of aggregated theme files
    */
   static aggregateThemes(options) {
-    if (!ThemeAggregator.validate(options)) {
-      return null;
-    }
+    if (!ThemeAggregator.validate(options)) return null;
 
-    // Create generated-themes directory.
```
emilyrohrbough, line 120: Should this comment be removed?

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -113,21 +48,18 @@ class ThemeAggregator {
...
     fs.ensureDir(OUTPUT_DIR, (err) => {
-      Logger.warn(err, { context });
+      Logger.warn(err, { LOG_CONTEXT });
```
emilyrohrbough, line 62: If this doesn't exist, do we create it somewhere or return null?

scripts/aggregate-themes/README.md
yuderekyu: I re-worded this, does that help clear the confusion? https://github.com/cerner/terra-toolkit/pull/376/commits/4c80666bf163c0a3509a771e3ce01ea969f1bb33#diff-a385b55d94aefec6207446ff49b50d96R40

What I forgot to mention is the CSS import order follows that order.
```js
import 'orion-fusion-theme/src/scss/root-theme.scss'; // existing root theme
import './root-orion-fusion-theme.scss'; // generated file. Takes precedence because it is imported after existing root theme.
```
On second thought, maybe I should include this example in the doc

scripts/aggregate-themes/theme-aggregator.js
yuderekyu: Inserted a comment back https://github.com/cerner/terra-toolkit/pull/376/commits/4c80666bf163c0a3509a771e3ce01ea969f1bb33#diff-1672fd55492cdce050f5aa314b17029bR53

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -156,6 +91,84 @@ class ThemeAggregator {
...
+    Logger.log(`Aggregating ${themeName} files...`, { LOG_CONTEXT });
+    const { theme, scoped = [] } = options;
+    const isRoot = themeName === theme && defaultFlag;
+    const isScoped = scoped.indexOf(themeName) > -1;
+    const themeScope = { isRoot, isScoped };
+
+    const aggregatedAssets = ThemeAggregator.aggregateTheme(themeName, themeScope, options);
+    const generatedAsset = ThemeAggregator.generateTheme(themeName, themeScope, options);
+    // Generated theme file should take precedence over aggregated files.
+    // Therefore, take advantage of css import precdence by adding the generated asset last.
```
emilyrohrbough: I'm confused, what is this the difference between the aggregated theme and the generated theme? I would have thought these were conceptually the sam thing, only placing the imports in one file?

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -156,6 +91,84 @@ class ThemeAggregator {
...
+    if (generatedAsset) {
+      aggregatedAssets.push(generatedAsset);
+    }
+
+    if (!aggregatedAssets.length) {
```
emilyrohrbough, line 117: Should this if check mach the `generatedAsset` if check?

scripts/aggregate-themes/theme-aggregator.js
yuderekyu: ensureDir creates the directory if it does not exist - if it exists, I don't think it does anything

scripts/aggregate-themes/theme-aggregator.js
yuderekyu: yeah we have an issue with semantics. We "aggregate" root and scoped theme files, "aggregate" theme-name files then "generate" root or scoped theme files. The aggregated root/scoped theme files and generated root/scoped theme files are then imported into the js file. 

Maybe I should reword "aggregated files" to "existing root and scoped theme files" here?

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -195,37 +208,29 @@ class ThemeAggregator {
...
    */
-  static writeSCSSFile({
-    assets, themeName, prefix, scopeSelector, outputPath,
-  }) {
-    const fileName = `${prefix}-${themeName}.scss`;
-    const intro = `${DISCLAIMER}${scopeSelector}`;
-
-    let file = assets.reduce((acc, s) => `${acc}  @import '../${s}';\n`, '');
-    file = `${intro} {\n${file}}\n`;
+  static findThemeVariableFilesForGeneration(themeName, options = {}) {
```
emilyrohrbough, line 223: This looks one-to-one to the previous? Why the re-ordering?

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -195,37 +208,29 @@ class ThemeAggregator {
...
-   *   @param {string} prefix - Prefix to append to generated file.
-   *   @param {string} scopeSelector - scss scope selector to encase theme.
-   *   @param {string} outputPath - path to write the scss file to. For testing purposes - overrides the default generated-themes path.
-   * @returns {string} - The path of the generated scss file, relative to the working home directory.
+   * Searches for theme files located within the following directory format: themes->theme-name->theme-name.scss.
+   * @param {string} themeName - The theme to aggregate.
+   * @param {Object} options - The aggregation options.
+   * @returns {array} - An array of ${themeName} files.
    */
-  static writeSCSSFile({
```
emilyrohrbough, line 215: Why remove this function? It looks like it's been moved to `generateTheme`?

scripts/aggregate-themes/theme-aggregator.js
yuderekyu: I mentioned this in the pull request description, but re-ordered the functions into a top-down execution format. I find this is easier to follow than the previous order

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -156,6 +91,84 @@ class ThemeAggregator {
...
+   * @param {boolean} defaultFlag - Whether the theme to be generated is a root or scope theme. Guards against default theme and scope theme being equivalent.
+   * @returns {array} - An array of file names.
+   */
+  static triggerAggregationAndGeneration(themeName, options = {}, defaultFlag) {
+    if (!themeName) return null;
+    Logger.log(`Aggregating ${themeName} files...`, { LOG_CONTEXT });
+    const { theme, scoped = [] } = options;
+    const isRoot = themeName === theme && defaultFlag;
+    const isScoped = scoped.indexOf(themeName) > -1;
+    const themeScope = { isRoot, isScoped };
```
emilyrohrbough: can we condense this to 
```
const themeScope = { isScoped: !isRoot };
```
It looks like this check is used multiple times throughout

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -139,14 +72,16 @@ class ThemeAggregator {
     }
 
     const assets = [];
-    let asset;
     themesToAggregate.forEach((theme) => {
-      asset = ThemeAggregator.aggregateTheme(theme, options, defaultFlag);
+      const asset = ThemeAggregator.triggerAggregationAndGeneration(theme, options, defaultFlag);
```
emilyrohrbough: Why the defaultFlag? Could we pass the boolean like: 
```diff
+ const isRootTheme = theme === defaultTheme;
+ const asset = ThemeAggregator.triggerAggregationAndGeneration(theme, options, isRootTheme);
- const asset = ThemeAggregator.triggerAggregationAndGeneration(theme, options, defaultFlag);
```

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -156,6 +91,84 @@ class ThemeAggregator {
...
+   * @param {Object} options - The aggregation options.
+   * @param {boolean} defaultFlag - Whether the theme to be generated is a root or scope theme. Guards against default theme and scope theme being equivalent.
+   * @returns {array} - An array of file names.
+   */
+  static triggerAggregationAndGeneration(themeName, options = {}, defaultFlag) {
+    if (!themeName) return null;
+    Logger.log(`Aggregating ${themeName} files...`, { LOG_CONTEXT });
+    const { theme, scoped = [] } = options;
+    const isRoot = themeName === theme && defaultFlag;
+    const isScoped = scoped.indexOf(themeName) > -1;
```
emilyrohrbough: Won't this always be true unless this is the root theme we are aggregating?

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -37,74 +38,8 @@ class ThemeAggregator {
         return ThemeAggregator.writeJsFile(assets);
       }
     }
-    return null;
```
emilyrohrbough, line 40: I found a bug with the aggregate function when trying to default the theme in webpack. It wouldn't generate the theme because I didn't have a theme configuration && the reason I wanted to default the theme was so consumer did not need one 😅  I think the following needs changed:

```diff
static aggregate(theme) {
-   let themeConfig;
+  let themeConfig = {};
    const defaultConfig = path.resolve(process.cwd(), CONFIG);
    if (fs.existsSync(defaultConfig)) {
      // eslint-disable-next-line global-require, import/no-dynamic-require
      themeConfig = require(defaultConfig);
-      const assets = ThemeAggregator.aggregateThemes({ ...themeConfig, ...theme && { theme } });
-      if (assets) {
-        return ThemeAggregator.writeJsFile(assets);
-      }
    }
+   const assets = ThemeAggregator.aggregateThemes({ ...themeConfig, ...theme && { theme } });
+   if (assets) {
+     return ThemeAggregator.writeJsFile(assets);
+   }
    return null;
```

scripts/aggregate-themes/theme-aggregator.js
yuderekyu: To improve `generateTheme`'s cohesion. Calling `writeSCSSFile` seems to make `generateTheme` a shell of a function. I can see the other side of the argument - keeping `writeSCSSFile` would be better for maintainability and extensibility.

If we kept the same structure, `generateTheme` would call `findThemeVariableFilesForGeneration`, pack the parameters into an object, pass this object into `writeSCSSFile`,  then return the value returned from `writeSCSSFile`. `writeSCSSFile` seems a little too coupled with `generatedThemes`.

What are your thoughts?
```jsx
// Alternative
static generateTheme(themeName, themeScope, options) {
const fileAttrs = {
      assets: ThemeAggregator.findThemeVariableFiles(themeName, options),
      themeName,
      prefix: isScoped && !isRoot ? SCOPED : ROOT,
      scopeSelector: isScoped && !isRoot ? `.${themeName}` : `:${ROOT}`,
    };

if (fileAttrs.assets) {
  return ThemeAggregator.writeSCSSFile(fileAttrs);
}

return null;
```

scripts/aggregate-themes/theme-aggregator.js
yuderekyu: hey, the doc explicitly says you need a terra-theme config! hahaa jk, i will update this

yuderekyu: https://github.com/cerner/terra-toolkit/pull/376/commits/1d1bb30aefe21cc8976ec3b164fd57947e4299bb#diff-1672fd55492cdce050f5aa314b17029bR31

scripts/aggregate-themes/theme-aggregator.js
yuderekyu: Ok, the defaultFlag, isRoot, and isScoped flags I use are just...bad 😓 . Here's my shot at refactoring. Use a single flag to indicate if a theme is root. If that's false, we know it's scoped.

https://github.com/cerner/terra-toolkit/pull/376/commits/1d1bb30aefe21cc8976ec3b164fd57947e4299bb#diff-1672fd55492cdce050f5aa314b17029bR72

scripts/aggregate-themes/theme-aggregator.js
yuderekyu: Flattened this to a `isRoot` flag. Use this to flex between root and scoped themes.
* https://github.com/cerner/terra-toolkit/pull/376/commits/1d1bb30aefe21cc8976ec3b164fd57947e4299bb#diff-1672fd55492cdce050f5aa314b17029bR72
* https://github.com/cerner/terra-toolkit/pull/376/commits/1d1bb30aefe21cc8976ec3b164fd57947e4299bb#diff-1672fd55492cdce050f5aa314b17029bR108

scripts/aggregate-themes/theme-aggregator.js
yuderekyu: Flattened this to a isRoot flag. Use this to flex between root and scoped themes.

* https://github.com/cerner/terra-toolkit/pull/376/commits/1d1bb30aefe21cc8976ec3b164fd57947e4299bb#diff-1672fd55492cdce050f5aa314b17029bR72
* https://github.com/cerner/terra-toolkit/pull/376/commits/1d1bb30aefe21cc8976ec3b164fd57947e4299bb#diff-1672fd55492cdce050f5aa314b17029bR108

scripts/aggregate-themes/theme-aggregator.js
yuderekyu: no, we want an existence check on all assets to correctly say whether no theme files were found, regardless if they were existing root/scoped or generated root/scoped files

emilyrohrbough (approved):

scripts/aggregate-themes/theme-aggregator.js
emilyrohrbough: THANKS 😄

tbiethman (approved):

BenBoersma (approved):

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -27,133 +28,144 @@ class ThemeAggregator {
...
+    if (!ThemeAggregator.validate(options)) {
       return null;
     }
 
-    const isRoot = themeName === theme && defaultFlag;
-    const isScoped = scoped.indexOf(themeName) > -1;
-    const file = isScoped && !isRoot ? SCOPED_THEME : ROOT_THEME;
-    const assets = ThemeAggregator.find(`**/themes/${themeName}/${file}`, options);
+    // This creates a generated-themes directory to house theme assets.
+    fs.ensureDir(OUTPUT_DIR, (err) => {
```
emilyrohrbough: It took me a bit to track this down, but when validating my webpack configuration with Jest, I was seeing this error:

Screen Shot 2020-01-22 at 9 22 33 AM

Could we update this to use Promise syntax to resolve this error? The change I tested looks like:
```js
fs.ensureDir(OUTPUT_DIR,).catch((err) => {
    Logger.warn(err, { LOG_CONTEXT });
});
```

yuderekyu: oops, good catch https://github.com/cerner/terra-toolkit/pull/376/commits/7a3892b2f4e5b5c2a5153b76c9d4f736c696596a

emilyrohrbough: much thanks!

PR #383, opened 2020-02-11
Add TERRA_AGGREGATED_LOCALES global to the webpack bundle

### LINKED ISSUE #382: Add a global to the webpack configuration that lists supported locales
# Feature Request

## Description

Add a webpack define plugin global that provides an array of locales that were successfully aggregated and are available to the browser.

Example of existing global constant in use today: 

https://github.com/cerner/terra-toolkit/blob/1a8909003b5c61d52b1f443883b45a6330584afd/config/webpack/webpack.config.js#L137-L139

Terra aggregate translations is being updated to return the list of locales the script was able to successfully generate messages and loaders for. This list of locales should be provided to the webpack bundle so that the client is aware of the supported locales.

https://github.com/cerner/terra-aggregate-translations/pull/19

Some applications may choose to fallback to the preferred browser language. The preferred language should be validated against the list of supported locales before being accepted.

Example of interpreting the browsers preferred language.

```
const locale = (navigator.languages && navigator.languages[0])
               || navigator.language
               || navigator.userLanguage
               || 'en';
```

https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages

https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language

StephenEsser:
### Summary

This PR adds a SUPPORTED_LOCALES global to the webpack bundle that includes the list of locales that were successfully aggregated. This gives the client access to the supported locales list to allow the preferred browser locale safety be used if it is a supported locale.

### Additional Details

This is a sibling PR to the changes in terra-aggregate-translations:  https://github.com/cerner/terra-aggregate-translations/pull/19

If locale aggregation is disabled SUPPORTED_LOCALES will be undefined.

Resolves #382 

@cerner/terra

### DIFF
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -18,6 +18,7 @@ const webpackConfig = (options, env, argv) => {
     rootPath,
     resolveModules,
     staticOptions,
+    supportedLocales,
     themeFile,
   } = options;
 
@@ -137,6 +138,9 @@ const webpackConfig = (options, env, argv) => {
       new webpack.DefinePlugin({
         CERNER_BUILD_TIMESTAMP: JSON.stringify(new Date(Date.now()).toISOString()),
       }),
+      new webpack.DefinePlugin({
+        SUPPORTED_LOCALES: JSON.stringify(supportedLocales),
+      }),
     ],
     resolve: {
       extensions: ['.js', '.jsx'],
@@ -210,8 +214,9 @@ const defaultWebpackConfig = (env = {}, argv = {}) => {
 
   const resolveModules = ['node_modules'];
 
+  let supportedLocales;
   if (!disableAggregateTranslations) {
-    aggregateTranslations({ baseDir: rootPath, ...env.aggregateOptions });
+    supportedLocales = aggregateTranslations({ baseDir: rootPath, ...env.aggregateOptions });
     resolveModules.unshift(path.resolve(rootPath, 'aggregated-translations'));
   }
 
@@ -222,6 +227,7 @@ const defaultWebpackConfig = (env = {}, argv = {}) => {
     rootPath,
     resolveModules,
     staticOptions,
+    supportedLocales,
     themeFile,
   };
 
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -139,7 +139,7 @@ const webpackConfig = (options, env, argv) => {
         CERNER_BUILD_TIMESTAMP: JSON.stringify(new Date(Date.now()).toISOString()),
       }),
       new webpack.DefinePlugin({
-        SUPPORTED_LOCALES: JSON.stringify(supportedLocales),
+        TERRA_SUPPORTED_LOCALES: JSON.stringify(supportedLocales),
       }),
     ],
     resolve: {
--- a/config/webpack/webpack.config.js
+++ b/config/webpack/webpack.config.js
@@ -137,7 +137,7 @@ const webpackConfig = (options, env, argv) => {
       }),
       new webpack.DefinePlugin({
         CERNER_BUILD_TIMESTAMP: JSON.stringify(new Date(Date.now()).toISOString()),
-        TERRA_SUPPORTED_LOCALES: JSON.stringify(supportedLocales),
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -95,7 +95,7 @@
     "postcss-rtl": "^1.3.3",
     "raw-loader": "^3.0.0",
     "sass-loader": "^7.0.1",
-    "terra-aggregate-translations": "^1.0.0",
+    "terra-aggregate-translations": "git+https://github.com/cerner/terra-aggregate-translations.git#supported-locales",
```
StephenEsser: Sibling PR: https://github.com/cerner/terra-aggregate-translations/pull/19

These changes can technically be merged without the changes in terra-aggregate-translations, but the intention would be to bump the minimum version to the version that releases with these changes so that supported locales won't be undefined.

config/webpack/webpack.config.js
```diff
@@ -137,6 +138,9 @@ const webpackConfig = (options, env, argv) => {
       new webpack.DefinePlugin({
         CERNER_BUILD_TIMESTAMP: JSON.stringify(new Date(Date.now()).toISOString()),
       }),
+      new webpack.DefinePlugin({
+        SUPPORTED_LOCALES: JSON.stringify(supportedLocales),
```
emilyrohrbough: Who is this consumed since this is converted to a string?

config/webpack/webpack.config.js
```diff
@@ -210,8 +214,9 @@ const defaultWebpackConfig = (env = {}, argv = {}) => {
 
   const resolveModules = ['node_modules'];
 
+  let supportedLocales;
```
emilyrohrbough: Should this default be []?

config/webpack/webpack.config.js
StephenEsser: It can be used as an array if it is defined. Here's a little more information about how the define plugin works. 

https://webpack.js.org/plugins/define-plugin/

Here's a little test I did locally.

```jsx
  console.log(SUPPORTED_LOCALES); // Prints an array of locales if not undefined

  if (SUPPORTED_LOCALES) {
    SUPPORTED_LOCALES.forEach((locale) => {
      console.log(locale); // Prints each locale
    });
  }
```

config/webpack/webpack.config.js
StephenEsser: It could be, but I don't think it would provide any additional value as opposed to undefined unless we can guarantee each consuming team is using >=  the version these changes will release on.

config/webpack/webpack.config.js
mjhenkes: I may be paranoid, but i like to namespace things to be unique to us as possible. How about...
```suggestion
        TERRA_SUPPORTED_LOCALES: JSON.stringify(supportedLocales),
```

mjhenkes (approved):

config/webpack/webpack.config.js
StephenEsser: That works for me: https://github.com/cerner/terra-toolkit/pull/383/commits/4bcd94ab3d5687509bed4dd444bdfd4cfce0eb70

config/webpack/webpack.config.js
```diff
@@ -137,6 +138,9 @@ const webpackConfig = (options, env, argv) => {
       new webpack.DefinePlugin({
         CERNER_BUILD_TIMESTAMP: JSON.stringify(new Date(Date.now()).toISOString()),
       }),
+      new webpack.DefinePlugin({
+        TERRA_SUPPORTED_LOCALES: JSON.stringify(supportedLocales),
```
tbiethman: You can add this as another key in the previous DefinePlugin.

tbiethman (approved):

config/webpack/webpack.config.js
StephenEsser: This checks out: https://github.com/cerner/terra-toolkit/pull/383/commits/de78977c4a6996477f242772d91449158353e46c

dkasper-was-taken (approved):

ryanthemanuel (approved):

mjhenkes (approved):

emilyrohrbough (approved):

PR #386, opened 2020-02-17
Correct theme-aggregation logging prefix.

yuderekyu:
### Summary

Oops, this fixes theme-aggregation logs to correctly have `[theme-aggregation]` as a prefix to logs.

Verification PR: https://github.com/cerner/terra-clinical/pull/610

### Additional Details

@cerner/terra

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Unreleased
 * Added `TERRA_AGGREGATED_LOCALES` global to the webpack bundle
 
 ### Fixed
-* Re-add log prefix to theme-aggregator
+* Fixed and updated theme-aggregator prefix to [Terra-Toolkit:theme-aggregator] to be consistent with logs from other WDIO services.
 
 5.20.0 - (January 28, 2020)
 ----------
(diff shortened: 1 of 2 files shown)


emilyrohrbough (approved):
🎉 yay. this has been bugging me!

StephenEsser (approved):
Since this is restoring previous behavior I wouldn't worry too much about it. But I do think it's strange the context/prefix is `[terra-theme-aggregator]`, which isn't a real package/identifiable library.  [terra-toolkit] or something similar might make more sense from a discovery/association perspective. Searching Github/Google for terra-theme-aggregator would return zero results, except maybe the const declaration.

yuderekyu:
@StephenEsser Very true - I am down to update the prefix to something like `[terra-toolkit: theme-aggregator]` or `[terra-toolkit]`. Any thoughts on the specific prefix?

StephenEsser:
Either one sounds good to me.

emilyrohrbough:
If you go with `[terra-toolkit:theme-aggregator]` this will match how we've named-spaced the wdio services when they log output

yuderekyu:
Sounds good, updated to `[Terra-Toolkit:theme-aggregator]` https://github.com/cerner/terra-toolkit/pull/386/commits/138468a875f374f4e55da07294b5545167f6b5dd

mjhenkes (approved):

PR #390, opened 2020-03-18
Postcss theme plugin

mjhenkes:
### Summary
Finally, as promised, the terra post css theme plugin.

#### Plugin
The plugin itself is pretty straight forward. It reads the same config as aggregate translations. watches for the default theme and applies it to root. It also watches for supported themes that are not marked as scoped in the config and removes them.

#### aggregate-themes
The changes here allow for passing an overriding config file (more than just default theme) and for a flag that aggregates the default theme as a scoped theme instead of as a root theme. This assumes the plugin is being run to turn that scoped theme into a root theme.
The config override was added so that we have more control over theme aggregation during testing.

#### Webpack
A new env was added to the webpack config to disable theme aggregation. This will allow projects to dis-continue theme aggregation once all of their components have been converted to the new style of including themes.

Once all components have been converted we could remove the `getThemeWebpackPromise` in a major version bump.

#### Testing
To improve testing times, scoped themes will not be aggregated during test runs if an override theme is specified. This seems to reduce build times by about 30 seconds per theme aggregated. I attempted to disable theme aggregation for testing all-together but determined it would not work in the case where a default theme is specified in the theme config.

All these changes are passive, though the testing change could be considered non-passive if someone is testing switching to the scoped theme (which wouldn't work in IE). I believe this to be unlikely.

#### New css include style
For this plugin to work css needs to be included in such a way that it can be processed by webpack. A simple example of this is presented in the theme plugin readme.

Check out the core PR for examples for both button and card:
Core: https://github.com/cerner/terra-core/pull/2857

#### Test PRs
Core: https://github.com/cerner/terra-core/pull/2857
Framework: https://github.com/cerner/terra-framework/pull/1051
Clinical: https://github.com/cerner/terra-clinical/pull/612

This has been tested internally in workflow server to ensure that these changes still work with both code splitting and IE. They do.

@cerner/terra


mjhenkes:
Terra toolkit's jest job is 2 minutes faster by mocking out glob in the aggregate-themes jest test.

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ Changelog
 
 Unreleased
 ----------
+### Added
+* Include a postcss plugin to apply a scoped theme as the default theme and to remove unlisted supported themes.
+
+### Changed
+* For testing, only aggregate the default theme.
```
emilyrohrbough, line 10: I'm noticing we are leasing our test themes 😬 can we add to the npm ignore?

mjhenkes: you mean we're releasing the generated themes folder right? the tests folder is already ignored.

scripts/postcss/README.md
```diff
@@ -0,0 +1,89 @@
+# Terra Postcss Theme plugin
+
+The purpose of this plugin is to create a default theme from a scoped theme and to remove any known themes that are not desired.
+
+The plugin will read from the same terra-theme.config.js that terra-aggregate translations reads from. Eventually terra-aggregate translations will be deprecated in favor of this strategy.
```
emilyrohrbough:
```suggestion
The plugin will read from the same terra-theme.config.js that terra-aggregate-themes script reads from. Eventually terra-aggregate-themes will be deprecated in favor of this strategy.
```

scripts/postcss/README.md
```diff
@@ -0,0 +1,89 @@
...
+    {children}
+  </div>
+);
+
+export default Component;
+```
+
+The scss file includes the themefile and applies the css variable to the css property. Breaking the theme variable out into their own files is purely convention and not required for this plugin. By convention we name the file ```<component>.<theme>.module.scss```.
+
+component.module.scss
```
emilyrohrbough: Maybe Italics or a sub header to differentiate these file names

scripts/postcss/README.md
```diff
@@ -0,0 +1,89 @@
...
+  }
+}
+```
+
+The theme file declares the ```.orion-fusion-theme``` as a global class and defines the css variable.
+
+component.orion-fusion-theme.module.scss
+```scss
+:global {
+  /* stylelint-disable suitcss/custom-property-no-outside-root */
```
emilyrohrbough:
```suggestion
  /* stylelint-disable suitcss/custom-property-no-outside-root */
```
Should we delete this & update our stylelint config to ignore these patterns?

scripts/postcss/README.md
mjhenkes: lol

config/webpack/webpack.config.js
```diff
@@ -218,15 +222,25 @@ const defaultWebpackConfig = (env = {}, argv = {}) => {
...
-  const themeFile = ThemeAggregator.aggregate(themeOverride);
+  const defaultTheme = process.env.THEME; // Flexes root theme for theme visual regression testing.
+  const themeConfig = defaultTheme ? { theme: defaultTheme } : null;
+  let themeFile;
+  if (!disableAggregateThemes) {
+    // Set the default theme and disable scoped theme aggregation.
+    themeFile = ThemeAggregator.aggregate(
+      null,
+      themeConfig,
+      true,
```
yuderekyu: can we add a comment signifying what this flag is doing?

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -27,17 +27,22 @@ class ThemeAggregator {
    * @param {string} theme - The theme to override the default theme. Used for visual regression testing.
    * @returns {string|null} - The output path of the aggregated theme file. Null if not generated.
    */
-  static aggregate(theme) {
+  static aggregate(theme, config, aggregateDefaultThemeAsScopedTheme) {
```
yuderekyu: let's add the new params into the jsdoc

scripts/postcss/ThemePlugin.js
```diff
@@ -0,0 +1,58 @@
...
+  // Find the set of known themes that should not be included.
+  const themesToRemove = SUPPORTED_THEMES.reduce((acc, theme) => {
+    if (!scopedThemes.includes(theme)) {
+      acc.push(`.${theme}`);
+    }
+    return acc;
+  }, []);
+
+  return (root) => {
+    if (defaultThemeSelector || themesToRemove.length > 0) {
```
yuderekyu: not a big deal, but i believe a length existence check does the same thing
```suggestion
    if (defaultThemeSelector || themesToRemove.length) {
```

yuderekyu (approved):

scripts/postcss/README.md
mjhenkes: https://github.com/cerner/stylelint-config-terra/pull/36

config/webpack/webpack.config.js
mjhenkes: I broke my own rules here, I prefer taking a single object as params so that they can be explicitly name. I'll switch the second param to an object so the API is self documenting.

mjhenkes:
I added the Do not merge label, I'm prototyping some more ideas.

scripts/postcss/README.md
```diff
@@ -35,14 +35,18 @@ Component.jsx
 ```jsx
 import React from 'react';
 import styles from './component.module.scss';
+import ThemeContext from 'terra-theme-context';
```
emilyrohrbough: do we plan to add this context to terra-theme-provider?

mjhenkes: Thats my post scrum/war room topic

mjhenkes: The plan is to add a context to terra-framework as a standalone context, and then re-provide it from terra-application

mjhenkes: terra-theme-context pr: https://github.com/cerner/terra-framework/pull/1087

config/webpack/getThemeWebpackPromise.js
```diff
@@ -93,8 +94,9 @@ const themeConfig = (rootPath, themeFile, cachedObject) => (
  * themeable variables and values.
  * @param {String} rootPath the root path from where the webpack is being run
  * @param {String} themeFile the file to be webpacked
+ *  * @param {Object} themeConfig configuration to pass to the post css theme plugin
```
emilyrohrbough:
```suggestion
 *  @param {Object} themeConfig configuration to pass to the post css theme plugin
```

config/webpack/webpack.config.js
```diff
@@ -12,6 +12,7 @@ const DuplicatePackageCheckerPlugin = require('@cerner/duplicate-package-checker
 const aggregateTranslations = require('terra-aggregate-translations');
 const ThemeAggregator = require('../../scripts/aggregate-themes/theme-aggregator');
 const getThemeWebpackPromise = require('./getThemeWebpackPromise');
+const ThemePlugin = require('../../scripts/postcss/ThemePlugin');
```
emilyrohrbough: If this is a webpack plugin, should it live under `config/webpack/plugins`?

mjhenkes: Well, technically it's a postcss plugin that the postcss webpack loader uses... 🤷‍♂

emilyrohrbough:
@mjhenkes  I can't help it, I get a kick outta the fact you keep saying aggregate-translations with these changes instead of aggregate-themes 😂 😄 

> The plugin itself is pretty straight forward. It reads the same config as aggregate translations.

config/webpack/webpack.config.js
emilyrohrbough: my vote is to at least move this to the config/webpack to keep these concepts close

config/webpack/webpack.config.js
```diff
@@ -218,15 +221,27 @@ const defaultWebpackConfig = (env = {}, argv = {}) => {
...
+    );
+  }
 
   const options = {
     rootPath,
     resolveModules,
     staticOptions,
     aggregatedLocales,
     themeFile,
+    defaultTheme,
```
emilyrohrbough: Why not pass the themeConfig object created on line 225 instead of re-creating it in the [webpack config](https://github.com/cerner/terra-toolkit/pull/390/files#diff-4acfd133078383adaf48ec081882c927R33)?

config/webpack/webpack.config.js
```diff
@@ -28,6 +30,7 @@ const webpackConfig = (options, env, argv) => {
   const filename = argv['output-filename'] || fileNameStategy;
   const outputPath = argv['output-path'] || path.join(rootPath, 'build');
   const publicPath = argv['output-public-path'] || '';
+  const themeConfig = defaultTheme ? { theme: defaultTheme } : undefined;
```
emilyrohrbough: If this filters out unspecified themes, how does not include all the themes when packing production assets? Is this configuration specifically for testing?

config/webpack/webpack.config.js
```diff
@@ -218,15 +221,27 @@ const defaultWebpackConfig = (env = {}, argv = {}) => {
     resolveModules.unshift(path.resolve(rootPath, 'aggregated-translations'));
   }
 
-  const themeOverride = process.env.THEME; // Flexes root theme for theme visual regression testing.
-  const themeFile = ThemeAggregator.aggregate(themeOverride);
+  const defaultTheme = process.env.THEME; // Flexes root theme for theme visual regression testing.
+  const themeConfig = defaultTheme ? { theme: defaultTheme } : null;
+  let themeFile;
+  if (!disableAggregateThemes) {
```
emilyrohrbough, line 226: This env name makes me think all theme aggregate should be disabled via the aggregator script & the postcss plugin. Should it disable both?

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -64,11 +71,22 @@ class ThemeAggregator {
 
     const {
       theme: defaultThemeToAggregate,
-      scoped: scopedThemesToAggregate,
+      aggregateDefaultThemeAsScopedTheme,
     } = options;
 
+    let scopedThemesToAggregate = options.scoped || [];
```
emilyrohrbough:
```suggestion
     scoped,
      aggregateDefaultThemeAsScopedTheme,
    } = options;

    let scopedThemesToAggregate = scoped || [];
```

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -64,11 +71,22 @@ class ThemeAggregator {
...
+
+    // The default theme is created by the post css theme plugin.
+    if (aggregateDefaultThemeAsScopedTheme && !scopedThemesToAggregate.includes(defaultThemeToAggregate)) {
+      scopedThemesToAggregate = [
+        defaultThemeToAggregate,
+        ...scopedThemesToAggregate,
+      ];
```
emilyrohrbough:
```suggestion
      scopedThemesToAggregate.push(defaultThemeToAggregate);
```

scripts/aggregate-themes/theme-aggregator.js
```diff
@@ -25,19 +25,26 @@ class ThemeAggregator {
...
+      const defaultConfig = path.resolve(process.cwd(), CONFIG);
+      if (fs.existsSync(defaultConfig)) {
+        // eslint-disable-next-line global-require, import/no-dynamic-require
+        themeConfig = require(defaultConfig);
+      }
     }
 
     const themeAssets = ThemeAggregator.aggregateThemes({
       ...themeConfig,
       ...theme && { theme },
```
emilyrohrbough, line 46: Could we check for `theme` and `aggregateDefaultThemeAsScopedTheme` here to remove the added logic below in ThemeAggregator.aggregateThemes?

emilyrohrbough: Nevermind. I see below we switch between root vs scope aggregation

scripts/aggregate-themes/theme-aggregator.js
emilyrohrbough: Do we need to update this check for length since the fallback is `[]`? https://github.com/cerner/terra-toolkit/pull/390/files#diff-1672fd55492cdce050f5aa314b17029bR96

scripts/postcss/README.md
```diff
@@ -0,0 +1,99 @@
+# Terra Postcss Theme plugin
+
+The purpose of this plugin is to create a default theme from a scoped theme and to remove any known themes that are not desired.
+
+The plugin will read from the same terra-theme.config.js that terra-aggregate-themes script reads from. Eventually terra-aggregate-themes will be deprecated in favor of this strategy.
```
emilyrohrbough:
```suggestion
The plugin will read from the same `terra-theme.config.js` that `terra-aggregate-themes` script reads from. Eventually `terra-aggregate-theme`s will be deprecated in favor of this strategy.
```

scripts/postcss/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Usage
+
+To use this plugin you must have strong conventions around your theme name, include the theme class in your css and add the postcss plugin.
+
+### CSS
+
+This plugin makes the assumption that you are declaring theme variables under a global css class i.e. ```.orion-fusion-theme``` and these theme files are included in such a way that they are processed by webpack.
+
+Consider the following component. The react component pulls in and applies the css styles to div.
```
emilyrohrbough:
```suggestion
Consider the following example. The React component, `Component.jsx` pulls in and applies css styles from `component.module.scss` to the wrapper div. You will also notice it is getting the current theme set on the application from the `ThemeContext` and is also applying the current theme name to the div. 

If the current theme was `orion-fusion-theme` the `orion-fusion-theme` classname would be applied and the styles from `component.orion-fusion-theme.module.scss` would be applied since it is imported into `component.module.scss` and was processed by webpack.
```

scripts/postcss/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+Component.jsx
+
+```jsx
+import React from 'react';
```
emilyrohrbough, line 36:
```suggestion

```jsx
// Component.jsx
import React from 'react';
```

scripts/postcss/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+To use this plugin you must have strong conventions around your theme name, include the theme class in your css and add the postcss plugin.
+
+### CSS
+
+This plugin makes the assumption that you are declaring theme variables under a global css class i.e. ```.orion-fusion-theme``` and these theme files are included in such a way that they are processed by webpack.
+
+Consider the following component. The react component pulls in and applies the css styles to div.
+
+Component.jsx
```
emilyrohrbough:
```suggestion
#### Component.jsx
```

config/webpack/webpack.config.js
mjhenkes: The postcss plugin isn't actually doing any aggregating and would eliminate the usefulness of this env. I added it so once we get terra-repos updated we can get the performance boos of removing aggregate-themes without breaking theming tests.

scripts/postcss/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+### terra-theme.config.js
+
+Below is an example of terra-theme.config
+
+```js
+const themeConfig = {
+  exclude: [], // Files to exclude. Accepts glob patterns.
+  theme: 'terra-dark-theme', // The default theme.
+  scoped: ['terra-light-theme', 'terra-lowlight-theme'], // An array of scoped themes.
```
emilyrohrbough: It might be helpful to explain the difference/use-case of theme vs scoped or link the original docs

scripts/postcss/README.md
```diff
@@ -0,0 +1,99 @@
...
+import { ThemeContext } from 'terra-application/lib/theme';
+
+import styles from './component.module.scss';
+
+const cx = classNames.bind(styles);
+
+const Component = ({children}) => {
+  const theme = React.useContext(ThemeContext);
+  return (
+    <div className={cx('div', theme.className)}>
```
emilyrohrbough: For this new theme strategy, does this mean all components will need a wrapping div to correctly flip themes? There are times we use React.Fragement to group component renders. Will we need to apply the global theme name to each element in those cases?

config/webpack/webpack.config.js
mjhenkes: Passivity? 🤷‍♂ I think the change makes sense.

scripts/postcss/ThemePlugin.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  // Add the . to find the selector.
+  const defaultThemeSelector = `.${themeConfig.theme}`;
+
+  const scopedThemes = themeConfig.scoped || [];
+
+  // Find the set of known themes that should not be included.
+  const themesToRemove = SUPPORTED_THEMES.reduce((acc, theme) => {
```
emilyrohrbough: For whatever themes should be included after this step, should they be looped to be transformed from `theme` to `.theme`?

config/webpack/webpack.config.js
mjhenkes: When programmatically setting the default theme you only get the default theme. I see it as specifically for testing, yes.

config/webpack/webpack.config.js
mjhenkes: Ok, it's already kinda weird living in scripts anyway.

scripts/postcss/README.md
mjhenkes: If we're applying styles we have some DOM element we're attaching those styles to. I would expect the uppermost DOM element in the component would have the styles applied to it. Or you could apply appropriate theme variables to each DOM element provided in some sort of react fragment. If your component doesn't have any DOM elements it probably doesn't need themeing :)

scripts/postcss/ThemePlugin.js
mjhenkes: I'm confused, we transform themes to remove below and we don't include any more themes after this step?

emilyrohrbough: I guess I'm assuming the scope list could include a theme not included in our supported list. My thought was they were defined as:
```
scope: [
    'custom-theme',
    'orion-fusion-theme',
]
```
Maybe I'm reading this wrong, but this is to filter out the un-sued themes right? so the themesToRemove value is:
```
themesToRemove = [
    '.clinical-lowlight-theme',
    '.cerner-clinical-theme'
]
```
Won't the remaining scoped themes need to be updated to be:
```
themesToInclude = [
    '.custom-theme',
    '.orion-fusion-theme',
]
```

emilyrohrbough: Actually, we only care about the ones to remove, we aren't dong anything special to pull in the scoped themes because we expect that to happen in the scss file via import. Whoops. I figured it out 👍

package.json
```diff
@@ -79,6 +79,7 @@
     "clean-webpack-plugin": "^3.0.0",
     "commander": "^2.19.0",
     "css-loader": "^3.2.0",
+    "css-selector-tokenizer": "^0.7.2",
```
emilyrohrbough, line 82: These are two known bugs for this dependency: https://github.com/css-modules/css-selector-tokenizer/issues

Any worries that this could be a problem for our theming? I'm assuming no given our recommended/expected usage

mjhenkes: no, neither of those issues affect us

scripts/aggregate-themes/theme-aggregator.js
mjhenkes: We can remove it since it defaults to an array.

scripts/postcss/README.md
mjhenkes: I'll add a stand alone theme config doc, but i'm not sure how i want to link theme yet.

config/webpack/postcss/README.md
```diff
@@ -0,0 +1,100 @@
+# Terra Postcss Theme plugin
+
+The purpose of this plugin is to create a default theme from a scoped theme and to remove any known themes that are not desired.
+
+The plugin will read from the same `terra-theme.config.js` that `terra-aggregate-themes` script reads from. Eventually `terra-aggregate-theme`s will be deprecated in favor of this strategy.
+
+## Configuration
+
+### terra-theme.config.js
+
```
emilyrohrbough, line 10: can we link: https://github.com/cerner/terra-toolkit/pull/390/files#diff-9c1c40ebc66be0d348d616ea7e89ccdfR1?

emilyrohrbough (approved):

config/webpack/postcss/ThemePlugin.js
```diff
@@ -0,0 +1,61 @@
...
+  'orion-fusion-theme',
+  'clinical-lowlight-theme',
+  'cerner-clinical-theme',
+];
+
+/**
+ * The purpose of this plugin is to create a default theme from a scoped theme
+ * and to remove any supported themes that are not desired.
+ */
+module.exports = postcss.plugin('postcss-test-plugin', (config) => {
```
ryanthemanuel: postcss-test-plugin?  is that a convention or is there a better name here?

ryanthemanuel (approved):

config/webpack/postcss/ThemePlugin.js
mjhenkes: there is definitely a better name, that was copied from the guide i started with on writing post css plugins :)

config/webpack/postcss/README.md
```diff
@@ -0,0 +1,100 @@
...
+Consider the following example. The React component, `Component.jsx` pulls in and applies css styles from `component.module.scss` to the wrapper div. You will also notice it is getting the current theme set on the application from the `ThemeContext` and is also applying the current theme name to the div.
+
+If the current theme was `orion-fusion-theme` the `orion-fusion-theme` classname would be applied and the styles from `component.orion-fusion-theme.module.scss` would be applied since it is imported into `component.module.scss` and was processed by webpack.
+
+#### Component.jsx
+
+```jsx
+import React from 'react';
+import { ThemeContext } from 'terra-application/lib/theme';
```
StephenEsser, line 38: Does this exist yet?

mjhenkes: nope, it can't until terra-theme-context gets merged and released https://github.com/cerner/terra-framework/pull/1087

mjhenkes: but this is ideally what it would look like :)

StephenEsser (approved):

PR #408, opened 2020-05-08
[eslint-config-terra] Add full stack specs to lint overrides to allow test globals

yuderekyu:
### Summary

Add full stack spec glob to lint overrides. this will ignore undefined global lint errors for our testing globals

### Additional Details

ideally, full-stack and component level wdio testing should not diverge on what testing technologies they use (mocha, visual regression, etc)
@cerner/terra

### DIFF
--- a/packages/eslint-config-terra/eslint.config.js
+++ b/packages/eslint-config-terra/eslint.config.js
@@ -66,7 +66,7 @@ module.exports = {
       },
     },
     {
-      files: ['**/wdio/**/*-spec.*'],
+      files: ['**/wdio/**/*-spec.*', '**/full-stack/**/*-spec.*'],
       rules: {
         'no-unused-expressions': 'off',
       },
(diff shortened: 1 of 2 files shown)


packages/eslint-config-terra/eslint.config.js
```diff
@@ -66,7 +66,7 @@ module.exports = {
       },
     },
     {
-      files: ['**/wdio/**/*-spec.*'],
+      files: ['**/wdio/**/*-spec.*', '**/full-stack/**/*-spec.*'],
```
emilyrohrbough, line 69: Are we for sure going with full-stack for this concept? You had mentioned potentially wanting to renaming this a week or so ago

yuderekyu: ok, thinking about this more, there are enough unknowns to warrant me closing this

yeah, i discussed changing the name to `integration-testing`, but i backed away from that. in the future, full-stack could potentially be used for application tests, not just engines -  at that point, `system-testing` seems like a better fit. this is also dependent on how we scope terra-toolkit v6. let's revisit this further down the road

PR #412, opened 2020-05-28
[terra-functional-testing] Setup initial docker service

### LINKED ISSUE #410: [terra-functional-testing] Create a Docker Service for setting up the testing environment
# Feature Request

## Description

Create a docker service for setting up the WDIO testing environment. The service should initialize a docker swam that additional nodes can be connected to. This service will be used the facilitate the [selenium-docker](https://github.com/SeleniumHQ/docker-selenium) infrastructure. 

The existing docker service is located [here](https://github.com/cerner/terra-toolkit-boneyard/blob/master/src/wdio/services/SeleniumDockerService.js). The docker service created should carry forward many of the same features and functionality. 

WebDriverIO lists [wdio-docker-service](https://github.com/stsvilik/wdio-docker-service) in their site documentation. This service may provide good information to reference for configuring docker and WebDriver.

StephenEsser:
### Summary

This PR sets up the initial docker service for running tests against selenium docker. This setup uses the minimum configurations needed to run the service. CI integration and any additional configuration options will be added as they become necessary.

Here are resources for viewing the existing selenium docker service implementation:

Docker Service:

https://github.com/cerner/terra-toolkit-boneyard/blob/master/src/wdio/services/SeleniumDockerService.js

Configurations:

https://github.com/cerner/terra-toolkit-boneyard/blob/master/config/wdio/services.default-config.js
https://github.com/cerner/terra-toolkit-boneyard/blob/master/config/wdio/selenium.config.js

Docker:

https://github.com/cerner/terra-toolkit-boneyard/blob/master/config/docker/local-docker-compose.yml

Resolves #410 

### Additional Details

I've brought back the logger to help log out information to the console in a consistent way. WDIO exposes a logger that custom services can implement. We can potentially use or shape our logger after. We'll likely need to implement our own or choose one from open source.

https://github.com/webdriverio/webdriverio/tree/master/packages/wdio-logger

The logger logs text with this format:

> [terra-functional-testing:wdio-docker-service] Initializing docker swarm.

Tests are in progress. More test coverage will be added for the logger and docker service.

@cerner/terra

### DIFF
--- a/packages/terra-functional-testing/lib/config/wdio.conf.js
+++ b/packages/terra-functional-testing/lib/config/wdio.conf.js
@@ -1,3 +1,5 @@
+const DockerService = require('../services/wdio-docker-service');
+
 exports.config = {
   //
   // ====================
@@ -80,6 +82,8 @@ exports.config = {
   // If you only want to run your tests until a specific amount of tests have failed use
   // bail (default is 0 - don't bail, run all tests).
   bail: 0,
+  // Set the path to connect to the selenium container.
+  path: '/wd/hub',
   //
   // Set a base URL in order to shorten url command calls. If your `url` parameter starts
   // with `/`, the base url gets prepended, not including the path portion of your baseUrl.
@@ -101,7 +105,9 @@ exports.config = {
   // Services take over a specific job you don't want to take care of. They enhance
   // your test setup with almost no effort. Unlike plugins, they don't add new
   // commands. Instead, they hook themselves up into the test process.
-  services: ['chromedriver'],
+  services: [
+    [DockerService],
+  ],
   // Framework you want to run your specs with.
   // The following are supported: Mocha, Jasmine, and Cucumber
   // see also: https://webdriver.io/docs/frameworks.html
--- a/packages/terra-functional-testing/lib/config/wdio.conf.js
+++ b/packages/terra-functional-testing/lib/config/wdio.conf.js
@@ -1,4 +1,4 @@
-const DockerService = require('../services/wdio-docker-service');
+const SeleniumDockerService = require('../services/wdio-selenium-docker-service');
 
 exports.config = {
   //
@@ -110,7 +110,7 @@ exports.config = {
   // your test setup with almost no effort. Unlike plugins, they don't add new
   // commands. Instead, they hook themselves up into the test process.
   services: [
-    [DockerService],
+    [SeleniumDockerService],
   ],
   // Framework you want to run your specs with.
   // The following are supported: Mocha, Jasmine, and Cucumber
--- a/packages/terra-functional-testing/lib/config/wdio.conf.js
+++ b/packages/terra-functional-testing/lib/config/wdio.conf.js
@@ -84,6 +84,10 @@ exports.config = {
   bail: 0,
   // Set the path to connect to the selenium container.
   path: '/wd/hub',
+  // The hostname of the driver server.
+  hostname: 'localhost',
+  // The port the driver server is on.
+  port: 4444,
   //
   // Set a base URL in order to shorten url command calls. If your `url` parameter starts
   // with `/`, the base url gets prepended, not including the path portion of your baseUrl.
--- a/packages/terra-functional-testing/lib/docker/docker-compose.yml
+++ b/packages/terra-functional-testing/lib/docker/docker-compose.yml
@@ -2,12 +2,17 @@ version: '3.7'
 
 services:
   hub:
-   image: selenium/hub:3.141.59-zirconium
+   image: selenium/hub:3.141.59-20200525
    ports:
      - "4444:4444"
 
   chrome:
-    image: selenium/node-chrome:3.141.59-zirconium
+    image: selenium/node-chrome:3.141.59-20200525
+    # Volume mounting is necessary to ensure the browser does not crash inside the docker container.
+    # For more details see: 
+    #  https://github.com/SeleniumHQ/docker-selenium#running-the-images
+    #  https://bugs.chromium.org/p/chromium/issues/detail?id=519952
+    #  https://bugzilla.mozilla.org/show_bug.cgi?id=1338771#c10
     volumes:
       - /dev/shm:/dev/shm
     environment:
@@ -18,7 +23,12 @@ services:
     entrypoint: bash -c 'SE_OPTS="-host $$HOSTNAME" /opt/bin/entry_point.sh'
 
   firefox:
-    image: selenium/node-firefox:3.141.59-zirconium
+    image: selenium/node-firefox:3.141.59-20200525
+    # Volume mounting is necessary to ensure the browser does not crash inside the docker container.
+    # For more details see: 
+    #  https://github.com/SeleniumHQ/docker-selenium#running-the-images
+    #  https://bugs.chromium.org/p/chromium/issues/detail?id=519952
+    #  https://bugzilla.mozilla.org/show_bug.cgi?id=1338771#c10
     volumes:
       - /dev/shm:/dev/shm
     environment:
--- a/packages/terra-functional-testing/lib/docker/docker-compose.yml
+++ b/packages/terra-functional-testing/lib/docker/docker-compose.yml
@@ -1,0 +1,31 @@
+version: '3.7'
+
+services:
+  hub:
+   image: selenium/hub:3.141.59-zirconium
+   ports:
+     - "4444:4444"
+
+  chrome:
+    image: selenium/node-chrome:3.141.59-zirconium
+    volumes:
+      - /dev/shm:/dev/shm
+    environment:
+      HUB_HOST: hub
+      HUB_PORT: 4444
+      TZ: 'America/Chicago'
+    deploy:
+        replicas: 1
+    entrypoint: bash -c 'SE_OPTS="-host $$HOSTNAME" /opt/bin/entry_point.sh'
+
+  firefox:
+    image: selenium/node-firefox:3.141.59-zirconium
+    volumes:
+      - /dev/shm:/dev/shm
+    environment:
+      HUB_HOST: hub
+      HUB_PORT: 4444
+      TZ: 'America/Chicago'
+    deploy:
+        replicas: 1
+    entrypoint: bash -c 'SE_OPTS="-host $$HOSTNAME" /opt/bin/entry_point.sh'
--- a/packages/terra-functional-testing/lib/docker/docker-compose.yml
+++ b/packages/terra-functional-testing/lib/docker/docker-compose.yml
@@ -13,7 +13,6 @@ services:
     environment:
       HUB_HOST: hub
       HUB_PORT: 4444
-      TZ: 'America/Chicago'
     deploy:
         replicas: 1
     entrypoint: bash -c 'SE_OPTS="-host $$HOSTNAME" /opt/bin/entry_point.sh'
@@ -25,7 +24,6 @@ services:
     environment:
       HUB_HOST: hub
       HUB_PORT: 4444
-      TZ: 'America/Chicago'
     deploy:
         replicas: 1
     entrypoint: bash -c 'SE_OPTS="-host $$HOSTNAME" /opt/bin/entry_point.sh'
--- a/packages/terra-functional-testing/lib/docker/docker-compose.yml
+++ b/packages/terra-functional-testing/lib/docker/docker-compose.yml
@@ -2,12 +2,12 @@ version: '3.7'
 
 services:
   hub:
-   image: selenium/hub:${IMAGE}
+   image: selenium/hub:${TERRA_SELENIUM_DOCKER_VERSION}
    ports:
      - "4444:4444"
 
   chrome:
-    image: selenium/node-chrome:${IMAGE}
+    image: selenium/node-chrome:${TERRA_SELENIUM_DOCKER_VERSION}
     # Volume mounting is necessary to ensure the browser does not crash inside the docker container.
     # For more details see: 
     #  https://github.com/SeleniumHQ/docker-selenium#running-the-images
@@ -23,7 +23,7 @@ services:
     entrypoint: bash -c 'SE_OPTS="-host $$HOSTNAME" /opt/bin/entry_point.sh'
 
   firefox:
-    image: selenium/node-firefox:${IMAGE}
+    image: selenium/node-firefox:${TERRA_SELENIUM_DOCKER_VERSION}
     # Volume mounting is necessary to ensure the browser does not crash inside the docker container.
     # For more details see: 
     #  https://github.com/SeleniumHQ/docker-selenium#running-the-images
--- a/packages/terra-functional-testing/lib/docker/docker-compose.yml
+++ b/packages/terra-functional-testing/lib/docker/docker-compose.yml
@@ -2,12 +2,12 @@ version: '3.7'
 
 services:
   hub:
-   image: selenium/hub:3.141.59-20200525
+   image: selenium/hub:${IMAGE}
    ports:
      - "4444:4444"
 
   chrome:
-    image: selenium/node-chrome:3.141.59-20200525
+    image: selenium/node-chrome:${IMAGE}
     # Volume mounting is necessary to ensure the browser does not crash inside the docker container.
     # For more details see: 
     #  https://github.com/SeleniumHQ/docker-selenium#running-the-images
@@ -23,7 +23,7 @@ services:
     entrypoint: bash -c 'SE_OPTS="-host $$HOSTNAME" /opt/bin/entry_point.sh'
 
   firefox:
-    image: selenium/node-firefox:3.141.59-20200525
+    image: selenium/node-firefox:${IMAGE}
     # Volume mounting is necessary to ensure the browser does not crash inside the docker container.
     # For more details see: 
     #  https://github.com/SeleniumHQ/docker-selenium#running-the-images
--- a/packages/terra-functional-testing/lib/logger/logger.js
+++ b/packages/terra-functional-testing/lib/logger/logger.js
@@ -1,0 +1,27 @@
+class Logger {
+  constructor(options) {
+    const { prefix } = options;
+
+    this.prefix = prefix;
+  }
+
+  /**
+   * Formats the text.
+   * @param {string} text - The text to format.
+   * @returns {string} - A formatted string.
+   */
+  format(text) {
+    return `[terra-functional-testing:${this.prefix}] ${text}`;
+  }
+
+  /**
+   * Logs the text to the console.
+   * @param {string} text - The text to be logged.
+   */
+  log(text) {
+    // eslint-disable-next-line no-console
+    console.log(this.format(text));
+  }
+}
+
+module.exports = Logger;
--- a/packages/terra-functional-testing/lib/logger/logger.js
+++ b/packages/terra-functional-testing/lib/logger/logger.js
@@ -20,7 +20,16 @@ class Logger {
    */
   log(text) {
     // eslint-disable-next-line no-console
-    console.log(this.format(text));
+    console.log(`[INFO] ${this.format(text)}`);
+  }
+
+  /**
+   * Logs an error message to the console.
+   * @param {string} text - The text to be logged.
+   */
+  error(text) {
+    // eslint-disable-next-line no-console
+    console.log(`[ERROR] ${this.format(text)}`);
   }
 }
 
--- a/packages/terra-functional-testing/lib/services/wdio-docker-service.js
+++ b/packages/terra-functional-testing/lib/services/wdio-docker-service.js
@@ -37,7 +37,7 @@ class DockerService {
   }
 
   /**
-   * Deploys the docker stack.
+   * Deploys the docker stack. The previous stack will be removed if it exists.
    */
   async deployStack() {
     // Remove the previous stack if one exists.
@@ -77,7 +77,7 @@ class DockerService {
    * @param {number} interval - The timeout between commands in milliseconds. Defaults to every two seconds.
    * @returns {Promise} - A promise that resolves when the callback accepts the command response.
    */
-  async pollCommand(command, callback, retries = RETRY_COUNT, interval = POLL_INTERVAL) {
+  async pollCommand(command, callback, retries = 30, interval = 2000) {
     return new Promise((resolve, reject) => {
       let retryCount = 0;
       let pollTimeout = null;
--- a/packages/terra-functional-testing/lib/services/wdio-docker-service.js
+++ b/packages/terra-functional-testing/lib/services/wdio-docker-service.js
@@ -46,7 +46,7 @@ class DockerService {
 
     await exec(`docker stack deploy -c ${composeFilePath} wdio`);
 
-    await this.awaitNetworkReady();
+    await this.waitForNetworkReady();
   }
 
   /**
@@ -61,8 +61,8 @@ class DockerService {
       await exec('docker stack rm wdio');
 
       // Ensure the services and network have been removed.
-      await this.awaitServiceRemoval();
-      await this.awaitNetworkRemoval();
+      await this.waitForServiceRemoval();
+      await this.waitForNetworkRemoval();
     }
   }
 
@@ -103,7 +103,7 @@ class DockerService {
   /**
    * Ensures the docker network has been shut down.
    */
-  async awaitNetworkRemoval() {
+  async waitForNetworkRemoval() {
     await this.pollCommand('docker network ls | grep wdio || true', (result) => (
       new Promise((resolve, reject) => {
         const { stdout: networkStatus } = result;
@@ -120,7 +120,7 @@ class DockerService {
   /**
    * Ensures the docker services have been shut down.
    */
-  async awaitServiceRemoval() {
+  async waitForServiceRemoval() {
     await this.pollCommand('docker service ls | grep wdio || true', (result) => (
       new Promise((resolve, reject) => {
         const { stdout: serviceStatus } = result;
@@ -137,7 +137,7 @@ class DockerService {
   /**
    * Ensures the docker network is ready.
    */
-  async awaitNetworkReady() {
+  async waitForNetworkReady() {
     logger.log('Waiting for docker to become ready.');
 
     await this.pollCommand(`curl -sSL http://${this.host}:${this.port}/wd/hub/status`, (result) => (
--- a/packages/terra-functional-testing/lib/services/wdio-docker-service.js
+++ b/packages/terra-functional-testing/lib/services/wdio-docker-service.js
@@ -1,0 +1,165 @@
+/* eslint-disable class-methods-use-this */
+const path = require('path');
+const util = require('util');
+const childProcess = require('child_process');
+const Logger = require('../logger/logger');
+
+const exec = util.promisify(childProcess.exec);
+const logger = new Logger({ prefix: 'wdio-docker-service' });
+
+const RETRY_COUNT = 30;
+const POLL_INTERVAL = 2000;
+
+class DockerService {
+  /**
+   * Prepares the docker testing environment.
+   */
+  async onPrepare(config) {
+    this.host = config.hostname;
+    this.port = config.port;
+
+    await this.initializeSwarm();
+    await this.deployStack();
+  }
+
+  /**
+   * Initializes a docker swarm instance.
+   */
+  async initializeSwarm() {
+    const { stdout: dockerInfo } = await exec('docker info --format "{{json .}}"');
+    const { Swarm } = JSON.parse(dockerInfo);
+
+    if (Swarm.LocalNodeState !== 'active') {
+      logger.log('Initializing docker swarm.');
+
+      await exec('docker swarm init');
+    }
+  }
+
+  /**
+   * Deploys the docker stack.
+   */
+  async deployStack() {
+    // Remove the previous stack if one exists.
+    await this.removeStack();
+
+    logger.log('Deploying docker stack.');
+
+    const composeFilePath = path.resolve(__dirname, '../docker/docker-compose.yml');
+
+    await exec(`docker stack deploy -c ${composeFilePath} wdio`);
+
+    await this.awaitNetworkReady();
+  }
+
+  /**
+   * Removes the docker stack.
+   */
+  async removeStack() {
+    const { stdout: stackInfo } = await exec('docker stack ls | grep wdio || true');
+
+    if (stackInfo) {
+      logger.log('Removing docker stack.');
+
+      await exec('docker stack rm wdio');
+
+      // Ensure the services and network have been removed.
+      await this.awaitServiceRemoval();
+      await this.awaitNetworkRemoval();
+    }
+  }
+
+  /**
+   * Waits for a command to complete successfully.
+   * @param {string} command - The shell command to run.
+   * @param {func} callback - A callback function to accept or reject the result of the command. Must return a promise.
+   * @returns {Promise} - A promise that resolves when the callback accepts the command response.
+   */
+  async pollCommand(command, callback) {
+    return new Promise((resolve, reject) => {
+      let retryCount = 0;
+      let pollTimeout = null;
+
+      const poll = async () => {
+        if (retryCount >= RETRY_COUNT) {
+          clearTimeout(pollTimeout);
+          pollTimeout = null;
+          reject(Error(logger.format('Timeout. Exceeded retry count.')));
+        }
+
+        try {
+          const result = await exec(command);
+
+          await callback(result).then(() => resolve());
+        } catch (error) {
+          retryCount += 1;
+          pollTimeout = setTimeout(poll, POLL_INTERVAL);
+        }
+      };
+
+      pollTimeout = setTimeout(poll, POLL_INTERVAL);
+    });
+  }
+
+  /**
+   * Ensures the docker network has been shut down.
+   */
+  async awaitNetworkRemoval() {
+    await this.pollCommand('docker network ls | grep wdio || true', (result) => (
+      new Promise((resolve, reject) => {
+        const { stdout: networkStatus } = result;
+
+        // Reject if there is an active network returned.
+        if (networkStatus) {
+          reject();
+        } else {
+          resolve();
+        }
+      })));
+  }
+
+  /**
+   * Ensures the docker services have been shut down.
+   */
+  async awaitServiceRemoval() {
+    await this.pollCommand('docker service ls | grep wdio || true', (result) => (
+      new Promise((resolve, reject) => {
+        const { stdout: serviceStatus } = result;
+
+        // Reject if there is an active service returned.
+        if (serviceStatus) {
+          reject();
+        } else {
+          resolve();
+        }
+      })));
+  }
+
+  /**
+   * Ensures the docker network is ready.
+   */
+  async awaitNetworkReady() {
+    logger.log('Waiting for docker to become ready.');
+
+    await this.pollCommand(`curl -sSL http://${this.host}:${this.port}/wd/hub/status`, (result) => (
+      new Promise((resolve, reject) => {
+        const { stdout } = result;
+        const { value } = JSON.parse(stdout);
+
+        if (value.ready) {
+          resolve();
+        } else {
+          reject();
+        }
+      })));
+  }
+
+  /**
+   * Removes the docker stack and network.
+   */
+  async onComplete() {
+    await this.removeStack();
+  }
+}
+
+module.exports = DockerService;
--- a/packages/terra-functional-testing/lib/services/wdio-docker-service.js
+++ b/packages/terra-functional-testing/lib/services/wdio-docker-service.js
@@ -7,9 +7,6 @@ const Logger = require('../logger/logger');
 const exec = util.promisify(childProcess.exec);
 const logger = new Logger({ prefix: 'wdio-docker-service' });
 
-const RETRY_COUNT = 30;
-const POLL_INTERVAL = 2000;
-
 class DockerService {
   /**
    * Prepares the docker testing environment.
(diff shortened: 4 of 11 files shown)


packages/terra-functional-testing/lib/docker/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+     - "4444:4444"
+
+  chrome:
+    image: selenium/node-chrome:3.141.59-zirconium
+    volumes:
+      - /dev/shm:/dev/shm
+    environment:
+      HUB_HOST: hub
+      HUB_PORT: 4444
+      TZ: 'America/Chicago'
```
emilyrohrbough: We should consider removing this. Internally this timezone used the default and running tests using the docker server compared to the selenium grid has given differing results for time components. 

- https://github.com/cerner/terra-toolkit-boneyard/issues/7

packages/terra-functional-testing/lib/logger/logger.js
```diff
@@ -0,0 +1,27 @@
+class Logger {
```
emilyrohrbough, line 1: Do we intent to create a logger for each package?

packages/terra-functional-testing/lib/logger/logger.js
```diff
@@ -0,0 +1,27 @@
...
+    this.prefix = prefix;
+  }
+
+  /**
+   * Formats the text.
+   * @param {string} text - The text to format.
+   * @returns {string} - A formatted string.
+   */
+  format(text) {
+    return `[terra-functional-testing:${this.prefix}] ${text}`;
```
emilyrohrbough, line 14: colored logs? 😄

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,167 @@
+/* eslint-disable class-methods-use-this */
```
emilyrohrbough: Looks like most of these methods are using this. Can we add to the ones that need to be disabled ?

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * Waits for a command to complete successfully.
+   * @param {string} command - The shell command to run.
+   * @param {func} callback - A callback function to accept or reject the result of the command. Must return a promise.
+   * @param {number} retries - The number of times to retry the command. Defaults to 30 retries.
+   * @param {number} interval - The timeout between commands in milliseconds. Defaults to every two seconds.
+   * @returns {Promise} - A promise that resolves when the callback accepts the command response.
+   */
+  async pollCommand(command, callback, retries = RETRY_COUNT, interval = POLL_INTERVAL) {
```
emilyrohrbough: Why not like the following?
```suggestion
  async pollCommand(command, callback, retries = 30, interval = 2000) {
```

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,167 @@
...
+        } else {
+          resolve();
+        }
+      })));
+  }
+
+  /**
+   * Ensures the docker network is ready.
+   */
+  async awaitNetworkReady() {
```
emilyrohrbough: Something I have been thinking about, and here may not be the correct place to do it, but we should considering polling to validate the deployed selenium grid is ready / available. There's been many times it's been down, test run and consumers test try to run and all will fail

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,167 @@
...
+  /**
+   * Deploys the docker stack.
+   */
+  async deployStack() {
+    // Remove the previous stack if one exists.
+    await this.removeStack();
+
+    logger.log('Deploying docker stack.');
+
+    const composeFilePath = path.resolve(__dirname, '../docker/docker-compose.yml');
```
emilyrohrbough: For windows developers, we will want to use path.join. I know supporting windows development isn't a priority, but I do think it is important for setting up the test harness.
```suggestion
    const composeFilePath = path.resolve(__dirname, path.join('..', 'docker', 'docker-compose.yml'));
```

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,167 @@
...
+      };
+
+      pollTimeout = setTimeout(poll, interval);
+    });
+  }
+
+  /**
+   * Ensures the docker network has been shut down.
+   */
+  async awaitNetworkRemoval() {
```
emilyrohrbough:
```suggestion
  async removeNetwork() {
```

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,167 @@
...
+        } else {
+          resolve();
+        }
+      })));
+  }
+
+  /**
+   * Ensures the docker services have been shut down.
+   */
+  async awaitServiceRemoval() {
```
emilyrohrbough:
```suggestion
  async removeService() {
```

If we want to keep this naming, then we should update `removeStack` to be `awaitStackRemoval`

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,167 @@
...
+
+    if (Swarm.LocalNodeState !== 'active') {
+      logger.log('Initializing docker swarm.');
+
+      await exec('docker swarm init');
+    }
+  }
+
+  /**
+   * Deploys the docker stack.
```
emilyrohrbough: Can we add this removes the previous stack if on exists?

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,167 @@
...
+    await exec(`docker stack deploy -c ${composeFilePath} wdio`);
+
+    await this.awaitNetworkReady();
+  }
+
+  /**
+   * Removes the docker stack.
+   */
+  async removeStack() {
+    const { stdout: stackInfo } = await exec('docker stack ls | grep wdio || true');
```
emilyrohrbough: why `|| true` at the end of this command? If this is returned, will it attempt to remove the wdio stack when it isn't needed?

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,167 @@
...
+
+    if (stackInfo) {
+      logger.log('Removing docker stack.');
+
+      await exec('docker stack rm wdio');
```
emilyrohrbough: If we wait for this command to complete, why do we need `await this.awaitServiceRemoval();`?

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,167 @@
...
+
+    await this.pollCommand(`curl -sSL http://${this.host}:${this.port}/wd/hub/status`, (result) => (
+      new Promise((resolve, reject) => {
+        const { stdout } = result;
+        const { value } = JSON.parse(stdout);
+
+        if (value.ready) {
+          resolve();
+        } else {
+          reject();
```
emilyrohrbough: Can we add a log for when the grid is not ready & it causes the service to fail the onPrepare setup?

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,167 @@
...
+  async removeStack() {
+    const { stdout: stackInfo } = await exec('docker stack ls | grep wdio || true');
+
+    if (stackInfo) {
+      logger.log('Removing docker stack.');
+
+      await exec('docker stack rm wdio');
+
+      // Ensure the services and network have been removed.
+      await this.awaitServiceRemoval();
```
emilyrohrbough: Shouldn't await syntax be wrapped in try-catch blocks?

packages/terra-functional-testing/tests/jest/logger/logger.test.js
```diff
@@ -0,0 +1,24 @@
+const Logger = require('../../../lib/logger/logger');
+
+describe('Logger', () => {
+  describe('format', () => {
+    it('should format test with the initialized prefix', async () => {
```
emilyrohrbough, line 5: Can we mock the console.log for this whole test to prevent log messages during the test run?

Also do these test need an async callback?

emilyrohrbough:
```suggestion
    it('should format test with the initialized prefix', () => {
```

packages/terra-functional-testing/tests/jest/logger/logger.test.js
```diff
@@ -0,0 +1,24 @@
...
+      const logger = new Logger({ prefix: 'mock-prefix' });
+
+      expect(logger.format('Example text')).toEqual('[terra-functional-testing:mock-prefix] Example text');
+    });
+  });
+
+  describe('log', () => {
+    it('should log a message to the console', async () => {
```
emilyrohrbough:
```suggestion
    it('should log a message to the console', () => {
```

packages/terra-functional-testing/tests/jest/logger/logger.test.js
```diff
@@ -0,0 +1,24 @@
...
+      const logger = new Logger({ prefix: 'mock-prefix' });
+
+      jest.spyOn(console, 'log').mockImplementationOnce(() => { });
+
+      logger.log('Example text');
+
+      expect(console.log).toHaveBeenCalledWith('[terra-functional-testing:mock-prefix] Example text');
```
emilyrohrbough:
```suggestion
      // eslint-diasable-next-line no-console
      expect(console.log).toHaveBeenCalledWith('[terra-functional-testing:mock-prefix] Example text');
```

packages/terra-functional-testing/tests/jest/services/wdio-docker-service.test.js
```diff
@@ -0,0 +1,302 @@
...
+const util = require('util');
+const path = require('path');
+
+// The util promisify mock needs to be defined before the docker service import.
+const mockExec = jest.fn();
+util.promisify.mockImplementation(() => mockExec);
+
+const DockerService = require('../../../lib/services/wdio-docker-service');
+
+describe('WDIO Docker Servoce', () => {
```
emilyrohrbough:
```suggestion
describe('WDIO Docker Service', () => {
```

packages/terra-functional-testing/tests/jest/services/wdio-docker-service.test.js
```diff
@@ -0,0 +1,302 @@
...
+
+    it('should not initialize the docker swarm if already active', async () => {
+      const service = new DockerService();
+
+      mockExec.mockImplementation(() => Promise.resolve({ stdout: JSON.stringify({ Swarm: { LocalNodeState: 'active' } }) }));
+
+      await service.initializeSwarm();
+
+      expect(mockExec).toHaveBeenCalledWith('docker info --format "{{json .}}"');
+      expect(mockExec).toHaveBeenCalledTimes(1);
```
emilyrohrbough: expect(mockExec).not.toHaveBeenCalledWith('docker swarm init');

packages/terra-functional-testing/tests/jest/services/wdio-docker-service.test.js
```diff
@@ -0,0 +1,302 @@
...
+  describe('removeStack', () => {
+    it('should not execute the docker stack removal if it is already removed', async () => {
+      const service = new DockerService();
+
+      mockExec.mockImplementation(() => Promise.resolve({ stdout: '' }));
+
+      await service.removeStack();
+
+      expect(mockExec).toHaveBeenCalledTimes(1);
+      expect(mockExec).toHaveBeenCalledWith('docker stack ls | grep wdio || true');
```
emilyrohrbough:
```suggestion
      expect(mockExec).toHaveBeenCalledWith('docker stack ls | grep wdio || true');
       expect(mockExec).not.toHaveBeenCalledWith('docker stack rm wdio');
```

packages/terra-functional-testing/tests/jest/services/wdio-docker-service.test.js
```diff
@@ -0,0 +1,302 @@
...
+    it('should timeout if the retry count is exceeded', async () => {
+      const service = new DockerService();
+
+      mockExec.mockImplementation(() => Promise.resolve());
+
+      const mockCallback = jest.fn().mockImplementation(() => Promise.reject());
+
+      try {
+        // Retry 3 times. Repeat every 10 milliseconds.
+        await service.pollCommand('mock command', mockCallback, 3, 10);
```
emilyrohrbough: is the the only place we expect the retry count & poll time to be used? I didn't notice it being used in the class

packages/terra-functional-testing/tests/jest/services/wdio-docker-service.test.js
```diff
@@ -0,0 +1,302 @@
...
+      const service = new DockerService();
+
+      jest.spyOn(service, 'pollCommand').mockImplementationOnce(async (_command, callback) => {
+        await expect(callback({ stdout: 'Active Network' })).rejects.toBeUndefined();
+
+        return Promise.resolve();
+      });
+
+      await service.awaitNetworkRemoval();
+
```
emilyrohrbough: Is this is rejected, does the service onPrepare fail? Is a message logged or a stack trace provided to the user?

packages/terra-functional-testing/lib/logger/logger.js
StephenEsser: I brought this question up in scrum last week, we don't currently have a re-usable Logger that can be used between packages. But each Logger we have is only a few lines of code. If we want logging in this project, at least for now, we need to create one.  A re-usable Logger would be nice for consistency between projects that utilize logging. I think it would be a nice-to-have but not a make-or-break between projects.

This would be a good war room topic or post-scrum discussion if we have time

packages/terra-functional-testing/lib/logger/logger.js
StephenEsser: Starting small with this logger. When I run into a situation that needs to log different tiers of output I'll likely add in colored logs. The [WebDriver logger](https://github.com/webdriverio/webdriverio/tree/master/packages/wdio-logger) has differences between their INFO, DEBUG, ERROR logs. Similar to our existing logger. For this PR I don't intend to add colored logs. I think it would be good to have when we discuss this in more detail:

https://github.com/cerner/terra-toolkit/pull/412#discussion_r433853473

packages/terra-functional-testing/lib/config/wdio.conf.js
```diff
@@ -80,6 +82,8 @@ exports.config = {
   // If you only want to run your tests until a specific amount of tests have failed use
   // bail (default is 0 - don't bail, run all tests).
   bail: 0,
```
emilyrohrbough, line 84: Did we need add the hostname & port to the config here? I know these are defaulted, but would be nice to be face up
```
hostname: `http://localhost`,
port: 4444,
```

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: We can discuss this a little more, normally I'd agree with this change, but because of how the hooks work in these services they need to be class functions regardless of the presence of `this`. Ideally any method that doesn't use `this` would be static, but the service hooks need to be instances on the class. My preference right now is to leave this disabled for the file, specifically for services.

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: These functions don't run the command to remove the service, they're waiting/polling the service/network to verify it has been removed. Unfortunately, I don't think `removeService` is appropriate here. Do you have naming suggestions that align more closely to the behavior of the function? I'm not a huge fan of the await prefix. waitFor seemed a little verbose, but is pretty standard for this functionality.

The `docker stack rm wdio` command run in `removeStack` is used to begin the process of removing the network. These functions poll until the network is verified to be inactive.

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: I'd like to get a few more opinions before making this change. I'm hesitant about developing packages that `might` work on windows. If this is all that is required for this package to fully support windows I think it would be a good addition, but if there are other issues with Windows compatibility I think we should resolve them or document our limited OS support.

Do you have any thoughts opinions @ryanthemanuel @mjhenkes  ?

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: The bash command  (`docker stack ls | grep wdio`) returns a non-zero exit code if the stack has successfully been removed because there is no output that matches the grep request. This allows the command to succeed if the stack has been removed. The alternative to this is a try catch block and analyzing the error in the catch block. A little awkward because a catch is a succeed in this case

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: `exec` returns a promise. The command is run asynchronously. The promise will resolve when the command has been executed, but the docker network isn't guaranteed to be taken down yet. So this waits for the command to run and then we need to wait for the network/service to be taken down

packages/terra-functional-testing/tests/jest/services/wdio-docker-service.test.js
StephenEsser: Yes, unfortunately I ended up adding it to simplify testing. Without these overrides the test to reach the timeout will take 60 seconds. The poll function is intended to be re-usable, so the additional parameters don't seem harmful and could be beneficial/utilized with configuration options in the future.

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: Originally these weren't used in the parameters and I wanted to avoid magic numbers. I've removed the constants and updated these to match your parameter suggestions:

https://github.com/cerner/terra-toolkit/pull/412/commits/81ec6b8b522dd42261437c12f789f6e373615c9a

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: Added a comment that the previous stack will be removed if it exists:

https://github.com/cerner/terra-toolkit/pull/412/commits/81ec6b8b522dd42261437c12f789f6e373615c9a

packages/terra-functional-testing/tests/jest/logger/logger.test.js
StephenEsser: Removed async:  https://github.com/cerner/terra-toolkit/pull/412/commits/81ec6b8b522dd42261437c12f789f6e373615c9a

packages/terra-functional-testing/tests/jest/logger/logger.test.js
StephenEsser: Removed async: https://github.com/cerner/terra-toolkit/pull/412/commits/81ec6b8b522dd42261437c12f789f6e373615c9a

packages/terra-functional-testing/tests/jest/logger/logger.test.js
StephenEsser: Added the console log eslint disable:

https://github.com/cerner/terra-toolkit/pull/412/commits/81ec6b8b522dd42261437c12f789f6e373615c9a

packages/terra-functional-testing/tests/jest/services/wdio-docker-service.test.js
StephenEsser: Corrected typo: https://github.com/cerner/terra-toolkit/pull/412/commits/81ec6b8b522dd42261437c12f789f6e373615c9a

packages/terra-functional-testing/tests/jest/services/wdio-docker-service.test.js
StephenEsser: Added the additional check. I kept the times(1) check as a precaution. The not have been called check would return a false positive if the command was updated or typoed in the test. 

https://github.com/cerner/terra-toolkit/pull/412/commits/81ec6b8b522dd42261437c12f789f6e373615c9a

packages/terra-functional-testing/tests/jest/services/wdio-docker-service.test.js
StephenEsser: Added the additional check:

https://github.com/cerner/terra-toolkit/pull/412/commits/81ec6b8b522dd42261437c12f789f6e373615c9a

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: More discussions on this thread:

https://github.com/cerner/terra-toolkit/pull/412#discussion_r433860971

packages/terra-functional-testing/lib/services/wdio-docker-service.js
emilyrohrbough: Isn't `initializeSwarm` the only method that isn't using this?

packages/terra-functional-testing/lib/logger/logger.js
emilyrohrbough: Should we add this to the list of v6 list, whether we do it or not, so we don't forget?

packages/terra-functional-testing/lib/services/wdio-docker-service.js
emilyrohrbough: `ensureServiceRemoved`

packages/terra-functional-testing/lib/services/wdio-docker-service.js
emilyrohrbough: When the service is remove, won't it always be removed, even if the network is still up? I do `this.awaitNetworkRemoval` is needed.

emilyrohrbough:
Is there docs for this service ?

StephenEsser:
> Is there docs for this service ?

I'm working on some docs locally, I'll get those pushed out here soon

packages/terra-functional-testing/lib/services/wdio-docker-service.js
mjhenkes: we don't support windows development

packages/terra-functional-testing/lib/docker/docker-compose.yml
StephenEsser: I'll remove this for now and add a note on the epic issue to discuss this further prior to an official release.

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: In my opinion, `ensure` sounds like a conditional check. I think these functions should indicate they are waiting for an action to complete.  `wait`, `waitFor`, `await`, `poll`. Anyone have an opinion/vote between all the options or a nomination for another option?

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: If the network is still up when we try to deploy another stack it will cause errors. We have to wait for the network to go down before we create a new one. This happens when the test run is hard-stopped or errors and the onComplete isn't run to take down the stack before launching a new stack. The remove stack will be executed on deploy stack if an existing network is active.

Here's a thread with a similar error I was encountering with out this wait:

https://github.com/moby/moby/issues/31226#issuecomment-281718968

packages/terra-functional-testing/tests/jest/services/wdio-docker-service.test.js
StephenEsser: If this is rejected or exceeds the retry count the timeout message will bubble and be logged to the console:

```
2020-06-02T21:22:49.364Z INFO @wdio/cli:launcher: Run onPrepare hook
[terra-functional-testing:wdio-docker-service] Deploying docker stack.
[terra-functional-testing:wdio-docker-service] Waiting for docker to become ready.
2020-06-02T21:22:58.681Z ERROR @wdio/cli:utils: A service failed in the 'onPrepare' hook
Error: [terra-functional-testing:wdio-docker-service] Timeout. Exceeded retry count.
    at Timeout.poll (/Users/se028266/workspaces/terra/terra-toolkit/packages/terra-functional-testing/lib/services/wdio-docker-service.js:86:18)
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at listOnTimeout (timers.js:263:5)
    at Timer.processTimers (timers.js:223:10)

Continue...
```

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: If this is rejected or exceeds the retry count the timeout message will bubble and be logged to the console:

```
2020-06-02T21:22:49.364Z INFO @wdio/cli:launcher: Run onPrepare hook
[terra-functional-testing:wdio-docker-service] Deploying docker stack.
[terra-functional-testing:wdio-docker-service] Waiting for docker to become ready.
2020-06-02T21:22:58.681Z ERROR @wdio/cli:utils: A service failed in the 'onPrepare' hook
Error: [terra-functional-testing:wdio-docker-service] Timeout. Exceeded retry count.
    at Timeout.poll (/Users/se028266/workspaces/terra/terra-toolkit/packages/terra-functional-testing/lib/services/wdio-docker-service.js:86:18)
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at listOnTimeout (timers.js:263:5)
    at Timer.processTimers (timers.js:223:10)

Continue...
```

We can catch the error and bubble a more descriptive one if we'd like. The timeout message is generic since it's used for different things

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: The try catch is only necessary when you want to explicitly catch errors. If a rejection occurs here it will reject the async promise and bubble up to the WDIO catch and log the message for the hook that failed

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: `initializeSwam` and `pollCommand` are the two functions not using `this`. I refactored the other functions to use smaller utility functions that worked out to satisfy the `this` lint rule. If we want to remove this eslint-disable I'll likely convert `initializeSwam` and `pollCommand` to static functions rather than ignore them individually. Only the hook methods are problematic with this rule, since they are required to be class methods regardless of `this` being used

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: I think that's a good idea, it might make sense to be another service with a few responsibilities to validate the testing environment on the remote grid

packages/terra-functional-testing/lib/docker/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
+version: '3.7'
+
+services:
+  hub:
+   image: selenium/hub:3.141.59-zirconium
+   ports:
+     - "4444:4444"
+
+  chrome:
+    image: selenium/node-chrome:3.141.59-zirconium
```
mjhenkes: do we want the latest?
```suggestion
    image: selenium/node-chrome:3.141.59-20200525
```
https://github.com/SeleniumHQ/docker-selenium/releases/tag/3.141.59-20200525

packages/terra-functional-testing/lib/docker/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    volumes:
+      - /dev/shm:/dev/shm
```
mjhenkes, line 17: Could you add comments about why we're volume mounting?

packages/terra-functional-testing/lib/docker/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
+version: '3.7'
+
+services:
+  hub:
+   image: selenium/hub:3.141.59-zirconium
+   ports:
+     - "4444:4444"
+
+  chrome:
```
mjhenkes, line 9: do chrome and firefox depend on hub?

https://docs.docker.com/compose/compose-file/#depends_on

packages/terra-functional-testing/lib/docker/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+  firefox:
+    image: selenium/node-firefox:3.141.59-zirconium
+    volumes:
+      - /dev/shm:/dev/shm
+    environment:
+      HUB_HOST: hub
+      HUB_PORT: 4444
+    deploy:
+        replicas: 1
+    entrypoint: bash -c 'SE_OPTS="-host $$HOSTNAME" /opt/bin/entry_point.sh'
```
mjhenkes, line 39: If you want to be cheeky,  you could use anchors to reduce your docker compose file.
```yml
  x-selenium-browser: &selenium-browser
    volumes:
      - /dev/shm:/dev/shm
    environment:
      HUB_HOST: hub
      HUB_PORT: 4444
    deploy:
        replicas: 1
    entrypoint: bash -c 'SE_OPTS="-host $$HOSTNAME" /opt/bin/entry_point.sh'

  chrome:
    << *selenium-browser
    image: selenium/node-chrome:3.141.59-zirconium

  firefox:
    << *selenium-browser
    image: selenium/node-firefox:3.141.59-zirconium
```
https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd

packages/terra-functional-testing/lib/docker/docker-compose.yml
StephenEsser: I'll update the image, I had thought the date images were beta releases. It looks like they switched and adopted the date suffix:

https://github.com/SeleniumHQ/docker-selenium/wiki/Tagging-Convention

Updated images: https://github.com/cerner/terra-toolkit/pull/412/commits/60f5cc78d2cbdf3782668304a3fe557dc03e2caf

packages/terra-functional-testing/lib/docker/docker-compose.yml
StephenEsser: Unfortunately depends_on is ignored when deploying docker in swarm mode:

> The depends_on option is ignored when deploying a stack in swarm mode with a version 3 Compose file.

packages/terra-functional-testing/lib/docker/docker-compose.yml
StephenEsser: I'll add some comments, this file is nearly a direct copy paste from the selenium-docker recommendation:

https://github.com/SeleniumHQ/docker-selenium#version-3-with-swarm-support

Added some documentation and references for more details:

https://github.com/cerner/terra-toolkit/pull/412/commits/60f5cc78d2cbdf3782668304a3fe557dc03e2caf

packages/terra-functional-testing/lib/docker/docker-compose.yml
StephenEsser: This is a cool. I'd like to stay closely with the recommended compose file if we can. Being explicit feels easier to follow:

https://github.com/SeleniumHQ/docker-selenium#version-3-with-swarm-support

packages/terra-functional-testing/lib/logger/logger.js
StephenEsser: I created a new issue to track the improvements to the logger. Feel free to add/edit the issue with more information/suggetions/requests:

https://github.com/cerner/terra-toolkit/issues/414

packages/terra-functional-testing/lib/services/wdio-docker-service.js
emilyrohrbough: spoke offline. Will keep as is.

packages/terra-functional-testing/lib/services/wdio-docker-service.js
emilyrohrbough: ah the hard stopping. Makes sense 👍

packages/terra-functional-testing/lib/config/wdio.conf.js
StephenEsser: Added the hostname and port:

https://github.com/cerner/terra-toolkit/pull/412/commits/5081644beac4fd9e527541810ddcd4c9c934531e

Note: hostname was added as `localhost` without the http protocol. This doesn't appear to cause any issues. I took the defaults as listed on the WebDriverIO doc site:

https://webdriver.io/docs/options.html

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: Updated the method names to be `waitFor`:

https://github.com/cerner/terra-toolkit/pull/412/commits/5081644beac4fd9e527541810ddcd4c9c934531e

emilyrohrbough (approved):

packages/terra-functional-testing/lib/services/wdio-docker-service.js
mjhenkes: IMO, this is fine to disable at the top of the file. It'll cover any new methods we add in the future as well or refactoring of functions that may no longer use this.

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,164 @@
+/* eslint-disable class-methods-use-this */
+const path = require('path');
+const util = require('util');
+const childProcess = require('child_process');
+const Logger = require('../logger/logger');
+
+const exec = util.promisify(childProcess.exec);
+const logger = new Logger({ prefix: 'wdio-docker-service' });
+
+class DockerService {
```
mjhenkes: Would this be better named, `LocalSeleniumService`? That seems to matter more than docker in this instance.

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,164 @@
...
+  /**
+   * Prepares the docker testing environment.
+   */
+  async onPrepare(config) {
```
mjhenkes: Should we check if docker is installed? And fail if it isn't?, just thinking of questions we've had in the past.

packages/terra-functional-testing/lib/services/wdio-docker-service.js
```diff
@@ -0,0 +1,164 @@
...
+  /**
+   * Deploys the docker stack. The previous stack will be removed if it exists.
+   */
+  async deployStack() {
+    // Remove the previous stack if one exists.
+    await this.removeStack();
+
+    logger.log('Deploying docker stack.');
+
+    const composeFilePath = path.resolve(__dirname, '../docker/docker-compose.yml');
```
mjhenkes: What if we gave people a beta channel docker compose file so we could provide the next version of selenium/browsers and they can try it out/switch to it before we force them to upgrade?

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: How do you feel about meeting in the middle?

`SeleniumDockerService`

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: I'll look into adding this check. We might end up in a weird situation, but at least we'll be able to log something to the console. We can't fail the test run in the services, so the service will fail and then the tests will attempt to run. 

https://github.com/webdriverio/webdriverio/issues/5398

packages/terra-functional-testing/lib/services/wdio-docker-service.js
mjhenkes: I'll take it. `SeleniumDockerService` it is.

ryanthemanuel (approved):

packages/terra-functional-testing/lib/services/wdio-selenium-docker-service.js
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * Waits for a command to complete successfully.
+   * @param {string} command - The shell command to run.
+   * @param {func} callback - A callback function to accept or reject the result of the command. Must return a promise.
+   * @param {number} retries - The number of times to retry the command. Defaults to 30 retries.
+   * @param {number} interval - The timeout between commands in milliseconds. Defaults to every two seconds.
+   * @returns {Promise} - A promise that resolves when the callback accepts the command response.
+   */
+  async pollCommand(command, callback, retries = 30, interval = 2000) {
```
yuderekyu, line 91: curious, how did we land on 30 retries?

StephenEsser: It's an arbitrary value intended to give these command attempts 60 seconds to start/stop. 

Interval - Every two seconds
Duration - Timeout after 1 minute

2000ms * 30 = 60000ms = 1 minute

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: Renamed to SeleniumDockerService:

https://github.com/cerner/terra-toolkit/pull/412/commits/98af84b2dcc2b282fe4b3bf350ac61487431ac9d

packages/terra-functional-testing/lib/services/wdio-docker-service.js
StephenEsser: Added a check that will fail if docker is not installed:

https://github.com/cerner/terra-toolkit/pull/412/commits/7d1d57f6fcc0a7f590c79a38a6222021d70786e6

mjhenkes (approved):

ryanthemanuel (approved):

emilyrohrbough (approved):

Issue #413, opened 2020-06-04
[terra-functional-testing] Create a Terra Service for setting up the test environment


StephenEsser:
# Feature Request

## Description

Create a Terra Service for setting up the testing environment. The service should carry forward the customizations and features of the existing [terra service](https://github.com/cerner/terra-toolkit-boneyard/blob/master/src/wdio/services/TerraService.js).

StephenEsser:
Closing as work has been completed for the terra-service:

- https://github.com/cerner/terra-toolkit/pull/452
- https://github.com/cerner/terra-toolkit/pull/416
- https://github.com/cerner/terra-toolkit/pull/438
- https://github.com/cerner/terra-toolkit/pull/464
- https://github.com/cerner/terra-toolkit/pull/545

PR #416, opened 2020-06-17
[terra-functional-testing] Add TerraService and axe-core integration

StephenEsser:
### Summary

This PR sets up the initial Terra Service and integrates axe-core into the testing environment.

This is part of https://github.com/cerner/terra-toolkit/issues/413 which is being expanded into multiple smaller PRs. 

### Additional Details

Axe is added as a global command to the browser object and is dynamically/conditionally injected onto the page.

For comparison the existing implementation can be found here:

Inject/running axe:

https://github.com/cerner/terra-toolkit-boneyard/blob/master/src/wdio/services/TerraCommands/axe-command.js

Tera Service:

https://github.com/cerner/terra-toolkit-boneyard/blob/master/src/wdio/services/TerraService.js

Differences:

The option to opt out of injecting axe has been removed. I don't know of anyone doing this today and I don't think consumers should have that level of granularity. The flag is also redundant because the page will be checked for axe prior to injecting it. So if they did add axe themselves we would not inject it.

https://github.com/cerner/terra-toolkit-boneyard/blob/master/src/wdio/services/TerraCommands/axe-command.js#L73

The `axe` option is expected to be included with the service options. The `option` key has been removed and flattened. Instead of `{ axe: { options: { rules } } }` rules are specified as `{ axe: { rules } }`:

https://github.com/cerner/terra-toolkit-boneyard/blob/master/config/wdio/wdio.conf.js#L97

Tests will be added prior to merging.

@cerner/terra


packages/terra-functional-testing/lib/commands/run-axe.js
```diff
@@ -0,0 +1,37 @@
...
+
+  // Inject axe-core onto the page if it has not already been initialize.
+  if (isAxeUnavailable) {
+    injectAxe();
+  }
+
+  // Extract the axe options for the Terra service from the global browser object.
+  const [, options = {}] = browser.options.services.find(([service]) => (
+    typeof service === 'function' && service.name === 'TerraService'
+  ));
```
emilyrohrbough, line 13: This logic is duplicated in inject-axe. Could we move this up before the `isAxeUnavailable` check & pass these options down? We could also deconstruction the axe config from the service options and simplify the rules merge as well
```suggestion
  const { axe } = options;
```

packages/terra-functional-testing/lib/config/wdio.conf.js
```diff
@@ -63,7 +64,7 @@ exports.config = {
   // Define all options that are relevant for the WebdriverIO instance here
   //
   // Level of logging verbosity: trace | debug | info | warn | error | silent
-  logLevel: 'info',
+  logLevel: 'silent',
```
emilyrohrbough, line 69: Why change to silent? Just curious. I have no objections to the change

packages/terra-functional-testing/package.json
```diff
@@ -37,6 +37,8 @@
     "@wdio/local-runner": "^6.1.4",
     "@wdio/mocha-framework": "^6.1.0",
     "@wdio/spec-reporter": "^6.0.16",
+    "@wdio/sync": "^6.1.14",
```
emilyrohrbough, line 40: did this dep needed to be added?

packages/terra-functional-testing/lib/terra-dev-site/doc/terra-service/TerraService.doc.mdx
```diff
@@ -0,0 +1,64 @@
+# Terra Service
+
+The terra service integrates [axe accessibility](https://github.com/dequelabs/axe-core) testing into the test environment.
+
+## Installation
+
+The terra service should be installed as a development dependency.
```
emilyrohrbough:
```suggestion
The Terra Service should be installed as a development dependency.
```

packages/terra-functional-testing/lib/terra-dev-site/doc/terra-service/TerraService.doc.mdx
```diff
@@ -0,0 +1,64 @@
+# Terra Service
+
+The terra service integrates [axe accessibility](https://github.com/dequelabs/axe-core) testing into the test environment.
```
emilyrohrbough: Can we add info that this is a static analysis of the current dom structure & we are checking against these standards: 'wcag2a', 'wcag2aa', 'wcag21aa', 'section508'?

packages/terra-functional-testing/lib/config/wdio.conf.js
StephenEsser: The default log level of `info` is kind enough to print out the entire command being run during `browser.execute`. In this case, ~20,000 lines of minified axe-code. The existing configuration is set to silent, I suspect for similar reasons:

https://github.com/cerner/terra-toolkit-boneyard/blob/master/config/wdio/wdio.conf.js#L72

packages/terra-functional-testing/lib/commands/run-axe.js
```diff
@@ -0,0 +1,37 @@
...
+
+  // Merge the global rules and overrides together.
+  const rules = {
+    ...options.axe && options.axe.rules && options.axe.rules.reduce((acc, rule) => ({ ...acc, [rule.id]: rule }), {}),
+    ...overrides.rules,
+  };
+
+  // eslint-disable-next-line prefer-arrow-callback, func-names
+  return browser.executeAsync(function (opts, done) {
```
emilyrohrbough, line 30: Should we be returning the value of the result?

packages/terra-functional-testing/package.json
StephenEsser: Yes, the existence of the package is the flag for running wdio synchronously.

https://webdriver.io/docs/sync-vs-async.html

It's the equivalent to this:

https://github.com/cerner/terra-toolkit-boneyard/blob/master/config/wdio/wdio.conf.js#L71

packages/terra-functional-testing/lib/config/wdio.conf.js
emilyrohrbough: oh fun.... did warn or error log these as well?

packages/terra-functional-testing/lib/commands/run-axe.js
StephenEsser: No, this must have been one of the non-passive changes from WDIO 4 to WDIO 6. Here's a compressed output of this return statement:

```
[0-0] { error: null,
  result:
   { inapplicable:
      [ [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object] ],
     incomplete: [ [Object] ],
     passes:
      [ [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object],
        [Object] ],
     testEngine: { name: 'axe-core', version: '3.5.4' },
     testEnvironment:
      { orientationAngle: 0,
        orientationType: 'landscape-primary',
        userAgent:
         'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.61 Safari/537.36',
        windowHeight: 884,
        windowWidth: 1050 },
     testRunner: { name: 'axe' },
     timestamp: '2020-06-17T15:22:26.237Z',
     toolOptions:
      { reporter: 'v1',
        restoreScroll: true,
        rules: {},
        runOnly: [Object] },
     url:
      'https://engineering.cerner.com/terra-ui/home/terra-ui/index',
     violations: [] } }
````

File: packages/terra-functional-testing/lib/commands/run-axe.js (hunk 1 above)
Comment (StephenEsser) [reply packages/terra-functional-testing/lib/commands/run-axe.js:13]: Consolidated the logic, these were split out intentionally before, but I don't think injectAxe will/should ever be called from anywhere else except this function:

https://github.com/cerner/terra-toolkit/pull/416/commits/2e4d70c646089a29c41346d30db7336ab1a5f41a

File: packages/terra-functional-testing/lib/terra-dev-site/doc/terra-service/TerraService.doc.mdx (hunk 4 above)
Comment (StephenEsser) [reply packages/terra-functional-testing/lib/terra-dev-site/doc/terra-service/TerraService.doc.mdx]: I met halfway and uppercased all occurrences of `Terra`. I left `service` lowercase because of how it's being used in a sentence. It's uppercase when used as a title or single noun. `TerraService`. My experience with grammar is lacking so this could be wrong.

https://github.com/cerner/terra-toolkit/pull/416/commits/2e4d70c646089a29c41346d30db7336ab1a5f41a

File: packages/terra-functional-testing/lib/terra-dev-site/doc/terra-service/TerraService.doc.mdx (hunk 5 above)
Comment (StephenEsser) [reply packages/terra-functional-testing/lib/terra-dev-site/doc/terra-service/TerraService.doc.mdx]: Added more documentation to cover the standards we test against. I'm somewhat intentionally vague about how the command is run for now. I don't think anyone should be calling `browser.axe` directly, but it will eventually be wrapped by our chai assertions and the documentation for those will be reflect how to use it when they are available.

https://github.com/cerner/terra-toolkit/pull/416/commits/2e4d70c646089a29c41346d30db7336ab1a5f41a

File: packages/terra-functional-testing/package.json [hunk 7]
```diff
@@ -37,6 +37,8 @@
...
+    "axe-core": "3.5.4",
```
emilyrohrbough: There was a release this morning: https://github.com/dequelabs/axe-core/releases/tag/v3.5.5

StephenEsser: Impressive timing. Upgraded to latest

https://github.com/cerner/terra-toolkit/pull/416/commits/c08e020d50c63b7932fe6e0e2db42a3ec2e9efdf

package.json
```diff
@@ -30,6 +30,7 @@
     "clean": "lerna clean --yes && rm -rf node_modules",
     "clean:install": "npm run clean && npm install",
     "jest": "jest",
+    "wdio": "node ./packages/terra-functional-testing/bin/test-runner.js",
```
StephenEsser, line 33: I wasn't sure how to hoist the `tt-test` command from terra-functional-testing to the root of the lerna repository. Its the reverse of the exposing the parent bin commands to the packages. If anyone knows how to get this working let me know and I'll update it.

packages/terra-functional-testing/tests/wdio/run-axe-spec.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * This spec is used for testing the global axe command.
+ */
+
+describe('Axe', () => {
+  it('should successfully run axe on the page', () => {
+    browser.url('https://engineering.cerner.com/terra-ui/');
+
+    const { result } = browser.axe();
```
emilyrohrbough, line 9: Can we return result object directly instead of returning an object with the result object?

emilyrohrbough:
👀 😄 
Screen Shot 2020-06-18 at 8 39 52 AM

packages/terra-functional-testing/tests/wdio/run-axe-spec.js
StephenEsser: I think we'll want to keep this as is unless we shift error handling into the the `run-axe` command. I was thinking about adding the error handling into the calling commands (the chai assertions/validates) and keeping the axe run command as a pass through:

https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#context-parameter-examples

I'm not really sure under what circumstances this will throw an error, but we'll want to catch it and log it/fail the assertion somewhere in the stack. I don't think consumers should be running this command directly, I think they should be running the assertions/validation commands

emilyrohrbough: hmm yeah It'd be some things to think about. I've been thinking about the current axe setup and it would be nice if we could wrap these in try-catch to try to prevent the ugly axe errors we sometimes see. I'm not sure if we could catch those or not depending on where those are being throw.

package.json
ryanthemanuel: This is what I had to do as well for some of the stuff I've been playing around with.  No problems on my end with it.

packages/terra-functional-testing/lib/commands/run-axe.js
```diff
@@ -0,0 +1,38 @@
...
+    // eslint-disable-next-line prefer-arrow-callback, func-names
+    axe.run(document, opts, function (error, result) {
+      done({ error, result });
+    });
+  }, { rules, restoreScroll: true, runOnly: ['wcag2a', 'wcag2aa', 'wcag21aa', 'section508'] });
```
yuderekyu, line 35: do we need to add `wcag21a` to our standards? i'm actually confused on how these keys translate to what standard we run against. if `wcag2aa` contains WCAG 2.0 and WCAG 2.1 Level AA, do we need to include `wcag21aa` then? or is that for explicit purposes

StephenEsser: The way these standards work is each tag has a set of rules that apply to that tag. So each tag here concats a new list of rules to be run. The rules don't need to be unique, so they can be in multiple standards but duplicates get filtered out. To add a new standard will add a new set of rules. 

Here's a little more info that helped me understand how this works:

https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#parameters

https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md

As far as adding a new standard I think that's worth a discussion. Adding a standard is non-passive because it adds a new set of rules, but we can be non-passive here if we want to.

packages/terra-functional-testing/lib/config/wdio.conf.js
```diff
@@ -1,6 +1,8 @@
 const SeleniumDockerService = require('../services/wdio-selenium-docker-service');
 const TerraService = require('../services/wdio-terra-service');
 
+const { ci } = process.env;
```
emilyrohrbough: I think we will want this to be all caps

StephenEsser: Whoops, updated. This should be picking up the environment variable in the docker image. Might take a little trial and error on this PR.

https://github.com/cerner/terra-toolkit/pull/416/commits/736ac70c422994dc652ef8de7a52ecb2781c56fd

packages/terra-functional-testing/lib/commands/run-axe.js
yuderekyu: ahh gotcha. the docs are confusing - the website lists the `wcag2a` and `wcag2aa` tags as also containing the WCAG2.1 standards


Screen Shot 2020-06-19 at 11 28 08 AM

emilyrohrbough (approved):

packages/terra-functional-testing/lib/commands/run-axe.js
```diff
@@ -0,0 +1,38 @@
+/* global browser, axe */
+const injectAxe = require('./inject-axe');
+
+/**
+ * Executes axe on the browser.
+ * @param {Object} overrides - The axe options.
+ * @param {Array} overrides.rules - The rule overrides
```
ryanthemanuel:
```suggestion
 * @param {Array} overrides.rules - The rule overrides.
```

packages/terra-functional-testing/lib/commands/run-axe.js
```diff
@@ -0,0 +1,38 @@
...
+const runAxe = (overrides = {}) => {
+  // Extract the axe options for the Terra service from the global browser object.
+  const [, options = {}] = browser.options.services.find(([service]) => (
+    typeof service === 'function' && service.name === 'TerraService'
+  ));
+
+  const { axe: axeOptions } = options;
+  const isAxeUnavailable = browser.execute(() => window.axe === undefined);
+
+  // Inject axe-core onto the page if it has not already been initialize.
```
ryanthemanuel:
```suggestion
  // Inject axe-core onto the page if it has not already been initialized.
```

packages/terra-functional-testing/lib/commands/run-axe.js
StephenEsser: Updated:

https://github.com/cerner/terra-toolkit/pull/416/commits/889f92b2e4122d0f78c6276fd09c2a27671f469c

packages/terra-functional-testing/lib/commands/run-axe.js
StephenEsser: Updated:

https://github.com/cerner/terra-toolkit/pull/416/commits/889f92b2e4122d0f78c6276fd09c2a27671f469c

ryanthemanuel (approved):

PR #417, opened 2020-07-02
[terra-functional-testing] Add WebDriverIO Asset Service

StephenEsser:
### Summary

This PR adds the asset service (name subject to change). The asset service enables serving assets to the webdriver testing environment. 

When a `site` is provided the asset service will initialize and start an express server to serve the assets. 

When a `config` is provided the asset service will initialize and start a webpack dev server that webpacks the assets using the provided configuration. The `config` is a file path.

### Additional Details

Open to suggestions on naming the service. Originally it was named `serve-static-service`. I briefly changed it to `webpack-service`, however switched to `asset-service` because an express service might be started instead of a webpack service. 

Offline discussions have landed (for now) on wdio-asset-server-service

### Resources/Reference

Existing express/serve static implementation: https://github.com/cerner/terra-toolkit-boneyard/blob/master/scripts/serve/serve-static.js

Existing serve-static-service implementation: https://github.com/cerner/terra-toolkit-boneyard/blob/master/src/wdio/services/ServeStaticService.js

@cerner/terra


packages/terra-functional-testing/lib/webpack-server/webpack-server.js
```diff
@@ -0,0 +1,139 @@
+const WebpackDevServer = require('webpack-dev-server');
```
emilyrohrbough, line 1: could we name this file index.js ?

package.json
```diff
@@ -30,15 +30,19 @@
     "clean": "lerna clean --yes && rm -rf node_modules",
     "clean:install": "npm run clean && npm install",
     "jest": "jest",
-    "wdio": "node ./packages/terra-functional-testing/bin/test-runner.js",
     "lint": "npm run lint:js",
     "lint:js": "eslint --ext .js,.jsx .",
+    "pack": "webpack --config webpack.config.js -p",
```
emilyrohrbough: webpack will pick up this file by default so you wouldn't have to pass in `--config webpack.config.js`

packages/terra-functional-testing/lib/config/wdio.conf.js
```diff
@@ -113,9 +125,15 @@ exports.config = {
   // your test setup with almost no effort. Unlike plugins, they don't add new
   // commands. Instead, they hook themselves up into the test process.
   services: [
+    [TerraService],
+    [AssetService, {
+      ...SITE && { site: SITE },
+      ...LOCALE && { locale: LOCALE },
+      ...THEME && { theme: THEME },
+      ...fs.existsSync(defaultWebpackPath) && { config: defaultWebpackPath },
```
emilyrohrbough: how will consumers override this value? Will they need to re-define all the services? or remove this service & re-define it?

packages/terra-functional-testing/lib/services/wdio-asset-service.js
```diff
@@ -0,0 +1,45 @@
+const ExpressServer = require('../express-server/express-server');
+const WebpackServer = require('../webpack-server/webpack-server');
+const Logger = require('../logger/logger');
+
+const logger = new Logger({ prefix: 'wdio-asset-service' });
```
emilyrohrbough: Should this be name spaced with terra?

packages/terra-functional-testing/lib/services/wdio-asset-service.js
```diff
@@ -0,0 +1,45 @@
...
+
+class AssetService {
```
emilyrohrbough: To me, AssetService doesn't seem representative of the service's intent.... ServerService maybe? TerraServerService?     I can't say those are great. Since these toggle two flavors of static servers. It's almost like we should provide a default webpack config when a site is provided to use the [Copy Webpack Plugin](https://webpack.js.org/plugins/copy-webpack-plugin/) and serve these pre-compiles files with webpack-dev-server as well. 🤷‍♀️

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
```diff
@@ -0,0 +1,199 @@
...
+
+### locale
+
+An optional locale that will be passed into the webpack configuration as an environment variable.
+
+Type: `string`
+
+Required: `false`
+
+Default: `undefined`
```
emilyrohrbough: shouldn't this be english?

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
```diff
@@ -0,0 +1,199 @@
...
+
+### theme
+
+An optional theme that will be passed into the webpack configuration as an environment variable.
+
+Type: `string`
+
+Required: `false`
+
+Default: `undefined`
```
emilyrohrbough: Should this be terra-default-theme?

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
```diff
@@ -0,0 +1,199 @@
...
+
+### port
+
+The port the service will be hosted on.
+
+Type: `number`
+
+Required: `false`
+
+Default: `8080`
```
emilyrohrbough: Thought on making this default value different than 8080. 8080 is the default value used in webpack-dev-server. This would allow someone to kick off the build and run while also developing. Or if someone kicks of builds (like me...) and has another project's site running already, the wdio run ends up failing and hanging because of https://github.com/cerner/terra-toolkit-boneyard/issues/30

packages/terra-functional-testing/lib/config/wdio.conf.js
```diff
@@ -1,7 +1,19 @@
+const fs = require('fs');
+const path = require('path');
+const ip = require('ip');
```
emilyrohrbough, line 3: Do we have an option to override this default value via an env?

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
StephenEsser: There's a note in the existing service to avoid defaulting this value:

> // Explicitly not providing a fallback locale. Providing a fallback will lock the locale for all test runs when using the tt-wdio-runner.

https://github.com/cerner/terra-toolkit-boneyard/blob/master/src/wdio/services/ServeStaticService.js#L105

packages/terra-functional-testing/lib/config/wdio.conf.js
StephenEsser: They'd need to remove this service and redefine it. We could potentially expose an environment variable now that it represents a relative path.

This is a continuation of the discussion [here](https://github.com/cerner/terra-toolkit/pull/412#discussion_r438309783). The strategy for merging/overriding configurations isn't ideal, but it is something consumers will likely need to do from time to time. The second option in the linked thread might make it slightly easier for overriding options, but isn't ideal either.

packages/terra-functional-testing/lib/services/wdio-asset-service.js
StephenEsser: Do you have concerns about issues that might arise without a prefix? My thoughts on this is they are going to be scoped by:

```
@cerner/terra-functional-testing/path/wdio-asset-service
```

packages/terra-functional-testing/lib/services/wdio-asset-service.js
StephenEsser: I renamed this a few times before pushing out the PR, picking a descriptive name has been challenging. I have a comment in the summary asking for suggestions, hopefully we find something fitting

packages/terra-functional-testing/lib/config/wdio.conf.js
StephenEsser: I plan to carry these forward when they become more relevant, but I can go ahead and add them in this PR:

https://github.com/cerner/terra-toolkit-boneyard/blob/master/config/wdio/wdio.conf.js#L14-L20

I think there might be some issues with how these are defined today, but I'll do some debugging to check. The base url being routed matters quite a bit, if the internal and external ports are mapped differently I'd be curious if this would work as intended when using `browser.url`

packages/terra-functional-testing/lib/config/wdio.conf.js
emilyrohrbough: > They'd need to remove this service and redefine it.

This doesn't seem like a great option for our consumers..And this will lock us into the service order. I wouldn't expect it to change, but if for some reason it did, we'd need to maintain the same order for passivity.

packages/terra-functional-testing/lib/services/wdio-asset-service.js
emilyrohrbough: I'm confused. What is the logged output here then?

`[@cerner/terra-functional-testing/path/wdio-asset-service] log message...`

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
emilyrohrbough: Could we carry that note forward in the service?

packages/terra-functional-testing/lib/express-server/express-server.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    return new Promise((resolve, reject) => {
+      const app = this.createApp();
+
+      logger.info('Starting the express server.');
+
+      this.server = app.listen(this.port, this.host, (error) => {
+        if (error) {
+          reject(error);
```
yuderekyu, line 54: if we encounter an error and reject, do we want to immediately return to terminate execution?

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
```diff
@@ -0,0 +1,199 @@
...
+
+A relative path to a directory of static assets. If provided, an express server will be launched to serve the directory assets.
+
+Note: The `site` configuration option will take precedence over `config` if both are provided.
+
+Type: `string`
+
+Required: `false`
+
+Default: `undefined`
```
yuderekyu: i thought this defaults to `./build`

packages/terra-functional-testing/lib/config/wdio.conf.js
StephenEsser: Do you have any ideas for making this easier? This does not lock us into a service order, even with the current API. Consumers shouldn't be dice rolling the index of a service. They can read the service name from the array.

A couple ideas mentioned above that could help mitigate this would be the environment variables or the keyed configuration value for each service. I don't think either of these fully resolve the issue when merging wdio configurations with the webpack merge strategy, but they could help.

packages/terra-functional-testing/lib/services/wdio-asset-service.js
StephenEsser: The log output is going to be:

```
[terra-functional-testing:wdio-asset-service] Log message....
```

The import, if needed, would be:

```js
const AssetService = require('@cerner/terra-functional-testing/lib/services/wdio-asset-service');
```

packages/terra-functional-testing/lib/express-server/express-server.js
StephenEsser: Can you explain this a in a little more detail? The rejection will end the lifecycle of the promise. I can wrap the resolve logic below in an `else` block to more safely guard that logic, but there's nothing to return from here.

With these changes we can return or throw a special error message to end the service execution:

https://github.com/webdriverio/webdriverio/pull/5565

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
StephenEsser: `site` and `config` are mutually exclusive. When run through the service, if `site` is not defined the express server will never be launched. There's a fallback in the ExpressServer.js, but it will never be initialized through the service if `site` is not defined. We can probably remove the fallback logic. There's no default for the service options

packages/terra-functional-testing/lib/webpack-server/webpack-server.js
StephenEsser: Added an index.js file that exports the service, but I kept the original webpack-server.js file. This is a similar pattern to what is being done in terra-application and I think it makes sense. This allows the consistent shortcut import without dumping all of the code into the index.js file

https://github.com/cerner/terra-toolkit/pull/417/commits/7b43d2d93a6c308d7990995e74934d87b098aace

package.json
StephenEsser: Removed:

https://github.com/cerner/terra-toolkit/pull/417/commits/7b43d2d93a6c308d7990995e74934d87b098aace

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
StephenEsser: I'll add more documentation into this mdx doc, there's really no-where in the service this would fit into, but it could potentially be moved into the servers.js files

packages/terra-functional-testing/lib/config/wdio.conf.js
emilyrohrbough: >I think there might be some issues with how these are defined today, but I'll do some debugging to check.

What issues are you references? What have you found out?

packages/terra-functional-testing/lib/services/wdio-asset-service.js
emilyrohrbough: TerraServerService, TestServerService, AssetServerService.

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
emilyrohrbough: yeah...forgot about that. I think it's a bit odd. Maybe the tt-wdio service needs to be updated or changes to prevent locking that value in if that is the case. I seems a little shady for this service to rely on undefined to be english since the screenshot naming falls back to `en`: https://github.com/cerner/terra-toolkit-boneyard/blob/main/config/wdio/visualRegressionConf.js#L56. Though maybe you are changing this in the next PR?

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
emilyrohrbough: I agree, if we are saying it's undefined in the docs for the Asset Service and users need to pass an option via the config to use the Express Server, we should remove the fallback: https://github.com/cerner/terra-toolkit/pull/417/files#diff-f4769da54e7a208b92034dfac0100018R19

packages/terra-functional-testing/lib/services/wdio-asset-service.js
yuderekyu: maybe we do a poll?

ryanthemanuel: Maybe TerraServerService or TerraAssetServerService?

StephenEsser: While/as long as this service is inside of `terra-functional-testing` I don't think the `terra` prefix adds a lot of value. A few more names to drop into the options list:

- StaticAssetService
- ExpressService // Assuming it's possible to consolidate the two services
- WebpackService // Assuming it's possible to consolidate the two services

`ServerService` feels a little misleading, although there is indeed a server being launched, it doesn't make it obvious what its purpose is, the `AssetServerService` suggestion is more clear about the functionality of the service

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
StephenEsser: Removed fallback: https://github.com/cerner/terra-toolkit/pull/417/commits/9d536c1de47e906b3e6248b5bdb0ce0b5b0a8087

packages/terra-functional-testing/lib/services/wdio-asset-service.js
StephenEsser: Renamed to `AssetServerService`:

https://github.com/cerner/terra-toolkit/pull/417/commits/5427b46718561ff890cc508fcad2a3be61bb2998

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
StephenEsser: I updated this to default to the terra-default-theme. We'll need to evaluate down the road any quirks this might have

https://github.com/cerner/terra-toolkit/pull/417/commits/b45495bdd558b5c9e9c714f2dfa019b061822aac

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-service/AssetService.doc.mdx
StephenEsser: I've left the default locale as undefined for now, I'll look more closely at this when I begin carrying forward the visual regression functionality and see if we can default it to 'en' without introducing any issues

packages/terra-functional-testing/lib/config/wdio.conf.js
StephenEsser: Added the port and host overrides, it's been a little while since I wrote the comment above. I don't actually remember where I was headed with that. I think I may have thought there is a situation where the overrides could conflict but I haven't noticed any issues. 

https://github.com/cerner/terra-toolkit/pull/417/commits/1318f5fe0af7567ddce1fce9b712dbe5bee0da8e

packages/terra-functional-testing/lib/express-server/express-server.js
StephenEsser: I wrapped the resolve/reject resolution in if/else blocks, I believe this will mitigate any issues:

https://github.com/cerner/terra-toolkit/pull/417/commits/732f2c36f32221c7c2ae9f20105a7a749711243f

In a future PR I'll add the new error in the appropriate places to terminate test execution if the any of the service hooks fail

emilyrohrbough (approved):

packages/terra-functional-testing/lib/config/wdio.conf.js
mjhenkes: I think its fine to expect wepback config to be defined at root.

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-server-service/AssetServerService.doc.mdx
```diff
@@ -0,0 +1,199 @@
...
+    services: [
+        [AssetServerService, {
+          config: './path/webpack.config.js'
+        }]
+    ],
+    // ...
+};
+```
+
+### index
```
mjhenkes: Do we need to support this option? Do we have any evidence of anyone not using index for this?

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-server-service/AssetServerService.doc.mdx
```diff
@@ -0,0 +1,199 @@
...
+    services: [
+        [AssetServerService, {
+          site: './build',
+        }]
+    ],
+    // ...
+};
+```
+
+### theme
```
mjhenkes, line 149: do we want to group theme and locale as wepback options? since they won't apply to a static site?

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-server-service/AssetServerService.doc.mdx
```diff
@@ -0,0 +1,199 @@
...
+            // ...
+        }]
+    ],
+    // ...
+};
+```
+
+## Options
+
+### config
```
mjhenkes: I think we should call this out as `webpackConfig`.

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-server-service/AssetServerService.doc.mdx
StephenEsser: I carried this forward from the existing implementation, I'm not aware of anyone using it today

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-server-service/AssetServerService.doc.mdx
StephenEsser: I could go either way on this one. `webpackOptions` as an option won't apply to the static site either, it's a little awkward either way. We could potentially name them `webpackTheme` and `webpackLocale` or just add a note here they only apply to webpack.

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-server-service/AssetServerService.doc.mdx
StephenEsser: Removed index as a configurable option, we can always add this passively back in if necessary:

https://github.com/cerner/terra-toolkit/pull/417/commits/4d8bce55c326b1e1d67889270a32b925702ca29d

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-server-service/AssetServerService.doc.mdx
StephenEsser: Renamed to `webpackConfig`:

https://github.com/cerner/terra-toolkit/pull/417/commits/c6011cf084d1e24c1a1df7097dd293f21552a0ae

packages/terra-functional-testing/lib/terra-dev-site/doc/asset-server-service/AssetServerService.doc.mdx
mjhenkes: At a minimum, add a note that these options only apply to webpack builds. I'm not stuck on any option.

mjhenkes (approved):

PR #422, opened 2020-07-29
[terra-functional-testing] Replace CI ENV with optional ENV overrides

StephenEsser:
### Summary

This PR removes the `CI` ENV variable and adds explicit overrides for each customizable flag.

- Remove `CI` ENV
- Add `WDIO_HOSTNAME` ENV to set the hostname
- Add `WDIO_DISABLE_SELENIUM_SERVICE` ENV to disable the Selenium Service

### Additional Details

The CI env was previously used to setup a testing environment specifically for continuous integration builds. The one ENV variable would toggle on and off functionality in the WDIO configuration. This was overextending the functionality of one variable setting behavior. The goal of these changes are to allow ENVs be set to explicitly determine feature functionality instead of relying on a single variable to alter the behavior of test runs.

This PR is set as a draft until the test runs are stable on travis which may require some alterations to the timeouts in the selenium service.

@cerner/terra


mjhenkes (approved):

emilyrohrbough:
With this change, we likely need to consider the other places CI was used..the terra docker images (remove the CI option all together?) and double check if it is still used to toggle disabling terra-dev-site evidence screenshots when tests are running.

emilyrohrbough (approved):
Do we need to update or change our

emilyrohrbough:
Did we need to update the docs on these or add docs for these?

StephenEsser:
> Did we need to update the docs on these or add docs for these?

Not right now, there's no comprehensive docs for these variables available. Prior to releasing we should definitely have them available. I'll add a TODO onto the Epic issue and link it to the issue I'm creating for the consumer override strategy. If I get any downtime I'll follow this PR up with a new Read me for all the current ENVs and how to use them.

emilyrohrbough:
Sounds like a plan! 👍

PR #438, opened 2020-09-18
[terra-functional-testing] Adds viewport commands

benbcai:
### Summary
This PR adds the viewport commands (`viewports` and `describeViewports`) that are provided by TerraService.

Part of #413 

### Additional Details
- Adds hideInputCare
- Adds Jest Expect and Matchers

@cerner/terra


packages/terra-functional-testing/src/config/wdio.conf.js
```diff
@@ -165,4 +166,17 @@ exports.config = {
...
+    tiny: { width: 470, height: 768, name: 'tiny' },
+    small: { width: 622, height: 768, name: 'small' },
+    medium: { width: 838, height: 768, name: 'medium' },
+    large: { width: 1000, height: 768, name: 'large' },
+    huge: { width: 1300, height: 768, name: 'huge' },
+    enormous: { width: 1500, height: 768, name: 'enormous' },
+  },
+
+  /* Use to change the form factor (test viewport) used in the wdio run. */
+  ...FORM_FACTOR && { formFactor: FORM_FACTOR },
```
StephenEsser: These should be scoped to the service they are a part of, either as options or as nested keys. There are trade offs to both, but I think we should avoid un-scoped flat keys on the configuration object.

Service Options:
```js
  services: [
    [TerraService, {
      viewports: {
        tiny: { width: 470, height: 768, name: 'tiny' },
        small: { width: 622, height: 768, name: 'small' },
        medium: { width: 838, height: 768, name: 'medium' },
        large: { width: 1000, height: 768, name: 'large' },
        huge: { width: 1300, height: 768, name: 'huge' },
        enormous: { width: 1500, height: 768, name: 'enormous' },
      },
      ...FORM_FACTOR && { formFactor: FORM_FACTOR },
    }],
  ],
```

Scoped keys:

```
// wdio.config.js

module.exports = {
  terraServiceOptions: {
      viewports: {
        tiny: { width: 470, height: 768, name: 'tiny' },
        small: { width: 622, height: 768, name: 'small' },
        medium: { width: 838, height: 768, name: 'medium' },
        large: { width: 1000, height: 768, name: 'large' },
        huge: { width: 1300, height: 768, name: 'huge' },
        enormous: { width: 1500, height: 768, name: 'enormous' },
      },
      ...FORM_FACTOR && { formFactor: FORM_FACTOR },
  }
}
```
I'll work on going back through previous conversations and linking them here to show some of the tradeoffs/concerns. The most challenging scenario is how do consumers extend and override service options.

Edit: 

Discussions:

https://github.com/cerner/terra-toolkit/pull/412#discussion_r438309783

https://github.com/cerner/terra-toolkit/pull/417#discussion_r448727545

(I think there's one more thread floating around but I haven't found it yet)

packages/terra-functional-testing/src/config/wdio.conf.js
```diff
@@ -165,4 +166,17 @@ exports.config = {
...
+
+  /* Terra defined viewport sizes. */
+  terraViewports: {
+    tiny: { width: 470, height: 768, name: 'tiny' },
+    small: { width: 622, height: 768, name: 'small' },
+    medium: { width: 838, height: 768, name: 'medium' },
+    large: { width: 1000, height: 768, name: 'large' },
+    huge: { width: 1300, height: 768, name: 'huge' },
+    enormous: { width: 1500, height: 768, name: 'enormous' },
+  },
```
StephenEsser: With these being static it might be worth discussing if these should be defaulted within the service and excluded from being defined in this top level configuration

mjhenkes: Would these be changed for mpages tests? Do we know if anyone ever changes these? I lean towards excluding this from config.

packages/terra-functional-testing/src/commands/hide-input-caret.js
```diff
@@ -0,0 +1,20 @@
+const Logger = require('../logger/logger');
+
+/**
+ * Hides the blinking input caret that appears within editable text areas to prevent inconsistent test failures.
+ *
+ * @param {string} selector The selector for the element to hide the caret of
+ */
+const hideInputCaret = (selector) => {
+  try {
+    global.browser.execute(`document.querySelector("${selector.replace(/"/g, '\\"')}").style.caretColor = "transparent";`);
```
mjhenkes, line 10: I wonder if we should co-locate any of these browser.execute commands since they are theoretically the only things we need to transpile, i think?

packages/terra-functional-testing/src/commands/hide-input-caret.js
```diff
@@ -0,0 +1,20 @@
...
+  } catch (error) {
+    if (!global.browser.isExisting(selector)) {
+      throw Logger.error(`No element could be found with the selector '${selector}'.`);
```
StephenEsser: I'd be interested in what the catch error message is. Is this if/else block necessary if we're always throwing an error?

```js
} catch (error) {
  throw error;
}
```

packages/terra-functional-testing/src/commands/hide-input-caret.js
StephenEsser: I think this is safe to not transpile, but now I'm second guessing that. After the recent changes to compile the visual regression service I'm starting to wonder if we should just compile everything so we can use ES6 and avoid mix and matching transpiled/non-transpiled

mjhenkes: I'm not against that, considering my almost oops with the optional chaining bug in terra-toolkit-boneyard

yuderekyu: i forget, did we initially choose to mix and match for efficiency gains?

emilyrohrbough: The visual regression service had code that also needs to be transpiled

packages/terra-functional-testing/src/commands/viewport-helpers.js
```diff
@@ -0,0 +1,77 @@
...
+* @param {string|string[]} sizes - the list of Terra viewport names.
+* @return {object} - Terra viewport sizes.
+*/
+const getViewports = (...sizes) => {
+  const { terraViewports } = global.browser.config;
+  let viewportSizes = Object.keys(terraViewports);
+  if (sizes.length) {
+    viewportSizes = sizes;
+  }
+  return viewportSizes.map(size => terraViewports[size]);
```
emilyrohrbough:
```suggestion
  return viewportSizes.map(formFactor => terraViewports[formFactor]);
```

packages/terra-functional-testing/src/commands/viewport-helpers.js
```diff
@@ -0,0 +1,77 @@
...
+  const { terraViewports } = global.browser.config;
+  if (formFactor) {
+    const terraViewport = terraViewports[formFactor];
+    if (terraViewport !== undefined && typeof terraViewport === 'object') {
+      global.browser.setWindowSize(terraViewport.width, terraViewport.height);
+    } else {
+      throw logger.error(`The ${formFactor} formFactor supplied is not a viewport size supported by Terra.`);
+    }
+  } else {
+    const defaultViewport = terraViewports.huge;
```
emilyrohrbough: We should make sure to call this out in the documentation that huge is the default viewport size.

packages/terra-functional-testing/src/commands/viewport-helpers.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+  return viewportSizes.map(size => terraViewports[size]);
+};
+
+/**
+* Sets the viewport for the test run to the Terra test viewport size.
+*
+* @param {string} [formFactor=huge] - the Terra test viewport.
+*/
+const setViewport = (formFactor) => {
```
emilyrohrbough: If default is huge, should this be `(formFactor = 'huge')`? This would remove the if-else

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -1,13 +1,57 @@
...
   /**
    * Service hook executed prior to test execution.
    * Initializes the Terra Service's custom commands.
    */
-  before() {
+  before(capabilities) {
     global.browser.addCommand('axe', axe);
+
+    /* Add the Jest expect module the use the Jest matchers. */
+    global.expect = expect;
```
emilyrohrbough, line 20: Isn't this provided by default? Why do we need to add this explicitly?

emilyrohrbough (commented):
I know on the first few PR we put into Terra toolkit we said docs would come later, but I think we should start adding docs now....There was content that was missing in our previous docs and if we add little by little now, we should have decent doc coverage.

packages/terra-functional-testing/src/config/wdio.conf.js
```diff
@@ -129,7 +130,19 @@ exports.config = {
...
+      terraViewports: {
+        tiny: { width: 470, height: 768 },
+        small: { width: 622, height: 768 },
+        medium: { width: 838, height: 768 },
+        large: { width: 1000, height: 768 },
+        huge: { width: 1300, height: 768 },
+        enormous: { width: 1500, height: 768 },
+      },
+      /* Use to change the form factor (test viewport) used in the wdio run. */
+      ...FORM_FACTOR && { formFactor: FORM_FACTOR },
```
emilyrohrbough, line 135: Should we update this to be huge by default?
```suggestion
      formFactor: FORM_FACTOR || 'huge',
```

packages/terra-functional-testing/src/util/viewports.js
```diff
@@ -0,0 +1,10 @@
+const Viewports = {
```
emilyrohrbough:
```suggestion
const viewports = {
```

packages/terra-functional-testing/src/config/wdio.conf.js
emilyrohrbough: These won't change in MPages

emilyrohrbough:
@benbcai any updates?

packages/terra-functional-testing/src/commands/viewport-helpers.js
```diff
@@ -0,0 +1,71 @@
...
+  let localViewports = viewports;
+
+  // If formFactor is defined and viewports contains the form factor, run that size else don't run the tests
+  const { formFactor } = global.browser.config;
+  if (formFactor) {
+    localViewports = viewports.includes(formFactor) ? [formFactor] : [];
+  }
+
+  let currentViewportSize;
+  localViewports.forEach(viewport => global.describe(`[${viewport}]`, () => {
```
emilyrohrbough, line 59: Doesn't matter, but we should consider using [describe.each](https://jestjs.io/docs/en/api#describeeachtablename-fn-timeout)

packages/terra-functional-testing/src/commands/viewport-helpers.js
```diff
@@ -0,0 +1,71 @@
...
+    global.before(() => {
```
emilyrohrbough: Should this be jest's `beforeAll` hook? I think `before` is mocha's hook

packages/terra-functional-testing/src/commands/viewport-helpers.js
```diff
@@ -0,0 +1,71 @@
...
+      currentViewportSize = global.browser.getWindowSize();
+      setViewport(viewport);
+    });
+    global.describe(title, fn);
+    global.after(() => global.browser.setWindowSize(currentViewportSize.width, currentViewportSize.height));
```
emilyrohrbough:
```suggestion
    global.afterAll(() => global.browser.setWindowSize(currentViewportSize.width, currentViewportSize.height));
```

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -1,13 +1,57 @@
...
+        if (global.browser.isExisting('[data-terra-dev-site-loading]')) {
+          global.browser.waitUntil(() => (
+            global.browser.isExisting('[data-terra-dev-site-content]')
+          ), global.browser.config.waitforTimeout + 2000, '', 100);
+        }
+      } catch (err) {
+        // intentionally blank
+        // if this fails we don't want to warn because the user can't fix the issue
+      }
+    }
```
emilyrohrbough: Should this be broken out into it's own afterCommand?

packages/terra-functional-testing/src/commands/viewport-helpers.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/a60b79003e729c292601de5f3d3ba40078b02ccd

packages/terra-functional-testing/src/commands/viewport-helpers.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/a60b79003e729c292601de5f3d3ba40078b02ccd

packages/terra-functional-testing/src/util/viewports.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/4d600076e3df0e9c99b327e4e0be28289810fa0a

packages/terra-functional-testing/src/config/wdio.conf.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/4d600076e3df0e9c99b327e4e0be28289810fa0a

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: This is not available in the before hook but we need it below to setup the Terra commands.

packages/terra-functional-testing/src/commands/viewport-helpers.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/5cf714558e6763787a83d19543932a2f57bc0c76

packages/terra-functional-testing/src/commands/hide-input-caret.js
benbcai: I think this was a mistake. I updated it [here](https://github.com/cerner/terra-toolkit/pull/438/commits/4d600076e3df0e9c99b327e4e0be28289810fa0a) to just log the message instead of throwing the message. If the selector does not exist, which is a common case, we would get a message like this `javascript error: Cannot read property 'style' of null`. I think logging that specific message is more helpful. Other types of error would just get thrown as is.

packages/terra-functional-testing/src/config/wdio.conf.js
```diff
@@ -165,4 +166,17 @@ exports.config = {
     ui: 'bdd',
     timeout: 60000,
   },
+
+  /* Terra defined viewport sizes. */
+  terraViewports: {
+    tiny: { width: 470, height: 768, name: 'tiny' },
```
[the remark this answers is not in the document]
benbcai: Updated and moved this our of the config: https://github.com/cerner/terra-toolkit/pull/438/commits/3db345e69d9b12382795557acce0285b0bd8abbc

packages/terra-functional-testing/src/config/wdio.conf.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/3db345e69d9b12382795557acce0285b0bd8abbc

packages/terra-functional-testing/src/config/wdio.conf.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/3db345e69d9b12382795557acce0285b0bd8abbc

packages/terra-functional-testing/src/commands/viewport-helpers.js
benbcai: This has been updated.

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: Are you referring to the hideInputCaret part or the part about waitForTimeout for terra-dev-site loading?

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -2,18 +2,58 @@
 const expect = require('expect');
 const { accessibility } = require('../commands/validates');
 const { toBeAccessible } = require('../commands/expect');
+const viewportHelpers = require('../commands/viewport-helpers');
```
StephenEsser: nit: obscuring the functions onto a random helper variable here feels strange. Can this be deconstructed into the specific functions? These could probably be organized a little differently and the function names could match 1-1 the global functions but that's not really necessary for this PR.

```js
const { getViewports, describeViewports, setViewport } = require('../commands/viewport-helpers');
```

packages/terra-functional-testing/src/config/wdio.conf.js
```diff
@@ -129,7 +130,10 @@ exports.config = {
   // your test setup with almost no effort. Unlike plugins, they don't add new
   // commands. Instead, they hook themselves up into the test process.
   services: [
-    [TerraService],
+    [TerraService, {
+      /* Use to change the form factor (test viewport) used in the wdio run. */
+      formFactor: FORM_FACTOR || 'huge',
```
StephenEsser: Can you add some documentation for this option in the TerraService MDX file?

packages/terra-functional-testing/src/commands/viewport-helpers.js
```diff
@@ -0,0 +1,71 @@
...
+* Sets the viewport for the test run to the Terra test viewport size.
+*
+* @param {string} [formFactor=huge] - the Terra test viewport.
+*/
+const setViewport = (formFactor = 'huge') => {
+  const terraViewport = terraViewports[formFactor];
+  if (terraViewport !== undefined && typeof terraViewport === 'object') {
+    global.browser.setWindowSize(terraViewport.width, terraViewport.height);
+  } else {
+    throw logger.error(`The ${formFactor} formFactor supplied is not a viewport size supported by Terra.`);
```
StephenEsser: This would `throw undefined`. The logger doesn't return an error. It logs a formatted error message to the terminal

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -2,18 +2,58 @@
...
+      viewports: viewportHelpers.getViewports,
+
+      /* `describeViewports` provides a custom Mocha `describe` block for looping test viewports. */
+      describeViewports: viewportHelpers.describeViewports,
+
+      /* `hideInputCaret` hides the blinking input caret that appears in inputs or editable text areas. */
+      hideInputCaret,
+    };
+
+    /* IE driver takes a longer to be ready for browser interactions. */
```
StephenEsser: Looks like there is an extra `a` or missing content in this comment

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -2,18 +2,58 @@
...
+    if (capabilities.browserName === 'internet explorer') {
+      global.browser.pause(10000);
```
StephenEsser: Does `waitForExist` work at this point in the service?

https://webdriver.io/docs/api/element/waitForExist.html

The pause documentation sort of applies to this situation, the selector could potentially be something on the page (html/body tag maybe?)

https://webdriver.io/docs/api/browser/pause.html

> Pauses execution for a specific amount of time. It is recommended to not use this command to wait for an element to show up. In order to avoid flaky test results it is better to use commands like waitForExist or other waitFor* commands.

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -2,18 +2,58 @@
...
+    }
+
+    /* Set the viewport size before the spec begins. */
+    viewportHelpers.setViewport(global.browser.config.formFactor);
```
StephenEsser: I don't think the `formFactor` key exists on the root configuration anymore. Should we be retrieving the form factor from the service options?

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -2,18 +2,58 @@
...
     global.expect = expect;
     global.expect.extend({ toBeAccessible });
 
-    global.Terra = {};
-    global.Terra.validates = { accessibility };
+    /* Add a Terra global with access to Mocha-Chai test helpers. */
+    global.Terra = {
+      validates: { accessibility },
+
+      /* `viewports` provides access Terra's list of test viewports. */
```
StephenEsser: nit: Do the back-ticks being used here do/mean something?

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -2,18 +2,58 @@
...
+  afterCommand(commandName, args, result, error) {
+    if ((commandName === 'refresh' || commandName === 'url') && !error) {
+      try {
+        // This is only meant as a convenience so failure is not particularly concerning
+        global.Terra.hideInputCaret('body');
+
+        if (global.browser.$('[data-terra-dev-site-loading]').isExisting()) {
+          global.browser.waitUntil(() => (
+            global.browser.$('[data-terra-dev-site-content]').isExisting()
+          ), global.browser.config.waitforTimeout + 2000, '', 100);
```
StephenEsser: Would waitForExists work here?

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/19c190061a6d0f428d0ab4b33b669bed04d9f219

packages/terra-functional-testing/src/config/wdio.conf.js
benbcai: Added doc. https://github.com/cerner/terra-toolkit/pull/438/commits/19c190061a6d0f428d0ab4b33b669bed04d9f219

packages/terra-functional-testing/src/commands/viewport-helpers.js
benbcai: This should just log the message. https://github.com/cerner/terra-toolkit/pull/438/commits/19c190061a6d0f428d0ab4b33b669bed04d9f219

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/19c190061a6d0f428d0ab4b33b669bed04d9f219

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: Good call. The form factor was initially at the root config but has since been moved to the service option level and this wasn't updated. https://github.com/cerner/terra-toolkit/pull/438/commits/19c190061a6d0f428d0ab4b33b669bed04d9f219

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: No, they just add more emphasis to these key names. They are now removed. https://github.com/cerner/terra-toolkit/pull/438/commits/19c190061a6d0f428d0ab4b33b669bed04d9f219

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: Yes, I like `waitForExists` here as it is simpler: https://github.com/cerner/terra-toolkit/pull/438/commits/19c190061a6d0f428d0ab4b33b669bed04d9f219

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: `waitForExist` requires a selector and there isn't a good selector to use here. Looking back at #183 that added this 10 seconds was the suggested wait time.

packages/terra-functional-testing/src/commands/viewport-helpers.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/3cd30c44ecf7cee37cafa3e12b9c4b129f38fb5b

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
```diff
@@ -2,6 +2,32 @@
 
 The terra-functional-testing library is a utility for developing automation tests. The library extends the [WebdriverIO Framework](https://webdriver.io/) to facilitate automating accessibility and functional testing for Terra projects.
 
+## Viewports
+
+Tests can be executed in any of the following supported viewports:
+- 'tiny' - width: 470; height 768
```
ryanthemanuel: We should be consistent with the semi colons:

```suggestion
- 'tiny' - width: 470; height: 768
```

packages/terra-functional-testing/src/commands/viewport-helpers.js
```diff
@@ -0,0 +1,73 @@
+const Logger = require('../logger/logger');
+const terraViewports = require('../util/viewports');
```
StephenEsser: Nit: I'd probably import this as a constant. It doesn't really matter, just a little more standard.

```js
const { TERRA_VIEWPORTS } = require('../util/viewports'); // This directory/file name makes it sounds functional rather than static constants
```

packages/terra-functional-testing/src/commands/viewport-helpers.js
```diff
@@ -0,0 +1,73 @@
...
+*
+* @param {string|string[]} sizes - the list of Terra viewport names.
+* @return {object} - Terra viewport sizes.
+*/
+const getViewports = (...sizes) => {
+  let viewportSizes = Object.keys(terraViewports);
+  if (sizes.length) {
+    viewportSizes = sizes;
+  }
+  return viewportSizes.map(formFactor => terraViewports[formFactor]);
```
StephenEsser: Are `formFactor` and `viewport` synonyms or is `formFactor` always the 'tiny, medium, small' and viewport the value (width, height)? The words seem interchangeable throughout this file.

packages/terra-functional-testing/src/commands/viewport-helpers.js
```diff
@@ -0,0 +1,73 @@
...
+};
+
+/**
+* Sets the viewport for the test run to the Terra test viewport size.
+*
+* @param {string} [formFactor=huge] - the Terra test viewport.
+*/
```
StephenEsser: nit: Each of these comment blocks are slightly off on their indentation

StephenEsser (approved):

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/402ee523e9eabedfc323de69d40318f8b075dd7d

packages/terra-functional-testing/src/commands/viewport-helpers.js
benbcai: That makes sense. Updated and changed the directory name from util to static-assets. https://github.com/cerner/terra-toolkit/pull/438/commits/402ee523e9eabedfc323de69d40318f8b075dd7d

packages/terra-functional-testing/src/commands/viewport-helpers.js
benbcai: `formFactor` is referring more to the FORM_FACTOR set in the env variable whereas `viewport` is referring to the window sizes we define. But yes, these seem to be used interchangeably. I've updated the variable names here and comment too `viewpoort` whenever referring to what we defined and use `formFactor` when referring to the env option. https://github.com/cerner/terra-toolkit/pull/438/commits/402ee523e9eabedfc323de69d40318f8b075dd7d

packages/terra-functional-testing/src/commands/viewport-helpers.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/402ee523e9eabedfc323de69d40318f8b075dd7d

packages/terra-functional-testing/src/commands/hide-input-caret.js
emilyrohrbough: There are a bunch of scripts co-located for the visual-regression service here: https://github.com/cerner/terra-toolkit/tree/main/packages/terra-functional-testing/src/services/wdio-visual-regression-service/scripts

Might not be bad to follow this pattern later on. We can also use old-school func syntax for this to run in IE
```js
global.browser.execute(function () {
  document.querySelector("${selector.replace(/"/g, '\\"')}").style.caretColor = "transparent";`
})
```

packages/terra-functional-testing/src/commands/hide-input-caret.js
```diff
@@ -0,0 +1,23 @@
...
+ * @param {string} selector The selector for the element to hide the caret of
+ */
+const hideInputCaret = (selector) => {
+  try {
+    global.browser.execute(`document.querySelector("${selector.replace(/"/g, '\\"')}").style.caretColor = "transparent";`);
+  } catch (error) {
+    const element = global.browser.$(selector);
+    if (!element.isExisting()) {
+      logger.error(`No element could be found with the selector '${selector}'.`);
+    } else {
```
emilyrohrbough: Should we check if the element exists before trying to update the style so a try-catch isn't needed?

packages/terra-functional-testing/src/commands/viewport-helpers.js
```diff
@@ -0,0 +1,73 @@
...
+ * @param {string[]} viewports - The list of Terra viewports to tests if formFactor in the service options or env is not set.
+ * @param {function} - the test function to execute for each viewport.
+ */
+const describeViewports = (title, viewports, fn) => {
+  let localViewports = viewports;
+
+  // If formFactor is defined in the service options/env and viewports contains this form factor, run that size else don't run the tests
+  const [, options = {}] = global.browser.options.services.find(([service]) => (
+    typeof service === 'function' && service.name === 'TerraService'
+  ));
```
emilyrohrbough, line 51: Would it be possible to initialize the viewport helpers as a class so we don't have to pull the configuration options from the global object?

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -2,18 +2,68 @@
...
+      global.browser.$('body').waitForExist({
+        timeout: global.browser.config.waitforTimeout,
+        interval: 100,
+      });
+    }
+
+    // Extract the Terra service options from the global browser object.
+    const [, options = {}] = global.browser.options.services.find(([service]) => (
+      typeof service === 'function' && service.name === 'TerraService'
+    ));
```
emilyrohrbough: We should add a constructor and set the formFactor like:
```
class TerraService {
  constructor(options) {
    this.formFactor = options.formFactor;
  }
  ..
   setViewport(this.formFactor);

```
we shouldn't have to pull this config from the global here

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-services/TerraService.tool.mdx
```diff
@@ -55,10 +55,42 @@ export.config = {
...
 
+### formFactor
+
+Tests can be executed in a specific form factor by setting the `formFactor` configuration option or the `FORM_FACTOR` environment variable . The form factors can be any of the following supported viewports: `tiny`; `small`; `medium`; `large`; `huge`; `enormous`. In order for the tests to run in this form factor, it must be one that is already specified in the `Terra.describeViewport` block. 
+
+Type: `String`
+
+Required: `false`
+
+Default: `undefined`
```
emilyrohrbough, line 73:
```suggestion
Default: `huge`
```

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
```diff
@@ -2,6 +2,32 @@
...
+
+Tests can be executed in any of the following supported viewports:
+- 'tiny' - width: 470; height: 768
+- 'small' - width: 622; height: 768
+- 'medium' - width: 838; height: 768
+- 'large' - width: 1000; height: 768
+- 'huge' - width: 1300; height: 768
+- 'enormous' - width: 1500; height: 768
+
+### Terra.describeViewports
```
emilyrohrbough, line 15: Should this live in the TerraService docs?

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
```diff
@@ -34,3 +60,18 @@ it('should override a rule configuration', () => {
   Terra.validates.accessibility(options);
 });
 ```
+
+## Input Caret
+
+### Terra.hideInputCaret
+
+An editable text field in focus will have a blinking caret. Often times this blinking caret causes inconsistent test failures due to the blinking of the caret during screenshot capture. This situation can be avoided by using `Terra.hideInputCaret` to set the CSS caret color to of the element to be transparent. This method must be placed in a `before`, `beforeEach`, or `it` block or it will not be ran. This method accepts a selector string as an argument and will only apply to the first selector if multiple are found.
```
emilyrohrbough: We should mention we will run this automatically on page load for body and is likely handled already

packages/terra-functional-testing/src/services/wdio-terra-service.js
emilyrohrbough: Of my latests comments, this is the main one that I feel really needs to be addressed.

packages/terra-functional-testing/src/commands/hide-input-caret.js
benbcai: Yes, I think that makes more logical sense. If the selector isn't there, we're just log the message and no need to throw an error. https://github.com/cerner/terra-toolkit/pull/438/commits/825dac196d8e70c66767ce89f6479d4e73a75951

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-services/TerraService.tool.mdx
benbcai: As discussed in the meeting that we don't want a default here as this is defined by the env variable. Tests will only run in this form factor and not ones defined in the test is this is set.

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
benbcai: I put this here because this explains how to write a wdio test using describeViewport.

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/438/commits/825dac196d8e70c66767ce89f6479d4e73a75951

packages/terra-functional-testing/src/commands/viewport-helpers.js
benbcai: What are the concerns about accessing the global object here?

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: Would this mean that anywhere that initializes TerraService would need to first get the options? My thought was to abstract that and get the service option here.

StephenEsser: The TerraService is initialized with the options, we can save those and then use them directly rather than looking for them on the global config.

Here's an example of the WDIO sauce service:

https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-sauce-service/src/service.js#L10-L14

The utility functions need to search for the global options since they don't have a reference to `this` for the TerraService (although we might be able to bind them with the options).

emilyrohrbough: Yes, but we are initializing this in the wdio configuration 

UPDATE: Stephen did a way better job as describing this.

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
StephenEsser: We'll probably need documentation in both places, but I think it makes sense to document how to use the commands in the WDIO Testing section

packages/terra-functional-testing/src/commands/viewport-helpers.js
emilyrohrbough: More than this is messy and it seems like these options could have been passed in by the TerraService when it defined this

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: After running some tests and saw that the service options is passed to the service and already available. So yes, getting it in the constructor is more efficient. https://github.com/cerner/terra-toolkit/pull/438/commits/bf231c014c5f0e7bb7c11400e6fe1b314f146a41

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -2,18 +2,67 @@
...
-    global.Terra.validates = { accessibility };
+    // Add a Terra global with access to Mocha-Chai test helpers.
+    global.Terra = {
+      validates: { accessibility },
+
+      // viewports provides access Terra's list of test viewports.
+      viewports: getViewports,
+
+      // describeViewports provides a custom describe block for looping test viewports.
+      describeViewports,
```
emilyrohrbough, line 31: If you bind this, can we access the formFactor in the describeViewports command like `this.formFactor`?

```js
describeViewports: describeViewports.bind(this)
```

emilyrohrbough (approved):

ryanthemanuel (approved):

PR #448, opened 2020-10-14
[terra-functional-testing] Upgrade axe-core to v4.0.2

### LINKED ISSUE #426: [terra-functional-testing] Upgrade to axe-core to 4.0.2
# Feature Request

## Description
axe-core v4.0.0 has been released. We should look at updating this with the wdio v6 uplift.

StephenEsser:
### Summary

This PR upgrades axe-core to v4.0.2.

Resolves https://github.com/cerner/terra-toolkit/issues/426

### Additional Details

This is a major version bump for axe-core. Release notes and breaking changes can be viewed here.

https://github.com/dequelabs/axe-core/releases/tag/v4.0.0

@cerner/terra


packages/terra-functional-testing/package.json
```diff
@@ -52,7 +52,7 @@
     "@wdio/mocha-framework": "^6.1.0",
     "@wdio/spec-reporter": "^6.0.16",
     "@wdio/sync": "^6.1.14",
-    "axe-core": "3.5.5",
+    "axe-core": "4.0.2",
```
emilyrohrbough, line 55: What new rules have been added between 3.5.5 and do we plan to fail on all violations?

emilyrohrbough: Also, did we intent to adjust the rules we run to align with our standards?
https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/commands/run-axe.js#L35

Recently we talked how these are not additive but distinct

StephenEsser: The [v4 axe-core release notes](https://github.com/dequelabs/axe-core/releases/tag/v4.0.0) are posted in the summary. Because this is a new package I would expect all new rules to be enabled, but we can warn on rules case by case

StephenEsser: > Also, did we intent to adjust the rules we run to align with our standards?

> Recently we talked how these are not additive but distinct

I don't think I was involved in that conversation. Align with the standards defined here?

https://engineering.cerner.com/terra-ui/about/terra-ui/component-standards#accessibility-a11y

emilyrohrbough: > Because this is a new package I would expect all new rules to be enabled, but we can warn on rules case by case

Typically I would agree, however was this testing against the terra repos? Any rule we fail will cause failures for our consumers as well. It'd be good to have an idea what to expect.

emilyrohrbough: Looking at the release notes, I wasn't sure which rules were added, I only found the removed /deprecated list...

Would this new feature be useful for our recent discussions on not failing on specific rules? https://github.com/dequelabs/axe-core/pull/2235

StephenEsser: I would expect the terra repositories to be the first adopters of the terra-functional-testing package and axe-core upgrades as part of this project. If we choose not to upgrade now, we'll need to major version bump consumers twice to get onto the latest set of axe core rules. Because some rules were changed, rather than added/deleted, some rules may not be able to be switched to warnings.

I'll do some investigation into the severity of the rule changes against some of our existing components and post the findings. However, bug fixes, enhancements, and current standards are being developed on axe-core v4 and I think terra-functional-testing should release with the latest available version of axe-core

Edit:

Initial investigation testing axe-core v4 with terra-core resulted in 32 errors (out of nearly 1500 tests) and seems to be limited/scoped to ~3 packages.

https://gist.github.com/StephenEsser/7e3e2f3dbeab5a2ca44bdc3b16e7be3b

StephenEsser: > Would this new feature be useful for our recent discussions on not failing on specific rules? dequelabs/axe-core#2235

Yes, I believe so. This is the current approach I am taking to report warnings and utilize the axe-core API directly rather than adding our own

StephenEsser: > Looking at the release notes, I wasn't sure which rules were added, I only found the removed /deprecated list...

I don't think any new rules were added. Here's a blog I found that was pretty interesting regarding the new axe linter and quotes no new rules were added:

>  No new rules this time, but new opportunities instead.

https://www.deque.com/blog/announcing-axe-core-4-0/

emilyrohrbough: I'm not disagreeing we should be on the latest version, just would be good to log issues for the errors we know exist. That's great there weren't many! We can log issues with work-arounds / guidance and hopefully those issues are resolved by the time teams upgrade.

emilyrohrbough: > I don't think I was involved in that conversation.  

yeah, @ryanthemanuel / @mjhenkes knows more, but I think we need to drop wcag2a and wcag21aa from the list we run so we align with the Terra standards.

mjhenkes: wcag2a is still part of our standards, and wcag21aa we might as well leave since we will want to get there anyway.

mjhenkes: in the future we would add wcag22aa as warning before moving it to an error.

mjhenkes (approved):

packages/terra-functional-testing/package.json
benbcai: When we upgraded axe-core from 3.2.0 to 3.5.5, we ran into several different types of violations in our repos for the new rules they added. We fixed all of them with the exception of the `scrollable-region-focusable` rule, which we currently [disabled](https://github.com/cerner/terra-toolkit-boneyard/blob/main/src/wdio/services/TerraCommands/chai-methods.js#L18). We logged [issues](https://github.com/cerner/terra-framework/issues/991) to each repo to figure out how best to handle this violation. I think we'll have to continue to disable this rule until we have a solution for it.

In comparing the rule sets in [3.5.5](https://github.com/dequelabs/axe-core/tree/v3.5.5/lib/rules) and [4.0.2](https://github.com/dequelabs/axe-core/tree/v4.0.2/lib/rules), the only new rule I see is [select-name](https://github.com/dequelabs/axe-core/blob/v4.0.2/lib/rules/select-name.json) which was added in the .2 patch version. Not sure why a new rule was added in a patch version.

StephenEsser: @benbcai the `scrollable-region-focusable` rule is now being disabled in https://github.com/cerner/terra-toolkit/pull/452. The select-name rule does not appear to report a violation for our native select component. Are there any other changes you want to recommend on this PR prior to merging?

benbcai (approved):

PR #452, opened 2020-10-15
[terra-functional-testing] Add accessibility testing validation assertion

StephenEsser:
### Summary

This PR adds the Terra.validates.accessibility command globally to the WDIO test run. The command runs an expect assertion on the axe results of the current document when invoked.

The assertion will fail the current test step and return the violation message if accessibility violations are found.

### Additional Details

This PR removes the global browser axe command that was previously added. Consumers should be using the Terra helpers and the terra helpers can invoke axe manually without exposing the command.

Note: Accessibility warnings/reporter will be added on a different pull request.

@cerner/terra


packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -1,13 +1,19 @@
...
 
 class TerraService {
   /**
    * Service hook executed prior to test execution.
    * Initializes the Terra Service's custom commands.
    */
   before() {
-    global.browser.addCommand('axe', axe);
+    global.expect = expect;
+    global.expect.extend({ toBeAccessible });
```
emilyrohrbough, line 13: Is this configurable from the global wdio level? Lowlight-theme will need color-contrast disabled by default.

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
```diff
@@ -0,0 +1,32 @@
+# WebDriverIO Testing
+
+The terra-functional-testing library is a utility for developing tests using the [WebdriverIO Framework](https://webdriver.io/).
+
+## Accessibility Testing
+
+The testing library integrates [axe-core](https://github.com/dequelabs/axe-core) utilities for automating accessibility testing.
```
emilyrohrbough:
```suggestion
The testing library integrates [axe-core](https://github.com/dequelabs/axe-core) utilities for automating accessibility testing of the current dom composition.
```

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+### Terra.validates.accessibility
+
+The accessibility assertion is available globally. Running the assertion will run acessibility checks on the current document. If acessibility violations are found the the test step will fail.
```
emilyrohrbough:
```suggestion
The accessibility assertion is available globally. Running the assertion will run accessibility checks on the current document. If accessibility violations are found, the the test step will fail. 
```

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
emilyrohrbough: Can we add docs on what axe rules we opt-in and test? It might be useful to add info that this captures only a fraction of accessibility and should be ran as various elements are changes / interacted with.

packages/terra-functional-testing/tests/jest/commands/expect/toBeAccessible.test.js
```diff
@@ -0,0 +1,42 @@
...
+  it('should return a message function that indicates the reason for the assertion failure', () => {
+    runAxe.mockImplementationOnce(() => ({
+      result: {
+        violations: [],
+      },
+    }));
+
+    const result = toBeAccessible();
+
+    expect(result.message()).toEqual('expected no accessibility violations but received []');
```
emilyrohrbough, line 40: How does this read when there is an existing violation?

packages/terra-functional-testing/src/commands/validates/index.js
```diff
@@ -0,0 +1,5 @@
+const accessibility = require('./accessibility');
```
emilyrohrbough, line 1: Just curious, why add the extra files to maintain and drill through instead of adding the src directly to index?

packages/terra-functional-testing/tests/jest/commands/validates/index.test.js
```diff
@@ -0,0 +1,7 @@
+const { accessibility } = require('../../../../src/commands/validates');
+
+describe('index', () => {
+  it('should export the accessibility function', () => {
+    expect(accessibility).toBeDefined();
```
emilyrohrbough, line 5: Was there a test for this helper?

packages/terra-functional-testing/src/services/wdio-terra-service.js
StephenEsser: Yes, the global axe rule overrides are already implemented in the Terra service/axe run functionality

packages/terra-functional-testing/tests/jest/services/wdio-terra-service.test.js
```diff
@@ -1,17 +1,11 @@
 const TerraService = require('../../../src/services/wdio-terra-service');
 
 describe('WDIO Terra Service', () => {
-  it('should add the axe command to the browser object', () => {
```
emilyrohrbough, line 4: Should we validate the expect helper is also available?

packages/terra-functional-testing/tests/wdio/terra-validates-spec.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * This spec file is used for integration testing the Terra.validates commands.
+ */
+
+describe('Axe', () => {
```
emilyrohrbough:
```suggestion
describe('Terra.validates', () => {
```

packages/terra-functional-testing/tests/jest/commands/expect/toBeAccessible.test.js
StephenEsser: It prints out all violations in the array

emilyrohrbough:
Can we add a migration documentation section for moving from terra-toolkit to terra-functional-testing for the axe helpers? It should be light wight, but would be good to call out the same or differing behaviors and APIs.

packages/terra-functional-testing/src/commands/expect/toBeAccessible.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Invokes axe analysis on the current document and asserts there are no accessibility violations.
+ * @param {*} _ - Unused param. The expect is always run on the current document.
+ * @param {Object} options - Optional axe configuration overrides.
+ * @returns {Object} - An object that indicates if the assertion passed or failed with a message.
+ */
+function toBeAccessible(_, options = {}) {
+  const { result } = runAxe(options);
+  const { violations } = result;
```
emilyrohrbough, line 11: Do we need to pull forward this disabled rule? 

https://github.com/cerner/terra-toolkit-boneyard/blob/main/src/wdio/services/TerraCommands/chai-methods.js#L11

emilyrohrbough (commented):
EDIT...not sure what this comment is...just ignore 😅

packages/terra-functional-testing/src/commands/validates/index.js
StephenEsser: Personal preference, in general I think it's bad practice to overextend a single file. The individual files make reading and testing easier/isolated/explicit and allows scaling as more commands are added

packages/terra-functional-testing/tests/jest/commands/validates/index.test.js
StephenEsser: There's not a jest test. This helper is being regression tested in WDIO. Setting up, mocking, and extending the internal `expect` used inside the function was a little tricky

packages/terra-functional-testing/src/commands/expect/toBeAccessible.js
StephenEsser: Yes, I think we should disable it in a slightly different way though. I think it should be disabled prior to running axe, rather than filtering it out of the returned violations. I'll work on adding it into this PR

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
StephenEsser: Updated the docs to add more information on rules and how axe runs on the document:

https://github.com/cerner/terra-toolkit/pull/452/commits/1aad2e2c77dc3a59fed32b3ce8c763a95b57d952

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
StephenEsser: Fixed typos and enabled my spell checker for .mdx files:

https://github.com/cerner/terra-toolkit/pull/452/commits/1aad2e2c77dc3a59fed32b3ce8c763a95b57d952

packages/terra-functional-testing/tests/wdio/terra-validates-spec.js
StephenEsser: Updated to `Terra.validates.accessibility`:

https://github.com/cerner/terra-toolkit/pull/452/commits/1aad2e2c77dc3a59fed32b3ce8c763a95b57d952

packages/terra-functional-testing/tests/jest/services/wdio-terra-service.test.js
StephenEsser: I haven't found a good way to use `expect` to mock `expect` and verify a custom `expect` is available if that makes sense. This gets tested indirectly via WDIO. I tried using `jest.spyOn(expect, 'extend')` but it didn't return the results I was expecting. I can probably mock axe and call the helper directly, but I don't know what that would really be testing. Do you have any ideas for a way to test this in jest?

packages/terra-functional-testing/src/commands/expect/toBeAccessible.js
StephenEsser: Disabled the rule prior to calling axe, rather than filtering it out of the violation array:

https://github.com/cerner/terra-toolkit/pull/452/commits/1aad2e2c77dc3a59fed32b3ce8c763a95b57d952

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
```diff
@@ -1,24 +1,28 @@
...
 
 ## Accessibility Testing
 
-The testing library integrates [axe-core](https://github.com/dequelabs/axe-core) utilities for automating accessibility testing.
+The testing library integrates [axe-core](https://github.com/dequelabs/axe-core) into the testing framework to help automate accessibility testing along side functional testing.
+
+Axe will analyze the entire document when run and report accessibility violations. The following [tags](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#axe-core-tags) are enabled `wcag2a`, `wcag2aa`, `wcag21aa`, and `section508`. Each tag has an associated list of [rules](https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md) that will be checked against the document when axe is run.
+
+Please note that not all accessibility testing can be automated. Axe provides a lightweight static analysis of the entire document to catch common accessibility violations, but it is the responsibility of each team and application to do thorough accessibility and functional testing manually when necessary.
```
emilyrohrbough, line 11: Nice description! 👍

emilyrohrbough (approved):
I think the only thing "missing" per-say at this point is a migration guide from terra-toolkit. It doesn't need to block this PR, but I think it'd be easier to add to it in small chunks rather than trying to collect/organize it at the ned.

StephenEsser:
> I think the only thing "missing" per-say at this point is a migration guide from terra-toolkit. It doesn't need to block this PR, but I think it'd be easier to add to it in small chunks rather than trying to collect/organize it at the ned.

I'm working on a migration guide to back fill some documentation for things we've changed and implemented. I'd like to add the initial migration guide on a follow up PR so discussions and topics don't block this PR. These code changes are necessary for the accessibility reporter.

packages/terra-functional-testing/src/commands/expect/toBeAccessible.js
```diff
@@ -0,0 +1,19 @@
+const { runAxe } = require('../axe');
+
+/**
+ * Invokes axe analysis on the current document and asserts there are no accessibility violations.
+ * @param {*} _ - Unused param. The expect is always run on the current document.
```
benbcai, line 5: Can you elaborate on this param? How is it used or why is it needed?

StephenEsser: expect requires a `received` argument

```
// Expect the result to equal expected.
expect(result).toEqual(expected);

// In this case the unused _ param would be the number 1.
expect(1).toEqual(2);
```

https://jestjs.io/docs/en/expect.html#expectextendmatchers

We run our expect against the entire page, we don't accept a parameter for this value.

```
// The expect takes no arguments, it is always run against the page.
expect().toBeAccessible();
```

https://github.com/cerner/terra-toolkit/blob/8223e5d9de7b3e23030966ccc0b483cf28bee317/packages/terra-functional-testing/src/commands/validates/accessibility.js#L6

The custom toBeAccessible matcher is necessary for reporting a custom message with the assertion so we can print out the violations as part of the assertion failure

https://github.com/cerner/terra-toolkit/blob/8223e5d9de7b3e23030966ccc0b483cf28bee317/packages/terra-functional-testing/src/commands/expect/toBeAccessible.js#L15

packages/terra-functional-testing/src/commands/axe/run.js
```diff
@@ -20,8 +20,19 @@ const runAxe = (overrides = {}) => {
...
+    */
+  const ruleOverrides = {
+    'scrollable-region-focusable': { enabled: false },
+  };
+
   // Merge the global rules and overrides together.
   const rules = {
+    ...ruleOverrides,
     ...axeOptions && axeOptions.rules.reduce((acc, rule) => ({ ...acc, [rule.id]: rule }), {}),
     ...overrides.rules,
```
benbcai, line 37: Looks like our [current implementation](https://github.com/cerner/terra-toolkit-boneyard/blob/385d3210d6045f44246fd9d8503dc0f7d6f003d3/src/wdio/services/TerraCommands/determine-test-options.js#L52) also honors `options.axeRules` if there is no `options.rules`. I can't find what the difference is between these `axeRules` and `rule`. Do we need to check for `axeRules` here?

StephenEsser: I believe that's left over from an earlier assertion implementation.  I'm voting to remove it. We shouldn't have two identical options keys.

I started a list of the breaking changes here I plan to PR after some some adjustments to get the migration guide started:

https://github.com/cerner/terra-toolkit/blob/migration-guide-v1/packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/migration-guide/Migration.tool.mdx

packages/terra-functional-testing/src/commands/expect/toBeAccessible.js
benbcai: I see now. Thanks for the clarification!

benbcai (approved):

PR #456, opened 2020-10-27
[terra-functional-testing] Add Accessibility Reporter

StephenEsser:
### Summary

This PR creates a [Custom WebdriverIO Reporter](https://webdriver.io/docs/customreporter.html) that tracks accessibility warnings and writes them to the console at the end of the test run.

Using the [axe-core API](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#parameters-1) certain rules are set to `reviewOnFail`. This configuration option will mark rules to be returned as `incomplete` upon failure instead of a `violation`.

> reviewOnFail - boolean(option, default false). Override the result of a rule to return "Needs Review" rather than "Violation" if the rule fails.

When running axe if warnings are found the `Terra.validates.accessibility()` command emits a process message the reporter subscribes to for communication. The reporter stores this information and uses it at the end of the test run to generate an accessibility report that logs warnings for any tests that reported them.

```js
const ruleOverrides = {
  'color-contrast': { reviewOnFail: true }
}
```

Example Output:

image

### Additional Details

Warnings do not fail the test run.

This approach only works globally when using `axe.configure`. Warnings cannot be set as options in the Terra.report.accessibility() command which run through `axe.run` which does not support the `reviewOnFail` option.

@cerner/terra


packages/terra-functional-testing/src/commands/axe/run.js
```diff
@@ -43,7 +47,7 @@ const runAxe = (overrides = {}) => {
     axe.run(document, opts, function (error, result) {
       done({ error, result });
     });
-  }, { rules, restoreScroll: true, runOnly: ['wcag2a', 'wcag2aa', 'wcag21aa', 'section508'] });
```
StephenEsser: `restoreScroll` was removed: https://github.com/dequelabs/axe-core/issues/2388

packages/terra-functional-testing/src/commands/expect/toBeAccessible.js
```diff
@@ -8,7 +8,13 @@ const { runAxe } = require('../axe');
  */
 function toBeAccessible(_, options = {}) {
   const { result } = runAxe(options);
-  const { violations } = result;
+  const { incomplete, violations } = result;
+
+  // Rules that fail but are marked for review are returned in the incomplete array.
+  // https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#results-object
+  if (incomplete && incomplete.length > 0) {
+    process.emit('terra:report:accessibility', { incomplete });
```
emilyrohrbough, line 16: Should we also add violations to the report's output? I could see two ways:

- one report, two sections
- two axe reports for incomplete and violations

packages/terra-functional-testing/src/reporters/wdio-accessibility-reporter.js
```diff
@@ -0,0 +1,123 @@
...
+    this.currentTest = test.uid;
+  }
+
+  /**
+   * Hook invoked when accessibility results are reported.
+   * @param {Object} results - Accessibility results.
+   * @param {array} results.incomplete - Incomplete accessibility results.
+   */
+  onReportAccessibility(results) {
+    this.accessibilityResults[this.currentTest] = results;
```
emilyrohrbough, line 41: Is it possible to know the suite info here instead of recursively building it out at the end?

packages/terra-functional-testing/src/reporters/wdio-accessibility-reporter.js
```diff
@@ -0,0 +1,123 @@
...
+  onRunnerEnd(runner) {
+    this.printReport(runner);
+  }
+
+  /**
+   * Print the accessibility report.
+   * @param {RunnerStats} runner - The test runner.
+   */
+  printReport(runner) {
+    const warnings = Object.keys(this.accessibilityResults);
```
emilyrohrbough, line 57: Is there any way to write to the console that an aggregated report can be found in the results dir

packages/terra-functional-testing/src/reporters/wdio-accessibility-reporter.js
```diff
@@ -0,0 +1,123 @@
...
+
+    if (warnings.length === 0) {
+      return;
+    }
+
+    const rootSuite = this.currentSuites.find((suite) => suite.uid === '(root)');
+    const output = this.travelSuite(rootSuite);
+    const spec = runner.specs[0];
+
+    this.write(`Spec: ${spec}\n${output}`);
```
emilyrohrbough, line 67: Can we output this as a md/json report as well that can be pulled in as an build artifact and as a release site artifact?

packages/terra-functional-testing/src/reporters/wdio-accessibility-reporter.js
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  /**
+   * Travels the suite tree to generate the accessibility report.
```
emilyrohrbough, line 71: What do you mean by travel?

packages/terra-functional-testing/tests/fixtures/reporters/test-data-1.json
```diff
@@ -0,0 +1,75 @@
+{
+  "title": "(root)",
```
emilyrohrbough, line 2: Should this be the spec file that was ran?

packages/terra-functional-testing/tests/jest/reporters/wdio-accessibility-reporter.test.js
```diff
@@ -0,0 +1,140 @@
...
+      const mockRunner = { specs: ['/mock/spec/file'] };
+
+      reporter.printReport(mockRunner);
+
+      expect(reporter.write).toHaveBeenCalledWith('Spec: /mock/spec/file\nmock output');
+    });
+  });
+
+  describe('travelSuite', () => {
+    it('should travel the suite tree and generate an accessibility report', () => {
```
emilyrohrbough: The recursiveness of this func, was that for nested suites?

packages/terra-functional-testing/src/reporters/wdio-accessibility-reporter.js
```diff
@@ -0,0 +1,123 @@
...
+  static indent(string, indent) {
+    if (indent > 0) {
+      return string.replace(/^(?!\s*$)/gm, Array(indent + 1).join(' '));
+    }
+
+    return string;
+  }
+
+  constructor(options) {
+    super(options);
```
emilyrohrbough, line 19: Shall we defined the outputDir and logFile option to pass to super ? It seems there is only one option for output files? maybe, which option 2 from this comment: https://github.com/cerner/terra-toolkit/pull/456/files#r513483163

emilyrohrbough:
Quick question:
>To set rules for review they can be set in the Terra Service axe options

I thought we discussed in the sync last week that we no longer wanted allow axe configure axe via the Service and it would be managed at the command level?

packages/terra-functional-testing/src/commands/expect/toBeAccessible.js
StephenEsser: Violations are printed to the console when the expect fails; adding to the reporter also would print them again. We'd want to change how violations are printed if we wanted to pursue this, keeping in mind the expect would fail the test and print a message and then the reporter would print a message

packages/terra-functional-testing/src/reporters/wdio-accessibility-reporter.js
StephenEsser: Probably, but I personally think traveling the tree is actually easier than tracking additional test data manually. A suite can be nested infinitely inside of other suites

packages/terra-functional-testing/src/reporters/wdio-accessibility-reporter.js
StephenEsser: We can write to the console, but we don't currently write any files to the results dir

packages/terra-functional-testing/src/reporters/wdio-accessibility-reporter.js
StephenEsser: Travels though each node in the tree

packages/terra-functional-testing/tests/fixtures/reporters/test-data-1.json
StephenEsser: No, this was generated directly by the test runner and copy pasted into this file. The root suite is always assigned '(root)':

https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-reporter/src/index.ts#L58

packages/terra-functional-testing/tests/jest/reporters/wdio-accessibility-reporter.test.js
StephenEsser: Yes

StephenEsser:
> I thought we discussed in the sync last week that we no longer wanted allow axe configure axe via the Service and it would be managed at the command level?

My understanding was that we didn't want to expose an additional API for consumers to set the axe options via the service when they are extending our wdio.conf which is still the case here. 

The axe API hasn't been removed from the service options, this is just using that existing pass through, this would work today without the changes on this PR.

Do we want to remove the `axe` option entirely from the service?

packages/terra-functional-testing/tests/jest/reporters/wdio-accessibility-reporter.test.js
emilyrohrbough: Can we add a test for this?

StephenEsser: This test is using data generated from nested suites and has 100% jest coverage. Is there something specific you'd like to see explicit test coverage for?

packages/terra-functional-testing/src/reporters/wdio-accessibility-reporter.js
StephenEsser: From the discussion in our WDIO 6 sync I'd say this PR is not going to write out to a file, but it is something we can discuss and circle back on down the road if we think it would be beneficial

packages/terra-functional-testing/src/commands/axe/run.js
```diff
@@ -1,40 +1,51 @@
...
+
 /**
  * Executes axe on the browser.
- * @param {Object} overrides - The axe options.
- * @param {Array} overrides.rules - The rule overrides.
+ * @param {Object} options - The axe options.
+ * @param {Object} options.rules - The rule overrides.
  */
-const runAxe = (overrides = {}) => {
+const runAxe = (options = {}) => {
```
emilyrohrbough: Can we bind this to the Terra service so we don't need to pull these config options from a global?

packages/terra-functional-testing/src/commands/axe/run.js
```diff
@@ -1,40 +1,51 @@
...
+    injectAxe({ ...axeServiceOptions, rules: globalRulesArray });
+  }
 
-  // Merge the global rules and overrides together.
   const rules = {
-    ...ruleOverrides,
-    ...axeOptions && axeOptions.rules,
-    ...overrides.rules,
+    ...GLOBAL_RULE_OVERRIDES,
+    ...axeServiceOptions.rules,
```
emilyrohrbough: I thought we decided to drop this Service Level configuration option in our sync last Wednesday? So no axe configuration when initialized the TerraService or flat on the wdio object, but instead allow per test overrides and then the global be defined like you have in `GLOBAL_RULE_OVERRIDES`.

packages/terra-functional-testing/src/commands/axe/run.js
StephenEsser: I'll experiment with different ways to retrieve the service options in a more direct manner, but I don't think those changes should hold up this PR

packages/terra-functional-testing/src/commands/axe/run.js
StephenEsser: I investigated removing the axe configuration from the service options, but we're going to need it to dynamically flex the color contrast rule while in lowlight theme. Either to disable it or adjust the contrast ratio.

Here's how we dynamically flex the rules today:

https://github.com/cerner/terra-toolkit-boneyard/blob/main/config/wdio/wdio.conf.js#L102

Here's an example of what removing the axe option looks like:

https://github.com/cerner/terra-toolkit/compare/remove-terra-service-axe-option

We might be able to flex the option within the service, but until we have a solution we'll need to keep it as a service option. These changes are just cleaning up the file, they're not introducing any new functionality

emilyrohrbough: I assumed we'd be flex that in the service like how we are overriding the scroll rule. It's always be disabled (or warn) for the lowlight theme.

packages/terra-functional-testing/tests/jest/reporters/wdio-accessibility-reporter.test.js
emilyrohrbough: it's not obvious from that was what being tested here by reading through the tests. Maybe update the description to indicate that then.

packages/terra-functional-testing/src/commands/axe/run.js
StephenEsser: The scroll rule is static and never changes, but the color contrast rule would be dynamic based on the theme.

Removing the axe option from the service is outside the scope of this PR but I'll do additional investigation into removing the axe options while still dynamically flexing rules based on the theme. If I come up with a solution I'll make a separate PR. Removing the axe options from the service doesn't impact the intended functionality of the accessibility reporter

packages/terra-functional-testing/src/commands/axe/run.js
benbcai: I didn't know the scrolling that it used to do was _only_ for color-contrast.

packages/terra-functional-testing/src/commands/axe/run.js
benbcai: The viewport helpers and validate elements also need to look up the service options. Making them classes and initialize with the options was not the easiest to do so I set the service options in the [Terra global object](https://github.com/cerner/terra-toolkit/pull/464/files#diff-a6ba3c1d16c08e58e7ee3994f3ab541b63b2a3a8dafb0de9daed2c8090c948d9R41). That's what I going with for now unless we have other better ways to do it.

packages/terra-functional-testing/src/reporters/wdio-accessibility-reporter.js
```diff
@@ -0,0 +1,123 @@
...
+      return `${AccessibilityReporter.indent(suiteTitle, (depth * 2))}\n${warnings.join('\n')}\n`;
+    }
+
+    return '';
+  }
+
+  /**
+   * Formats an accessibility warning summary for a test.
+   * @param {string} uid - The unique identifier of the test.
+   * @param {number} indent - The indentation the report should have,
```
benbcai: Did you intend to have more comment at the end or should it be a period?

packages/terra-functional-testing/src/reporters/wdio-accessibility-reporter.js
```diff
@@ -0,0 +1,123 @@
...
+
+      if (output.length > 0) {
+        warnings.push(output);
+      }
+    });
+
+    tests.forEach((test) => {
+      const { uid } = test;
+
+      if (this.accessibilityResults[uid]) {
```
benbcai, line 92: Do we need to ensure that this result specifically has the `incomplete` type or is it guaranteed that the type is `incomplete` by the time it gets here?

StephenEsser: It's guaranteed, with the current implementation the results are only stored if they contain an incomplete array

packages/terra-functional-testing/tests/jest/commands/axe/inject.test.js
```diff
@@ -15,14 +15,14 @@ describe('Inject Axe', () => {
 
   it('should inject axe into the document with the provided options', () => {
     const mockExecute = jest.fn();
-    const mockRules = { rules: [{ id: 'mock-rule', enabled: true }] };
+    const mockRules = { rules: [{ enabled: true, id: 'mock-rule' }] };
```
benbcai: Curious what the reason is for swapping these? It's more natural to me to see the rule name first followed by their attributes.

benbcai (approved):

packages/terra-functional-testing/src/commands/axe/run.js
StephenEsser: I've created a PR to address these changes, I'd like to move axe configuration discussions onto that PR:

https://github.com/cerner/terra-toolkit/pull/466

packages/terra-functional-testing/src/commands/axe/run.js
StephenEsser: I've created a PR to address these changes, I'd like to move axe configuration discussions onto that PR:

https://github.com/cerner/terra-toolkit/pull/466

I've removed all changes not directly associated to the accessibility reporter from this PR

packages/terra-functional-testing/src/reporters/wdio-accessibility-reporter.js
StephenEsser: This should be a period, updated:

https://github.com/cerner/terra-toolkit/pull/456/commits/51b3981ed4debda410ada81996f7a3cde1566c98

packages/terra-functional-testing/tests/jest/commands/axe/inject.test.js
StephenEsser: This is the order the objects will be defined for the real axe inject due to how the map generates the rules array and the order matters for the strict string comparison below. The id is always appended. For this test it doesn't matter, I've reverted the changes to how it was before

packages/terra-functional-testing/src/commands/expect/toBeAccessible.js
mjhenkes: I don't think we need to add violations to the report. The failed test run should be indication enough. I see the report as valuable on non-failed test runs.

mjhenkes (approved):

PR #460, opened 2020-10-28
terra-cli initial release

### LINKED ISSUE #457: Create a CLI package to support running a variety of terra commands
# Feature Request

### Overview

We would like to create a terra CLI, to allow various terra related commands to be run via one base command. The [lerna command implementation](https://github.com/lerna/lerna) is a basis for what we want to implement. The structure of the commands will be:

```
terra [command] <command parameters>
```

Some of the requirements we would like to shoot for with this package include:

1. The ability to have commands defined across multiple packages
2. The ability to only have the commands exposed that are installed in a given project

### Implementation Specifics

To accomplish requirement 1 above, we need to have a mechanism to define a command as a part of a given package. [yargs](https://github.com/yargs/yargs) has a [convenient way](https://github.com/yargs/yargs/blob/master/docs/advanced.md#providing-a-command-module) to define commands via an object that can be exported from a given module.

We can then satisfy number 1 and 2 by having a convention where we place command modules in paths of the form `lib/terra-cli/<command>/index.js`. If that's the case, we can set up terra-cli to search through dependencies of a given project for that structure. To cover all cases, we will search the following for a list of projects that we control in an allow list:

* Direct dependencies of the project
* Direct dev dependencies of the project
* The current project itself
* To handle the special case of terra-toolkit itself, we will search through the packages in the packages directory as well, but only if we are currently in the terra-toolkit project.

ryanthemanuel:
### Summary

This is the initial implementation of the terra-cli package. See the issue for the design for this project.

Fixes #457


packages/terra-open-source-scripts/src/terra-cli/git-tag-mono-repo/gitTagMonoRepoHandler.js
```diff
@@ -0,0 +1,50 @@
+const fs = require('fs');
```
ryanthemanuel: I'd like to flex these scripts automatically based on whether the current repo is a monorepo

emilyrohrbough:
Do you have a high level tech design/summary on how additional commands will hook in?

ryanthemanuel:
> Do you have a high level tech design/summary on how additional commands will hook in?

@emilyrohrbough It's somewhat covered on the issue. Is there something I can add to clarify things? The issue describes how you would define your command in a directory of the form `/lib/terra-cli/<command>/index.js` and links out to the yargs documentation on what that command would need to look like, namely:

* exports.command: string (or array of strings) that executes this command when given on the command line, first string may contain positional args
* exports.aliases: array of strings (or a single string) representing aliases of exports.command, positional args defined in an alias are ignored
* exports.describe: string used as the description for the command in help text, use false for a hidden command
* exports.builder: object declaring the options the command accepts, or a function accepting and returning a yargs instance
* exports.handler: a function which will be passed the parsed argv.
* exports.deprecated: a boolean (or string) to show deprecation notice.

package.json
```diff
@@ -72,12 +75,13 @@
     "mocha": "cd packages/terra-functional-testing && npm run mocha",
     "pack": "webpack -p",
     "postinstall": "link-parent-bin && npm run compile && npm run bootstrap:hoist",
+    "prepare-for-release": "npm run terra -- prepare-for-release",
```
mjhenkes, line 75: In the other repos would we use this syntax?

```
    "prepare-for-release": "terra prepare-for-release",
```

ryanthemanuel: Yes.  This is just to get around mono repo weirdness.

.travis.yml
```diff
@@ -37,10 +37,8 @@ jobs:
         - rm -rf ./travis-build
         # add auth token to .npmrc for lerna publish
         - echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" >> $HOME/.npmrc 2> /dev/null
```
mjhenkes: Theoretically you could also add this to the release script.

packages/terra-open-source-scripts/src/terra-cli/release/releaseMonoRepoHandler.js
```diff
@@ -0,0 +1,36 @@
...
+  if (lastIndexOf < 0) {
+    return null;
+  }
+  const substring = output.substring(lastIndexOf + successString.length);
+  const withoutDashes = substring.replace(/ - /g, '').trim();
+  return withoutDashes.split('\n');
+};
+
+module.exports = async () => {
+  const output = await exec(`npx lerna publish from-package --yes > ${filePath}`);
```
mjhenkes: do you still need to write this out to the filepath?

packages/terra-open-source-scripts/src/terra-cli/release/releaseRepoHandler.js
```diff
@@ -0,0 +1,39 @@
...
+  const packageFile = path.resolve(process.cwd(), 'package.json');
+
+  // Read package.json and pull out version
+  const { name, version } = await fs.readJSON(packageFile);
+
+  await setupGit(name, version);
+
+  const tag = `v${version}`;
+
+  await exec('npm publish');
```
mjhenkes, line 47: can we verify successful publish before tagging? or will the script blow up if this command fails?

mjhenkes:
This PR looks really good to me. I'd propose we get it merged sooner rather than later and get some sample usage. Even if that means docs are added in a subsequent pull request.

packages/terra-open-source-scripts/src/terra-cli/release/releaseMonoRepoHandler.js
ryanthemanuel: oops forgot to remove that.

packages/terra-open-source-scripts/src/terra-cli/release/releaseRepoHandler.js
ryanthemanuel: i believe it'll blow up if it fails, but i'll verify.

packages/terra-open-source-scripts/src/terra-cli/git-tag-mono-repo/gitTagMonoRepoHandler.js
ryanthemanuel: This is already being done?

.travis.yml
ryanthemanuel: @mjhenkes I added it.

packages/terra-open-source-scripts/src/terra-cli/release/releaseMonoRepoHandler.js
ryanthemanuel: I have removed it.

packages/terra-open-source-scripts/src/terra-cli/release/releaseMonoRepoHandler.js
```diff
@@ -0,0 +1,37 @@
...
+  if (lastIndexOf < 0) {
+    return null;
+  }
+  const substring = output.substring(lastIndexOf + successString.length);
+  const withoutDashes = substring.replace(/ - /g, '').trim();
+  return withoutDashes.split('\n');
+};
+
+module.exports = async () => {
+  await fs.writeFile(path.join(process.env.HOME, '.npmrc'), `//registry.npmjs.org/:_authToken=${process.env.NPM_TOKEN}`, 'utf-8');
```
mjhenkes: consider commenting this file up to describe what you're doing and why for any future travelers.

package.json
```diff
@@ -32,6 +32,9 @@
   "bugs": {
     "url": "https://github.com/cerner/terra-toolkit/issues"
   },
+  "bin": {
+    "terra": "./packages/terra-cli/bin/terra"
+  },
```
emilyrohrbough: Is this to pull forward the bin from the terra-ci package?

packages/terra-cli/index.js
```diff
@@ -0,0 +1,78 @@
...
+};
+
+const subDirectoriesOfDirectory = (directoryPath) => (
+  fs.readdirSync(directoryPath, { withFileTypes: true })
+    .filter(directoryEntry => directoryEntry.isDirectory())
+    .map(directoryEntry => path.join(directoryPath, directoryEntry.name))
+);
+
+const terraClIDirectoriesFromPaths = (paths, preferSrc) => (
+  paths.map(dependencyPath => path.join(dependencyPath, preferSrc ? 'src' : 'lib', 'terra-cli')).filter(directory => fs.pathExistsSync(directory))
```
emilyrohrbough: why support both src and lib?

packages/terra-cli/index.js
```diff
@@ -0,0 +1,78 @@
...
+    .wrap(cli.terminalWidth());
+};
+
+const subDirectoriesOfDirectory = (directoryPath) => (
+  fs.readdirSync(directoryPath, { withFileTypes: true })
+    .filter(directoryEntry => directoryEntry.isDirectory())
+    .map(directoryEntry => path.join(directoryPath, directoryEntry.name))
+);
+
+const terraClIDirectoriesFromPaths = (paths, preferSrc) => (
```
emilyrohrbough: Can we add in-line docs on these ?

packages/terra-cli/index.js
```diff
@@ -0,0 +1,78 @@
...
+  paths.map(dependencyPath => path.join(dependencyPath, preferSrc ? 'src' : 'lib', 'terra-cli')).filter(directory => fs.pathExistsSync(directory))
+);
+
+const loadTerraCommands = () => {
+  const { dependencies = [], devDependencies = [], name } = fs.readJSONSync(path.join(process.cwd(), 'package.json'));
+  const firstLevelDependencies = [...Object.keys(dependencies), ...Object.keys(devDependencies)];
+  const firstLevelDependencyPaths = firstLevelDependencies.map(directory => path.join(process.cwd(), 'node_modules', directory));
+  const additionalDependencyPaths = (name === 'terra-toolkit' ? subDirectoriesOfDirectory(path.join(process.cwd(), 'packages')) : []);
```
emilyrohrbough:
```suggestion
  const terraToolkitPackagesPaths = (name === 'terra-toolkit' ? subDirectoriesOfDirectory(path.join(process.cwd(), 'packages')) : []);
```

packages/terra-cli/index.js
```diff
@@ -0,0 +1,78 @@
...
+      // exit non-zero so the CLI can be usefully chained
+      cli.exit(actual.code > 0 ? actual.code : 1, actual);
+    })
+    .alias('h', 'help')
+    .alias('v', 'version')
+    .wrap(cli.terminalWidth());
+};
+
+const subDirectoriesOfDirectory = (directoryPath) => (
+  fs.readdirSync(directoryPath, { withFileTypes: true })
```
emilyrohrbough: Does this need to be sync?

packages/terra-cli/index.js
```diff
@@ -0,0 +1,78 @@
...
+    ...terraClIDirectoriesFromPaths(firstLevelDependencyPaths, false),
+    // Prefer src from local dependencies
+    ...terraClIDirectoriesFromPaths([...additionalDependencyPaths, process.cwd()], true),
+  ];
+  return terraCLIDirectories.map((directory) => subDirectoriesOfDirectory(directory)).reduce((previousOutput, value) => previousOutput.concat(value), []);
+};
+
+const terraCLI = (argv) => {
+  let cli = setupCLI();
+  const commands = loadTerraCommands();
```
emilyrohrbough: These are file paths right?

packages/terra-cli/index.js
```diff
@@ -0,0 +1,78 @@
...
+);
+
+const terraClIDirectoriesFromPaths = (paths, preferSrc) => (
+  paths.map(dependencyPath => path.join(dependencyPath, preferSrc ? 'src' : 'lib', 'terra-cli')).filter(directory => fs.pathExistsSync(directory))
+);
+
+const loadTerraCommands = () => {
+  const { dependencies = [], devDependencies = [], name } = fs.readJSONSync(path.join(process.cwd(), 'package.json'));
+  const firstLevelDependencies = [...Object.keys(dependencies), ...Object.keys(devDependencies)];
+  const firstLevelDependencyPaths = firstLevelDependencies.map(directory => path.join(process.cwd(), 'node_modules', directory));
```
emilyrohrbough: This will allow anyone to hook in for a command correct, if they add a terra-cli directory to their project?

packages/terra-functional-testing/package.json
```diff
@@ -14,7 +14,7 @@
     "*.md"
   ],
   "scripts": {
-    "compile": "babel src --out-dir lib --copy-files",
+    "compile": "babel src --out-dir lib --copy-files --ignore src/terra-cli",
```
emilyrohrbough, line 17: Won't this need to be in lib once terra-functional-testing is released?

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
```diff
@@ -29,19 +32,21 @@ class TestRunner {
...
 
-      const exitCode = await testRunner.run();
-
-      process.exit(exitCode);
+      exitCode = await testRunner.run();
     } catch (error) {
-      // eslint-disable-next-line no-console
-      console.error('[terra-functional-testing] Launcher failed to start the test.\n', error);
-      process.exit(1);
+      logger.error('Launcher failed to start the test');
```
emilyrohrbough, line 44: Should this set the exit code to 1 still?

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
```diff
@@ -29,19 +32,21 @@ class TestRunner {
...
+      throw error;
+    }
+
+    if (exitCode !== 0) {
+      throw new Error(`Launcher returned with an exit code of ${exitCode}`);
```
emilyrohrbough: Should this be more descriptive, for example: terra-functional-testing cli or something along those lines?

packages/terra-open-source-scripts/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@cerner/terra-open-source-scripts",
+  "private": true,
+  "version": "0.0.0",
+  "description": "Terra package with various scripts for open source development",
+  "main": "index.js",
+  "bin": {
+    "terra": "./bin/terra"
+  },
```
emilyrohrbough: I wouldn't have thought we needed to provide a bin if the cli is loading these commands.

packages/terra-open-source-scripts/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-toolkit/issues"
+  },
+  "engines": {
+    "node": "^10.13.0"
+  },
+  "homepage": "https://github.com/cerner/terra-toolkit",
+  "scripts": {
+    "compile": "babel src --out-dir lib --copy-files --ignore src/terra-cli",
```
emilyrohrbough, line 29: Same question as before, won't this need to be compiled to lib once this is released and consumed else where?

packages/terra-open-source-scripts/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "lint": "eslint --ext .js,.jsx ."
+  },
+  "dependencies": {
+    "@cerner/terra-cli": "^0.0.0",
```
emilyrohrbough: Shouldn't this be a dev-dependency and peer-dependency?

I expected "@cerner/terra-cli" to be added directly by the project want to use it's commands

packages/terra-open-source-scripts/src/terra-cli/express-server/index.js
```diff
@@ -0,0 +1,30 @@
...
+      host: {
+        type: 'string',
+        default: '0.0.0.0',
+        defaultDescription: '0.0.0.0',
+        describe: 'The host on which to host the server',
+      },
+      site: {
+        type: 'string',
+        demandOption: true,
+        describe: 'The site from which to host the server',
```
emilyrohrbough:
```suggestion
        describe: 'The directory path from which to serve the static assets',
```

packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/prepareForReleaseRepoHandler.js
```diff
@@ -0,0 +1,30 @@
...
+const log = require('npmlog');
+const childProcess = require('child_process');
+const { promisify } = require('util');
+
+const updateChangelogForPackage = require('./updateChangelogForPackage');
+
+const exec = promisify(childProcess.exec);
+
+const promptForVersionType = async () => {
+  log.pause();
```
emilyrohrbough, line 11: Why is this pause needed?

packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/prepareForReleaseMonoRepoHandler.js
```diff
@@ -0,0 +1,33 @@
...
+const stripAnsi = require('strip-ansi');
+const Logger = require('@cerner/terra-cli/lib/utils/Logger');
+
+const logger = new Logger({ prefix: '[terra-open-source-scripts:version]' });
+
+const exec = promisify(childProcess.exec);
+
+const updateChangelogForPackage = require('./updateChangelogForPackage');
+
+const VERSION_OUTPUT_PATH = path.join(process.cwd(), 'version-output.txt');
```
emilyrohrbough:
```suggestion
const VERSION_OUTPUT_PATH = path.join(process.cwd(), 'tmp', 'version-output.txt');
```

packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/prepareForReleaseMonoRepoHandler.js
```diff
@@ -0,0 +1,33 @@
...
+
+module.exports = async () => {
+  try {
+    childProcess.execSync(`script -q ${VERSION_OUTPUT_PATH} lerna version --no-git-tag-version`, { stdio: 'inherit' });
```
emilyrohrbough, line 23: Feels like lerna should be a dependency if we are going to use it here

packages/terra-open-source-scripts/src/terra-cli/release/index.js
```diff
@@ -0,0 +1,11 @@
+const releaseMonoRepoHandler = require('./releaseMonoRepoHandler');
+const releaseRepoHandler = require('./releaseRepoHandler');
+const isMonoRepo = require('../../utils/isMonoRepo');
+
+const release = {
+  command: 'release',
+  describe: 'Tag the repo and release',
+  handler: async () => (await isMonoRepo() ? releaseMonoRepoHandler() : releaseRepoHandler()),
```
emilyrohrbough, line 8: IMO we should require the one running the command to pass a flag if it's a mono repo project instead to trying to check if a lerna.json file exists

packages/terra-open-source-scripts/src/terra-cli/release/releaseMonoRepoHandler.js
```diff
@@ -0,0 +1,37 @@
+const childProcess = require('child_process');
```
emilyrohrbough, line 1: nit: rename file to `release-monorepo` and rename the non-mono-release file to `release`

packages/terra-open-source-scripts/src/terra-cli/release/releaseRepoHandler.js
```diff
@@ -0,0 +1,41 @@
...
+const { promisify } = require('util');
+const Logger = require('@cerner/terra-cli/lib/utils/Logger');
+
+const setupGit = require('./setupGit');
+
+const logger = new Logger({ prefix: '[terra-open-source-scripts:release]' });
+
+const exec = promisify(childProcess.exec);
+
+const isPublished = async (name, version) => {
```
emilyrohrbough:
```suggestion
const checkIfPublished = async (name, version) => {
```

emilyrohrbough: This should be pulled forward into Jenkins.

packages/terra-open-source-scripts/src/terra-cli/release/releaseRepoHandler.js
```diff
@@ -0,0 +1,41 @@
...
+  const pkgJson = await pacote.packument(name, {
+    registry: 'https://registry.npmjs.org/',
+  });
+  const publishedVersions = Object.keys(pkgJson.versions);
+  return publishedVersions.includes(version);
+};
+
+module.exports = async () => {
+  if (await isPublished()) {
```
emilyrohrbough: Do you need to pass the name and version to isPublished?

packages/terra-open-source-scripts/src/terra-cli/release/releaseRepoHandler.js
```diff
@@ -0,0 +1,41 @@
...
+    return;
+  }
+  await fs.writeFile(path.join(process.env.HOME, '.npmrc'), `//registry.npmjs.org/:_authToken=${process.env.NPM_TOKEN}`, 'utf-8');
+
+  const packageFile = path.resolve(process.cwd(), 'package.json');
+
+  // Read package.json and pull out version
+  const { name, version } = await fs.readJSON(packageFile);
+
+  await setupGit(name, version);
```
emilyrohrbough, line 50: Can we make the order of publish, setupGit, tagging in git and pushing tags consistent between mono-repo and non-mono-repo packages?

packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/index.js
```diff
@@ -0,0 +1,11 @@
+const prepareForReleaseMonoRepoHandler = require('./prepareForReleaseMonoRepoHandler');
+const prepareForReleaseRepoHandler = require('./prepareForReleaseRepoHandler');
+const isMonoRepo = require('../../utils/isMonoRepo');
+
+const prepareForRelease = {
+  command: 'prepare-for-release',
+  describe: 'Prepare the project for release by updating the version and changelog of the project(s)',
+  handler: async () => (await isMonoRepo() ? prepareForReleaseMonoRepoHandler() : prepareForReleaseRepoHandler()),
```
emilyrohrbough, line 8: Same comment as below about using a flag to flip between mono-repo and non-mono-repo commands.

packages/terra-open-source-scripts/src/terra-cli/release/index.js
```diff
@@ -0,0 +1,11 @@
+const releaseMonoRepoHandler = require('./releaseMonoRepoHandler');
+const releaseRepoHandler = require('./releaseRepoHandler');
+const isMonoRepo = require('../../utils/isMonoRepo');
+
+const release = {
+  command: 'release',
+  describe: 'Tag the repo and release',
```
emilyrohrbough: Can we be a little more descriptive?

packages/terra-cli/index.js
```diff
@@ -0,0 +1,78 @@
...
+  const additionalDependencyPaths = (name === 'terra-toolkit' ? subDirectoriesOfDirectory(path.join(process.cwd(), 'packages')) : []);
+  const terraCLIDirectories = [
+    // Prefer lib from installed dependencies
+    ...terraClIDirectoriesFromPaths(firstLevelDependencyPaths, false),
+    // Prefer src from local dependencies
+    ...terraClIDirectoriesFromPaths([...additionalDependencyPaths, process.cwd()], true),
+  ];
+  return terraCLIDirectories.map((directory) => subDirectoriesOfDirectory(directory)).reduce((previousOutput, value) => previousOutput.concat(value), []);
```
emilyrohrbough: Any way for this to async?

packages/terra-cli/index.js
ryanthemanuel: This is to help out local development so you don't have to remember to compile to use the command.

packages/terra-open-source-scripts/package.json
ryanthemanuel: This was a copy paste err

packages/terra-cli/index.js
mjhenkes: We should impose tight control over what commands can be added to terra-cli. Any command in terra cli should be supported and maintained by the terra team.

packages/terra-functional-testing/package.json
mjhenkes: it still gets moved, just not transpiled.

packages/terra-open-source-scripts/package.json
ryanthemanuel: --copy-files and --ignore the way that is set up will copy the files over but not try to compile the ones that are ignored.

packages/terra-open-source-scripts/src/terra-cli/release/releaseRepoHandler.js
ryanthemanuel: I'm not sure I know what you mean here.

mjhenkes: what do you mean by pulling this forward into jenkins?

packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/index.js
ryanthemanuel: @mjhenkes I thought I remember asking you about this and you leaned this way.  Do you have a strong preference?

packages/terra-open-source-scripts/src/terra-cli/release/index.js
mjhenkes: edit: oops responded to the wrong comment

~If you're a mono repo you already have lerna, if your not this won't get called. I'm ok with this being loosely coupled.~

packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/prepareForReleaseMonoRepoHandler.js
mjhenkes: If you're a mono repo you already have lerna, if your not this won't get called. I'm ok with this being loosely coupled.

packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/prepareForReleaseMonoRepoHandler.js
ryanthemanuel: I'm fine doing this, but we delete the file so it shouldn't necessarily be hanging around or anything.

packages/terra-open-source-scripts/src/terra-cli/release/index.js
mjhenkes: I'd rather keep it simple and detect if we're in a mono repo. I use this same logic in terra dev site to determine which files to pull from.

packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/index.js
mjhenkes: I'd rather keep it simple and detect if we're in a mono repo. I use this same logic in terra dev site to determine which files to pull from.

packages/terra-open-source-scripts/package.json
ryanthemanuel: I need it for the cli logging so it definitely needs to be some form of dependency.  I'm not sure peer needs to happen.  What are your concerns with making it a direct dependency?

packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/prepareForReleaseMonoRepoHandler.js
mjhenkes: Do we have to write this file to disk? does the script output not get returned by the execSync call?

packages/terra-open-source-scripts/package.json
ryanthemanuel: For posterity, the way you would both ignore *and* not copy a js file is by using the --no-copy-ignored option.

packages/terra-open-source-scripts/package.json
mjhenkes: I think i agree with making it a peer dependency. I think of terra-open-source-scripts as a plugin to the terra-cli package. So the user 'opts' into terra-cli then adds the plugins to add scripts.

package.json
ryanthemanuel: This is something I saw lerna do in their top level.  But I don't think it actually does anything from what I can tell.  I can remove it.

packages/terra-cli/index.js
ryanthemanuel: I'm fine with that.  I misunderstood things coming out of the war room.  I had it documented as above on the design, but updated it to be (additions in bold):

To cover all cases, we will search the following **for a list of projects that we control in an allow list**

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
ryanthemanuel: I can make it more descriptive.  There will be a stack trace if this happens though.

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
ryanthemanuel: From what I've read, using process.exit is strongly discouraged as it immediately kills the process.  Sometimes things get buffered like stdout or other asynchronous tasks and if you process.exit it will end them prematurely. Throwing the error will pass up to the command runner and ultimately end the process with a non success status, so I think this should work unless there was a reason we were hard exiting I'm not aware of.

packages/terra-open-source-scripts/package.json
ryanthemanuel: Fair enough and I don't have a strong preference necessarily

packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/prepareForReleaseMonoRepoHandler.js
ryanthemanuel: It does not return here. I will document this, but this particular functionality was somewhat difficult to implement.  

The only way I could properly get interactive behavior with this command was to use stdio 'inherit'. Once you do that it no longer returns the output from the command. Even doing something like `> <file>` doesn't really work the way that it needs to. `script` seemed to be the only thing that got me what I needed.

packages/terra-open-source-scripts/src/terra-cli/release/releaseRepoHandler.js
ryanthemanuel: Yep.  I missed that when refactoring.  Good catch.

packages/terra-open-source-scripts/src/terra-cli/release/releaseRepoHandler.js
ryanthemanuel: Since this returns a boolean, I'd rather keep this as isPublished.  Maybe I'll change it to `isAlreadyPublished` though?  Thoughts?

packages/terra-cli/index.js
ryanthemanuel: I found a way.  It's a little more awkward but I like being consistent:

https://github.com/cerner/terra-toolkit/pull/460/commits/6c35f4191a2115b9652c22dc9f42105554d0b08c

packages/terra-open-source-scripts/src/terra-cli/release/index.js
ryanthemanuel: Updated: https://github.com/cerner/terra-toolkit/pull/460/commits/6c35f4191a2115b9652c22dc9f42105554d0b08c

packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/index.js
ryanthemanuel: Keeping it for now unless @emilyrohrbough or others feel strongly. If so, we can discuss

packages/terra-open-source-scripts/src/terra-cli/release/releaseRepoHandler.js
ryanthemanuel: Done. https://github.com/cerner/terra-toolkit/pull/460/commits/6c35f4191a2115b9652c22dc9f42105554d0b08c

packages/terra-open-source-scripts/src/terra-cli/release/releaseRepoHandler.js
ryanthemanuel: https://github.com/cerner/terra-toolkit/pull/460/commits/6c35f4191a2115b9652c22dc9f42105554d0b08c

packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/prepareForReleaseMonoRepoHandler.js
ryanthemanuel: https://github.com/cerner/terra-toolkit/pull/460/commits/6c35f4191a2115b9652c22dc9f42105554d0b08c

packages/terra-open-source-scripts/src/terra-cli/express-server/index.js
ryanthemanuel: https://github.com/cerner/terra-toolkit/pull/460/commits/6c35f4191a2115b9652c22dc9f42105554d0b08c

packages/terra-open-source-scripts/package.json
ryanthemanuel: https://github.com/cerner/terra-toolkit/pull/460/commits/6c35f4191a2115b9652c22dc9f42105554d0b08c

packages/terra-open-source-scripts/package.json
ryanthemanuel: Fixed: https://github.com/cerner/terra-toolkit/pull/460/commits/6c35f4191a2115b9652c22dc9f42105554d0b08c

packages/terra-cli/index.js
ryanthemanuel: Updated to behave as mentioned: https://github.com/cerner/terra-toolkit/pull/460/commits/6c35f4191a2115b9652c22dc9f42105554d0b08c

packages/terra-cli/index.js
ryanthemanuel: Updated the name to be clearer.

packages/terra-cli/index.js
ryanthemanuel: Updated to be async: https://github.com/cerner/terra-toolkit/pull/460/commits/6c35f4191a2115b9652c22dc9f42105554d0b08c

packages/terra-cli/index.js
ryanthemanuel: Good call: https://github.com/cerner/terra-toolkit/pull/460/commits/6c35f4191a2115b9652c22dc9f42105554d0b08c

package.json
ryanthemanuel: Removed: https://github.com/cerner/terra-toolkit/pull/460/commits/6c35f4191a2115b9652c22dc9f42105554d0b08c

ryanthemanuel:
I went through the functional changes in the PR just now.  I'll add some doc tomorrow.

packages/terra-open-source-scripts/src/terra-cli/release/releaseRepoHandler.js
ryanthemanuel: I verified that an exception is thrown if this exec fails.

packages/terra-cli/index.js
emilyrohrbough: IMO we should setup local dev to run compile before running tests so it's not a problem so the CLI can be simplified

packages/terra-cli/index.js
```diff
@@ -35,37 +37,39 @@ const setupCLI = () => {
...
     // Prefer lib from installed dependencies
-    ...terraClIDirectoriesFromPaths(firstLevelDependencyPaths, false),
-    // Prefer src from local dependencies
-    ...terraClIDirectoriesFromPaths([...additionalDependencyPaths, process.cwd()], true),
+    ...(await terraClIDirectoriesFromPaths(firstLevelDependencyPaths, false)),
+    // Prefer src from local mono repo packages and the current project
+    ...(await terraClIDirectoriesFromPaths([...toolkitPackagePaths, ...currentProjectPaths], true)),
   ];
-  return terraCLIDirectories.map((directory) => subDirectoriesOfDirectory(directory)).reduce((previousOutput, value) => previousOutput.concat(value), []);
+  return (await Promise.all(terraCLIDirectories.map((directory) => subDirectoriesOfDirectory(directory)))).reduce((previousOutput, value) => previousOutput.concat(value), []);
```
emilyrohrbough: totally random, but is this file getting linted? I think the linter typically throws for

```(directory) => subDirectoriesOfDirectory(directory))``` 
since you aren't using braces it'd want:
```directory => subDirectoriesOfDirectory(directory))```

**Review by emilyrohrbough — COMMENTED** [review]
Looks good. I'm still not a fan of toggling src and lib but seems to work!

Can you run the pack script to pull this into terra-core (or equivalent) and terra-aggregated-themes to check that things are working as expected?

File: packages/terra-cli/index.js [hunk 33]
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * 1. Dependencies in the top node_modules directory that are in the dependencies allow list
+ * 2. Monorepo package directories if the current project is terra-toolkit and the directory is in the dependency allow list
+ * 3. The current directory if its in the dependencies allow list
+ * @returns {Array} an array of all qualifying terra-cli command paths
+ */
+const loadTerraCommandPaths = async () => {
+  const { name } = fs.readJSONSync(path.join(process.cwd(), 'package.json'));
+  // Search the top level node_modules for dependencies in the allow list
+  const firstLevelDependencyPaths = DEPENDENCY_ALLOW_LIST.map(dependency => path.join(process.cwd(), 'node_modules', dependency));
```
mjhenkes: do you have to take the `@cerner` scoping into account here? Both for the package name well as the path?

ryanthemanuel: I have updated this to fix this.  I use `@cerner` but then drop it when looking inside of the terra toolkit monorepo

mjhenkes: That covers the name from the package, but in node modules `@cerner` scoped packages will be under the `@cerner` directory. You'll have to split the name off the dependency.... or cheat and only search the `@cerner` directory.

```suggestion
  const firstLevelDependencyPaths = DEPENDENCY_ALLOW_LIST.map(dependency.replace('@cerner/', '') => path.join(process.cwd(), 'node_modules', '@cerner', dependency));
```

mjhenkes: lol, never mind, this is already covered

mjhenkes (approved):

ryanthemanuel:
> Looks good. I'm still not a fan of toggling src and lib but seems to work!
> 
> Can you run the pack script to pull this into terra-core (or equivalent) and terra-aggregated-themes to check that things are working as expected?

Looks like that rule got turned off a year ago: https://github.com/cerner/eslint-config-terra/commit/23252cd1de1342f5b63bf16561098512e5b91426

I did verify that file is being linted though.

packages/terra-cli/index.js
ryanthemanuel: Looks like that rule got turned off a year ago: cerner/eslint-config-terra@23252cd

I did verify that file is being linted though.

packages/terra-cli/index.js
```diff
@@ -0,0 +1,119 @@
...
+const setupCLI = () => {
+  const cli = yargs();
+
+  return cli.usage('Usage: $0 <command> [options]')
+    .demandCommand(1, 'A command is required. Pass --help to see all available commands and options.')
+    .recommendCommands()
+    .strict()
+    .fail((msg, err) => {
+      // certain yargs validations throw strings
+      const actual = err || new Error(msg);
```
yuderekyu: nitpick, can we rename this to something more specific, like `actualErr`

packages/terra-cli/index.js
```diff
@@ -0,0 +1,119 @@
...
+    .map(directoryEntry => path.join(directoryPath, directoryEntry.name));
+};
+
+/**
+ * Takes an array of paths, appends either src or lib depending on preferSrc, appends terra-cli, and then returns an array of the the resulting directory paths that exist
+ * @param {Array} paths - the array of paths to act upon
+ * @param {boolean} preferSrc - indicates whether or not src should be used instead of lib. This should be used in cases where the package being searched is local
+ * @returns {Array} an array of paths resolved out to the terra-cli directory
+ */
+const terraClIDirectoriesFromPaths = async (paths, preferSrc) => {
```
yuderekyu: can we update this to an object param? it'll clear up what each param is for, especially the boolean
```suggestion
const terraClIDirectoriesFromPaths = async ({ paths, preferSrc }) => {
```

packages/terra-cli/index.js
```diff
@@ -0,0 +1,119 @@
...
+  const { name } = fs.readJSONSync(path.join(process.cwd(), 'package.json'));
+  // Search the top level node_modules for dependencies in the allow list
+  const firstLevelDependencyPaths = DEPENDENCY_ALLOW_LIST.map(dependency => path.join(process.cwd(), 'node_modules', ...dependency.split('/')));
+  // Only include mono repo packages if it's terra-toolkit and they're in the allow list and drop the @cerner since that's not included in the package directory
+  const toolkitPackagePaths = (name === 'terra-toolkit' ? DEPENDENCY_ALLOW_LIST.map(dependency => dependency.replace('@cerner/', '')).map(dependency => path.join(process.cwd(), 'packages', dependency)) : []);
+  // Only include the current directory if it's in the allow list
+  const currentProjectPaths = (DEPENDENCY_ALLOW_LIST.includes(name) ? [process.cwd()] : []);
+  const terraCLIDirectories = [
+    // Prefer lib from installed dependencies
+    ...(await terraClIDirectoriesFromPaths(firstLevelDependencyPaths, false)),
```
yuderekyu:
```suggestion
    ...(await terraClIDirectoriesFromPaths({ paths: firstLevelDependencyPaths, preferSrc: false })),
```

packages/terra-cli/index.js
```diff
@@ -0,0 +1,119 @@
...
+    // Prefer src from local mono repo packages and the current project
+    ...(await terraClIDirectoriesFromPaths([...toolkitPackagePaths, ...currentProjectPaths], true)),
```
yuderekyu:
```suggestion
    ...(await terraClIDirectoriesFromPaths({ paths: [...toolkitPackagePaths, ...currentProjectPaths], preferSrc: true })),
```

packages/terra-cli/src/terra-dev-site/TerraCLI.tool.mdx
```diff
@@ -0,0 +1,43 @@
+# Terra CLI
+
+Terra CLI is a command line builder that supports creating [yargs commands](https://github.com/yargs/yargs/blob/master/docs/advanced.md#commands) via defining [command modules](https://github.com/yargs/yargs/blob/master/docs/advanced.md#providing-a-command-module). Currently, terra-cli searches for a `terra-cli` directory in a given dependency and assumes that all subdirectories are separate command modules defined via `index.js` files in those subdirectories.
+
+## Allow List
+
+To limit, who can create these commands, terra-cli currently uses an allow list to only search for the following dependencies:
+
+* @cerner/terra-functional-testing
```
yuderekyu, line 9: in the future, are we adding a list for closed source, such as full stack testing?

yuderekyu (approved):

packages/terra-cli/package.json
```diff
@@ -28,9 +30,16 @@
...
+    "compile": "babel src --out-dir lib --copy-files --ignore src/utils",
+    "jest": "jest --config ../../jest.config.js",
+    "lint": "eslint --ext .js,.jsx .",
+    "test": "npm run jest"
   },
   "dependencies": {
+    "fs-extra": "^9.0.1",
+    "lockfile": "^1.0.4",
+    "npmlog": "^4.1.2",
+    "yargs": "^14.2.2"
```
emilyrohrbough: Out of curiosity, why not install yargs v16?

packages/terra-cli/src/terra-dev-site/TerraCLI.tool.mdx
mjhenkes: Yeah we would.

ryanthemanuel:
> Looks good. I'm still not a fan of toggling src and lib but seems to work!
> 
> Can you run the pack script to pull this into terra-core (or equivalent) and terra-aggregated-themes to check that things are working as expected?

I pulled the tars in terra-app and terra-clinical and ran a few commands and things worked as I would expect them to.

packages/terra-cli/package.json
ryanthemanuel: Good catch. I'm not sure how I got this version other than maybe starting from lerna's starting point. The breaking changes seem to be pretty straightforward, so I'll update.

ryanthemanuel: Updated: https://github.com/cerner/terra-toolkit/pull/460/commits/5314365d74c3ef7d44104a23b5d3143723176f6d

packages/terra-cli/index.js
ryanthemanuel: Updated: https://github.com/cerner/terra-toolkit/pull/460/commits/5314365d74c3ef7d44104a23b5d3143723176f6d

packages/terra-cli/index.js
ryanthemanuel: Updated: https://github.com/cerner/terra-toolkit/pull/460/commits/5314365d74c3ef7d44104a23b5d3143723176f6d

packages/terra-cli/index.js
ryanthemanuel: Updated: https://github.com/cerner/terra-toolkit/pull/460/commits/5314365d74c3ef7d44104a23b5d3143723176f6d

packages/terra-cli/index.js
ryanthemanuel: Updated: https://github.com/cerner/terra-toolkit/pull/460/commits/5314365d74c3ef7d44104a23b5d3143723176f6d

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
StephenEsser: I think throwing the error is fine here instead of process.exit. For historical context the process.exit is the documented approach to using the test runner programmatically:

https://webdriver.io/docs/clioptions.html#run-the-test-runner-programmatically

It's also how the default runner is setup:

https://github.com/webdriverio/webdriverio/blob/master/packages/wdio-cli/src/commands/run.js#L120-L136

packages/terra-open-source-scripts/src/terra-cli/express-server/ExpressServer.js
```diff
@@ -0,0 +1,84 @@
+const fs = require('fs-extra');
```
StephenEsser, line 1: Is this essentially a copy paste of the terra-functiona-testing express server?

packages/terra-cli/src/terra-dev-site/TerraCLI.tool.mdx
```diff
@@ -0,0 +1,43 @@
+# Terra CLI
+
+Terra CLI is a command line builder that supports creating [yargs commands](https://github.com/yargs/yargs/blob/master/docs/advanced.md#commands) via defining [command modules](https://github.com/yargs/yargs/blob/master/docs/advanced.md#providing-a-command-module). Currently, terra-cli searches for a `terra-cli` directory in a given dependency and assumes that all subdirectories are separate command modules defined via `index.js` files in those subdirectories.
+
+## Allow List
+
+To limit, who can create these commands, terra-cli currently uses an allow list to only search for the following dependencies:
```
benbcai:
```suggestion
To limit who can create these commands, terra-cli currently uses an allow list to only search for the following dependencies:
```

packages/terra-open-source-scripts/src/terra-cli/express-server/ExpressServer.js
ryanthemanuel: Yes. @mjhenkes and I discussed this briefly and decided it might be best to do that until we land on a long term location for that server. It felt weird to have it exposed from functional tests or to have a dependency of this script on functional tests. This is a temporary solution

benbcai (approved):

PR #464, opened 2020-11-03
[terra-functional-testing] Validates Element

benbcai:
### Summary
Add the Terra.validates.element command to the Terra Service.

### Additional Details
This PR does not include screenshot comparison because that is currently not available. It will be added included once that is available.

@cerner/terra


packages/terra-functional-testing/tests/jest/commands/validates/element.test.js
```diff
@@ -0,0 +1,48 @@
...
+      testName: 'test-name',
+      rules: { 'mock-rule': { enabled: true } },
+      selector: 'test-selector',
+      misMatchTolerance: 10,
+    });
+
+    expect(accessibility).toHaveBeenCalledWith({ 'mock-rule': { enabled: true } });
+  });
+
+  it('should run element with service selector options', () => {
```
emilyrohrbough, line 40: Is this in place for when the screenshot logic is ready to pull in?

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -5,10 +5,11 @@ const { toBeAccessible } = require('../commands/expect');
 const { getViewports, describeViewports, setViewport } = require('../commands/viewport-helpers');
 
 const hideInputCaret = require('../commands/hide-input-caret');
+const { element } = require('../commands/validates');
 
 class TerraService {
   constructor(options = {}) {
-    this.formFactor = options.formFactor;
+    this.serviceOptions = options;
```
emilyrohrbough: This gets applied on the global browser? Nice

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -22,7 +23,11 @@ class TerraService {
 
     // Add a Terra global with access to Mocha-Chai test helpers.
     global.Terra = {
-      validates: { accessibility },
+      // validates provides access to the chai assertions to use in an it blocks.
```
emilyrohrbough:
```suggestion
      // validates provides access to the jest assertions to use in an it blocks.
```

packages/terra-functional-testing/src/commands/viewport-helpers.js
```diff
@@ -46,23 +46,19 @@ const describeViewports = (title, viewports, fn) => {
...
-  const { formFactor } = options;
+  const { formFactor } = global.Terra.serviceOptions;
   if (formFactor) {
     localViewports = viewports.includes(formFactor) ? [formFactor] : [];
   }
 
   let currentViewportSize;
   localViewports.forEach(viewport => global.describe(`[${viewport}]`, () => {
-    global.beforeAll(() => {
+    global.before(() => {
```
emilyrohrbough, line 57: :/ bummer

emilyrohrbough (commented):
Will docs be added on the next PR for this that will add the screenshot comparison?

packages/terra-functional-testing/src/commands/validates/element.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {Number} [options.misMatchTolerance] - the mismatch tolerance for the screenshot comparison.
+ * @param {string} [options.selector] - the element selector to use for the screenshot comparison.
+ */
+const element = (options = {}) => {
+  const { rules, misMatchTolerance } = options;
+
+  const selector = options.selector || global.Terra.serviceOptions.selector;
+  const testName = options.testName || 'default';
+
+  accessibility(rules);
```
StephenEsser: This will need to be passed as an object. 

```js
accessibility({ rules }); // Rules is a key
```

packages/terra-functional-testing/src/commands/validates/element.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {string} [options.testName] - the test case name.
+ * @param {Object} [options.rules] - the axe rules to use to assert accessibility.
+ * @param {Number} [options.misMatchTolerance] - the mismatch tolerance for the screenshot comparison.
+ * @param {string} [options.selector] - the element selector to use for the screenshot comparison.
+ */
+const element = (options = {}) => {
+  const { rules, misMatchTolerance } = options;
+
+  const selector = options.selector || global.Terra.serviceOptions.selector;
+  const testName = options.testName || 'default';
```
StephenEsser: nit: I think we should comment this code out until we need it (or remove it) so we don't have to disable the unused vars at the top of the file. It can be added back/uncommented when we integrate the screenshot logic

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: Yes, the options object is added to `global.Terra`.

packages/terra-functional-testing/tests/jest/commands/validates/element.test.js
benbcai: Yes, the implementation has the arguments extracted and ready to pass down to the screenshot logic.

packages/terra-functional-testing/src/commands/viewport-helpers.js
benbcai: This is using mocha. `beforeAll` and `afterAll` are not their APIs so switching back to `before` and `after`.

benbcai:
> Will docs be added on the next PR for this that will add the screenshot comparison?

Oops. Doc will be add in this PR. I just forgot.

packages/terra-functional-testing/src/commands/validates/element.js
benbcai: Good call: https://github.com/cerner/terra-toolkit/pull/464/commits/2f33038f4383a4ecf1cfc25ecdd75d06d47d5469

packages/terra-functional-testing/src/commands/validates/element.js
benbcai: Agree: https://github.com/cerner/terra-toolkit/pull/464/commits/2f33038f4383a4ecf1cfc25ecdd75d06d47d5469

packages/terra-functional-testing/src/config/wdio.conf.js
```diff
@@ -133,6 +133,7 @@ exports.config = {
     [TerraService, {
       /* Use to change the form factor (test viewport) used in the wdio run. */
       ...FORM_FACTOR && { formFactor: FORM_FACTOR },
+      selector: '[data-terra-dev-site-content] *:first-child',
```
yuderekyu: did we land on providing an override for the selector? i know we talked about making this an env. full-stack testing will ideally have this set to an orion-application selector

packages/terra-functional-testing/src/commands/validates/element.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {Object} [options] - the test options
+ * @param {string} [options.testName] - the test case name.
+ * @param {Object} [options.rules] - the axe rules to use to assert accessibility.
+ * @param {Number} [options.misMatchTolerance] - the mismatch tolerance for the screenshot comparison.
+ * @param {string} [options.selector] - the element selector to use for the screenshot comparison.
+ */
+const element = (options = {}) => {
+  const { rules } = options;
+
+  accessibility({ rules });
```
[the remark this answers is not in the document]
mjhenkes: One thing to watch out with optional chaining, it must be transpiled for node versions less that 14.0.0

packages/terra-functional-testing/src/config/wdio.conf.js
mjhenkes: I'd love to genericize this selector to not reference terra-dev-site, but i'm not sure how to do that without tying this update to terra-dev site in one way or another.

We could patch terra-dev-site v6 to provide both this data element and a generic one, same with terra dev-site v7.

I *think* it's fair to say that terra-functional-testinig is compatible with both terra dev site version 6 and 7 but not 5 or lower.

Thoughts?

mjhenkes: Proposed selectors
~Options A: `[data-terra-functional-testing-content] *:first-child`~
Option B: `[data-terra-test-content] *:first-child`

Option A links the selector to this package, Option B hedges in case we change package name in the future. 🤷‍♂️ 

Proposed loading selectors
~Option A: `[data-terra-functional-testing-loading]`~
Option B: `[data-terra-test-loading]`

edit: The people have spoken, Option B wins.

benbcai: Moved the selector key to top level to ease configuration and updated with the name to be `data-terra-test-content`. https://github.com/cerner/terra-toolkit/pull/464/commits/9f3cf680ca8314a43c1bcf131747bfa07ad4b683

mjhenkes: Added the attributes to dev site:
v6: https://github.com/cerner/terra-dev-site/pull/314
v7: https://github.com/cerner/terra-dev-site/pull/313 (Admittedly theres a lot more going on in this pr than just data attributes).

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -36,6 +41,8 @@ class TerraService {
       // Hides the blinking input caret that appears in inputs or editable text areas.
       hideInputCaret,
 
+      serviceOptions: this.serviceOptions,
```
StephenEsser: I think we can rework how these options get applied in the utility function. Rather than binding these service options to the global environment we can just call the function from the service with the options. I'll work on a branch with an example to make this more clear using the `accessibility` helper and axe options.

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -5,27 +5,32 @@ const { toBeAccessible } = require('../commands/expect');
...
+const { element } = require('../commands/validates');
 
 class TerraService {
   constructor(options = {}) {
-    const { formFactor, theme } = options;
-
-    this.formFactor = formFactor;
-    this.theme = theme || 'terra-default-theme';
+    const theme = { theme: 'terra-default-theme' };
+    this.serviceOptions = { ...theme, ...options };
```
StephenEsser: Theme can probably be inlined here rather than created into an object and then deconstructed, but I think these service options might go away

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -5,27 +5,32 @@ const { toBeAccessible } = require('../commands/expect');
...
   }
 
   /**
    * Service hook executed prior to test execution.
    * Initializes the Terra Service's custom commands.
    */
   before(capabilities) {
+    this.serviceOptions = { ...this.serviceOptions, ...global.browser.config.serviceOptions };
```
StephenEsser: We should use a hook like `onPrepare` that returns the config option to set these options rather than reading from the global scope

packages/terra-functional-testing/src/config/wdio.conf.js
```diff
@@ -171,4 +171,8 @@ exports.config = {
     ui: 'bdd',
     timeout: 60000,
   },
+
+  serviceOptions: {
+    selector: '[data-terra-test-content] *:first-child',
```
StephenEsser: What are your thoughts on defaulting this in the service rather than inline in the config? This would ensure a default for the service and consumers wouldn't have to merge serviceOptions together if more override options are added

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-services/TerraService.tool.mdx
```diff
@@ -62,6 +62,29 @@ export.config = {
...
 
+### selector
+
+An optional selector can be configured to specify the element to be captured in the screenshot. The default selector is `data-terra-test-content`. Accessibility testing ignores this option and always tests the entire document.
+
+Type: `string`
+
+Required: `false`
+
+Default: `[data-terra-test-content] *:first-child`
```
StephenEsser, line 73: With the current implementation the default value for the service is actually `undefined` because the value is set in the `wdio.conf`, but a different `wdio.conf` or one that accidentally overwrite `serviceOptions` using the Terra service would end up with an undefined selector

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/464/commits/04718224294139afe8b92e05a16b2910ba550b35

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: That makes sense: https://github.com/cerner/terra-toolkit/pull/464/commits/04718224294139afe8b92e05a16b2910ba550b35

packages/terra-functional-testing/src/config/wdio.conf.js
benbcai: In that case, do we still keep `serviceOption` but leave it empty? 
https://github.com/cerner/terra-toolkit/pull/464/commits/04718224294139afe8b92e05a16b2910ba550b35

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-services/TerraService.tool.mdx
benbcai: This [update](https://github.com/cerner/terra-toolkit/pull/464/commits/04718224294139afe8b92e05a16b2910ba550b35) now defaults the selector in the service.

packages/terra-functional-testing/src/config/wdio.conf.js
```diff
@@ -171,4 +171,6 @@ exports.config = {
     ui: 'bdd',
     timeout: 60000,
   },
+
+  serviceOptions: {},
```
yuderekyu: do we need to define this if we aren't setting any custom service options at the config level?

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-services/TerraService.tool.mdx
```diff
@@ -62,6 +62,29 @@ export.config = {
...
+
+Example:
+
+```js
+// wdio.conf.js
+export.config = {
+    // ...
+    serviceOptions: {
```
yuderekyu, line 81: doc wise, i think we could split the constructor options and serviceOptions into separate sections

yuderekyu (approved):

packages/terra-functional-testing/src/config/wdio.conf.js
StephenEsser: +1 to removing this if there are no options

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -5,13 +5,38 @@ const { toBeAccessible } = require('../commands/expect');
...
+       before hook. The reason being webdriverio v6 now executed the describe block prior to running the before hook.
+       Therefore, this command needs to be defined before the test starts in the testing life cycle.
 
-    this.formFactor = formFactor;
-    this.theme = theme || 'terra-default-theme';
+       Reference: https://github.com/webdriverio/webdriverio/issues/6119
+      */
+      describeViewports,
+
+      serviceOptions: this.serviceOptions,
```
StephenEsser, line 38: To move this PR forward I think we should keep this as is, but in the long run I think we can/should avoid adding the options to the global namespace and call the function directly with the necessary parameters. It might end up looking something like this:

https://github.com/cerner/terra-toolkit/compare/bind-axe-options

In this case we bind the accessibility function to the instance and proxy the call with the configuration override. We could also bind the utility functions with the options we expect. 

```js
describeViewports: describeViewports.bind(null, formFactor); // Always call describeViewports with the form factor as the first parameter.
```

StephenEsser (approved):

PR #465, opened 2020-11-03
[terra-functional-testing] Inline Documentation in VR Service

emilyrohrbough:
### Summary

Adds some inline documentation int he visual regression server. This is not an exhaustive list, I'm just adding them as I come across. Figured less is more with these types of changes.


packages/terra-functional-testing/src/services/wdio-visual-regression-service/scripts/virtualScroll.js
```diff
@@ -1,8 +1,17 @@
-export default function virtualScroll(x, y, remove) {
-  const w = x === 0 ? 0 : -1 * x;
-  const h = y === 0 ? 0 : -1 * y;
+/**
+ * Script to apply a negative 2D-translate transformation method to move the document from its current position
+ * to up and to the right for the specified number of pixels horizontally and vertically.
```
benbcai: I wasn't sure what to make of this. "...current position to up and to the right...".  Should it just be "...current position to the specified..."?

benbcai (approved):

packages/terra-functional-testing/src/services/wdio-visual-regression-service/scripts/virtualScroll.js
emilyrohrbough: yeah this one was weird. was trying to caputre that this changes the value to the negative x or y value provided

packages/terra-functional-testing/src/services/wdio-visual-regression-service/scripts/pageHeight.js
```diff
@@ -1,3 +1,9 @@
+/**
+ * Script to set the document height.
+ *
+ * @param {String} height - The height to set the document to.
```
StephenEsser, line 4: Is height a string to account for units? IE: `"20px"`

packages/terra-functional-testing/src/services/wdio-visual-regression-service/scripts/scroll.js
```diff
@@ -1,3 +1,10 @@
+/**
+ * Script to scroll to a particular set of coordinates in the document.
+ *
+ * @param {Number} x - The pixel along the horizontal axis of the document.
+ * @param {Number} y - The pixel along the vertical axis of the document.
+ * @returns null
```
StephenEsser: Functions without an explicit return will return `undefined`

`null !== undefined`

packages/terra-functional-testing/src/services/wdio-visual-regression-service/scripts/pageHeight.js
emilyrohrbough: Yeah exactly. It's used in these places:
- https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/services/wdio-visual-regression-service/modules/makeAreaScreenshot.js#L56
- https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/services/wdio-visual-regression-service/modules/makeAreaScreenshot.js#L98

StephenEsser (approved):

PR #466, opened 2020-11-04
[terra-functional-testing] Remove axe option from the Terra Service

### LINKED ISSUE #463: [terra-functional-testing] Investigate removing the axe option from the terra service
# Feature Request

## Description

Investigate removing the axe option from the Terra Service. The `axe` option is currently necessary for dynamically enabling the color contrast rule for different themes.

```
  axe: {
    rules: [
      // The lowlight theme adheres to a non-default color contrast ratio and fails the default ratio check.
      // The color-contrast ratio check must be disabled for lowlight theme testing.
      { id: 'color-contrast', enabled: theme !== 'clinical-lowlight-theme' },
    ],
  },
```

The rule overrides should instead be applied and flex internally by the service.

StephenEsser:
### Summary

This PR removes the `axe` override from the Terra Service configuration options. These rules are now set explicitly by the service.

Resolves https://github.com/cerner/terra-toolkit/issues/463

### Additional Details

To remove the `axe` option a `theme` option was required to properly configure the axe rules based on the current theme. The theme is currently used for the lowlight theme to adjust the color contrast rules.

https://github.com/cerner/terra-toolkit-boneyard/blob/main/config/wdio/wdio.conf.js#L100-L102

@cerner/terra


packages/terra-functional-testing/src/commands/axe/run.js
```diff
@@ -1,49 +1,38 @@
...
-    injectAxe(axeOptions);
-  }
+    /**
+     * Converts the global rule overrides into an array.
+     * The axe.configure API requires the rules to be an array of objects. The axe.run API requires
+     * the rules to be an object keyed by the rule ID.
+     */
+    const globalRuleArray = Object.keys(Terra.axe.rules).map((rule) => (
+      { ...Terra.axe.rules[rule], id: rule }
+    ));
```
emilyrohrbough, line 21:
```suggestion
    const globalRuleArray = Object.entries(Terra.axe.rules).map((rule, settings) => (
      { ...settings, id: rule }
    ));
```

emilyrohrbough (approved):

yuderekyu (approved):

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -8,30 +8,54 @@ const hideInputCaret = require('../commands/hide-input-caret');
 
 class TerraService {
   constructor(options = {}) {
-    this.formFactor = options.formFactor;
+    const { formFactor, theme } = options;
+
+    this.formFactor = formFactor;
+    this.theme = theme || 'terra-default-theme';
```
benbcai, line 14: In [#464](https://github.com/cerner/terra-toolkit/pull/464/files#diff-a6ba3c1d16c08e58e7ee3994f3ab541b63b2a3a8dafb0de9daed2c8090c948d9) I'm exposing the entire service options object thru `Terra.serviceOptions`. I should probably default `Terra.serviceOptions.theme` to `terrra-default-theme` to keep them in sync. Would it be better to default the theme to `terrra-default-theme` in wdio.config.js or are we concerned that this config may get overridden by consumers.

StephenEsser: I think the TerraService should default the option if it needs a default rather than requiring a theme be set in the wdio.config

benbcai (approved):

PR #487, opened 2020-11-20
Update to use prompts instead of inquirer. It has a smaller memory footprint

### LINKED ISSUE #485: [terra-open-source-scripts] replace inquirer with prompts
# Feature Request

Inquirer is a fairly large package. We should replace it with [prompts](https://www.npmjs.com/package/prompts) which is much smaller and fairly similar.

ryanthemanuel:
### Summary

Update to use prompts instead of inquirer. It has a smaller memory footprint

Fixes #485


package.json
```diff
@@ -92,6 +92,7 @@
     "@babel/preset-env": "^7.11.0",
     "@babel/preset-react": "^7.0.0",
     "@babel/register": "^7.10.5",
+    "@octokit/core": "^3.1.2",
```
emilyrohrbough, line 95: accidental add?

emilyrohrbough (approved):

mjhenkes (approved):

package.json
ryanthemanuel: Nope.  It's a peer dependency of something from danger :-/

PR #492, opened 2020-12-01
[webpack-config-terra] Upgrade to postcss 8 and support webpack 5

mjhenkes:
### Summary

This PR resolves unmeet peer dependency warnings for several postcss plugins by upgrading to postcss 8

Previously we had not upgraded due to postcss-rtl. I've put out a [PR to upgrade postcss-rtl to postcss 8](https://github.com/vkalinichev/postcss-rtl/pull/73), but it seems like postcss-rtl isn't very active. To that ends I looked into converting over to [rtlcss](https://github.com/MohammadYounes/rtlcss) (which is on postcss 6 but still seems compatible with postcss 8 🤷‍♂️ ). However a trick we're using to ignore rtl switches that works with postcss-rtl, doesn't work with rtlcss. See [popup overlay](https://github.com/cerner/terra-framework/blob/main/packages/terra-popup/src/PopupOverlay.module.scss#L7-L8) for an example.

So I landed on forking postcss-rtl to solve the current problem. I released [@mjhenkes/postcss-rtl](https://www.npmjs.com/package/@mjhenkes/postcss-rtl) with my postcss 8 upgrade PR until we either convert to rtlcss or the postcss-rtl PR gets merged.

I've also added support for webpack 5. This pr still uses webpack 4 but this validation pr #495 uses webpack 5.

### Additional Details

I've also upgraded our theme plugin to use the postcss 8 apis per this upgrade guide: https://evilmartians.com/chronicles/postcss-8-plugin-migration.

Validation PRs using a tgz generated from this package
[terra-framework#1337](https://github.com/cerner/terra-framework/pull/1337)

@cerner/terra


packages/webpack-config-terra/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Upgrade to postcss 8.
```
emilyrohrbough, line 6: Can we add the other dependency changes here as well?

packages/webpack-config-terra/lib/postcss/ThemePlugin.js
```diff
@@ -13,7 +13,7 @@ const SUPPORTED_THEMES = [
  * The purpose of this plugin is to create a default theme from a scoped theme
  * and to remove any supported themes that are not desired.
  */
-module.exports = postcss.plugin('terra-theme-plugin', (config) => {
```
emilyrohrbough, line 16: This syntax is compatible for both webpack 4 and 5?

packages/webpack-config-terra/package.json
```diff
@@ -56,7 +56,7 @@
     "@babel/core": "^7.4.4",
     "core-js": "^3.1.3",
     "regenerator-runtime": "^0.13.2",
-    "webpack": "^4.30.0",
-    "webpack-cli": "^3.3.2"
+    "webpack": "^4.30.0 || 5.0.0",
+    "webpack-cli": "^3.3.2 ||4.0.0"
   }
```
emilyrohrbough: Should allowing for webpack 5 be its PR?

mjhenkes: I mean, sure it could be, but it builds on this PR, it's fairly small, and I haven't gotten much feeback on this review yet.

packages/webpack-config-terra/CHANGELOG.md
mjhenkes, line 6:
```suggestion
  * Upgrade to postcss 8.
    * AutoPrefixer to v10
    * PostCss config to a forked version to support postcss8
  * Support webpack 5
```

packages/webpack-config-terra/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
...
+    * AutoPrefixer to v10
+    * PostCss config to a forked version to support postcss8
```
mjhenkes:
```suggestion
    * Postcss-rtl to a forked version to support postcss8
```

packages/webpack-config-terra/package.json
```diff
@@ -32,19 +32,19 @@
...
     "file-loader": "^6.2.0",
     "mini-css-extract-plugin": "^1.3.1",
     "node-sass": "^5.0.0",
-    "postcss": "^7.0.0",
+    "postcss": "^8.0.0",
     "postcss-assets-webpack-plugin": "^4.0.0",
     "postcss-custom-properties": "^10.0.0",
     "postcss-loader": "^4.1.0",
-    "postcss-rtl": "^1.3.3",
+    "@mjhenkes/postcss-rtl": "^2.0.0",
```
StephenEsser, line 47: Long term, is there any way we could remove this dependency? There's not as many downloads of this package from npm as I was expecting. I'm curious if other projects are manually writing the rtl rules, ignoring them entirely, or using something else.

As a stepping stone to postcss 8 I think this makes sense, but I suspect this package might end up being something that persists for quite a while. Is this something we want to move into @cerner if that ends up being the case?

StephenEsser (approved):

packages/webpack-config-terra/package.json
mjhenkes: I've put out a PR to upgrade postcss-rtl to postcss 8, but it seems like postcss-rtl isn't very active. To that ends I looked into converting over to rtlcss (which is on postcss 6 but still seems compatible with postcss 8 🤷‍♂️ ). However a trick we're using to ignore rtl switches that works with postcss-rtl, doesn't work with rtlcss. See popup overlay for an example.

I think we can switch to rtlcss but we'd have to upgrade all our usages of the ignore trick to a better ignore directive that happens to be cross compatible.

I'm with you, i want this to be temporary but if it's not, we'll have to move it to `@cerner`

packages/webpack-config-terra/lib/postcss/ThemePlugin.js
```diff
@@ -30,22 +30,26 @@ module.exports = postcss.plugin('terra-theme-plugin', (config) => {
...
+            node.parent.removeChild(node);
+          }
+        });
+      }
     }
-  };
-});
+  }
+};
+module.exports.postcss = true
```
yuderekyu: not a big deal, but i had to dig a little to see what this meant
```suggestion
module.exports.postcss = true // distinguish between require('plugin') and require('plugin')(opts) end-user calls
```

yuderekyu (approved):

packages/webpack-config-terra/lib/postcss/ThemePlugin.js
mjhenkes: I was just following the plugin upgrade guide 😅

packages/webpack-config-terra/package.json
```diff
@@ -56,7 +56,7 @@
     "@babel/core": "^7.4.4",
     "core-js": "^3.1.3",
     "regenerator-runtime": "^0.13.2",
-    "webpack": "^4.30.0",
-    "webpack-cli": "^3.3.2"
+    "webpack": "^4.30.0 || 5.0.0",
+    "webpack-cli": "^3.3.2 ||4.0.0"
```
ryanthemanuel: Should this be:

```suggestion
    "webpack": "^4.30.0 || ^5.0.0",
    "webpack-cli": "^3.3.2 || ^4.0.0"
```

mjhenkes: good call

ryanthemanuel (approved):

benbcai (approved):

PR #498, opened 2020-12-04
[terra-functional-testing] Restructure utility functions

StephenEsser:
### Summary

This PR restructures the viewport utility methods into individual files to align more closely with the other commands. 

A breakdown of the changes:

- `describeViewports`, `setViewports`, and `getViewports` are split out of `viewport-helpers.js` into individual commands and moved into `commands/utils/` directory.
- `hide-input-caret` was renamed to `hideInputCaret` and moved into `commands/utils/`. The naming was different than any of our other command files
- `static-assets/viewports` was renamed to `constants/`
- `setViewport` utility function doesn't default to `huge`. It now requires a size. The size `huge` is defaulted by the service.
- Viewport helper tests restructured to match the new file structure

### Additional Details

@cerner/terra


packages/terra-functional-testing/src/commands/utils/describeViewports.js
```diff
@@ -0,0 +1,40 @@
...
+    if (formFactor && viewport !== formFactor) {
+      return;
+    }
+
+    global.describe(`[${viewport}]`, () => {
+      global.before(() => {
+        // Store the original browser window size so it can be restored after the test has run.
+        previousViewportSize = global.browser.getWindowSize();
+
+        setViewport(viewport);
```
StephenEsser, line 27: It has been mentioned before that our `describeViewports` function can be slow. I haven't done too much investigation into why, but a good place to start might be looking into how expensive accessing/resizing the window is. When we run in a specific form factor nearly all of these calls are redundant (get size, set size, reset size again). I left these in here for passivity, but I think there are some optimizations we can do here.

packages/terra-functional-testing/src/commands/utils/getViewports.js
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Returns an array of viewport dimensions.
+ * @param {string|string[]} sizes - A list of Terra viewport sizes.
+ * @return {array} - An array of viewports. Returns all viewports if sizes are not provided.
+ */
+const getViewports = (...sizes) => {
+  const viewports = sizes.length > 0 ? sizes : ALL_VIEWPORTS;
+
+  return viewports.map((name) => TERRA_VIEWPORTS[name]);
```
StephenEsser, line 13: I kept this the same, but it might make sense to have this return the name too depending on how people are consuming this. It might look something like this:

```js
  return viewports.map((name) => {
    const { height, width } = TERRA_VIEWPORTS[name];

    return { height, width, name }; // or { ...TERRA_VIEWPORTS[name], name }
  });
```

mjhenkes: Do we even need this helper? Could we just expose the 'All_VIEWPORTS' map directly? with a better name.

mjhenkes: Consumers could get all viewports by Object.values, and always know what name relates to what size.

packages/terra-functional-testing/src/commands/utils/describeViewports.js
mjhenkes: I hope retrieving the window size isn't the slow part but we could add a quick check to not try to reset the viewport if the sizes are the same.

packages/terra-functional-testing/src/commands/utils/getViewports.js
StephenEsser: I was thinking something similar when I started making these changes. This is one of the topics that was brought up early in our WDIO sync and we settled on keeping this for passivity. Quite a few consumers are using this, but they should probably be using `describeViewports`

mjhenkes: If this is hanging out for passivity, i wouldn't update the api.

mjhenkes (approved):

packages/terra-functional-testing/src/commands/utils/describeViewports.js
```diff
@@ -0,0 +1,40 @@
+const setViewport = require('./setViewport');
+
+/**
+ * Mocha `describe` block to run tests through a defined list of viewports.
```
ryanthemanuel, line 4: Is this still mocha?

StephenEsser: We're still using the mocha testing framework here, but we use jest (expect) as the assertion library

packages/terra-functional-testing/tests/jest/commands/utils/describeViewports.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+    global.Terra = {
+      serviceOptions: {
+        formFactor: 'tiny',
+      },
+    };
+
+    describeViewports('viewport', ['tiny', 'large'], () => {});
+
+    expect(global.describe).toHaveBeenCalled();
```
emilyrohrbough, line 21:
```suggestion
    expect(global.describe).toHaveBeenCalledOnce();
```

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -1,42 +1,51 @@
...
+     * This command must be defined in the beforeSession hook instead of together with the other Terra custom commands in the
+     * before hook. The reason being WebdriverIO v6 now executes the describe block prior to running the before hook.
+     * Therefore, this command needs to be defined before the test starts in the testing life cycle.
+     *
+     * Reference: https://github.com/webdriverio/webdriverio/issues/6119
+     */
+    global.Terra.describeViewports = describeViewports;
+
+    // Add the service options to the global.
+    global.Terra.serviceOptions = this.serviceOptions;
```
emilyrohrbough, line 48: Why do these need to be added to the Terra global?

StephenEsser: This exposes the service options globally so the utility functions can reference them. This is existing functionality:

- [Related discussion/PR that introduced this change](https://github.com/cerner/terra-toolkit/pull/464#discussion_r516734946)
- [Example of where it's used](https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/commands/viewport-helpers.js#L50)

Long term we can probably bind service options to the function to pass them along, but it can create a long chain of information for certain functions that call other functions. I didn't make any changes to exposing the options globally on this PR

packages/terra-functional-testing/tests/jest/commands/utils/describeViewports.test.js
StephenEsser: This doesn't appear to be a supported expect assertion

```
 FAIL  packages/terra-functional-testing/tests/jest/commands/utils/describeViewports.test.js
  describeViewports
    ✕ should describe viewport (5 ms)

  ● describeViewports › should describe viewport

    TypeError: expect(...).toHaveBeenCalledOnce is not a function

      19 |     describeViewports('viewport', ['tiny', 'large'], () => {});
      20 | 
    > 21 |     expect(global.describe).toHaveBeenCalledOnce();
         |                             ^
      22 |   });
      23 | });
      24 | 

      at Object.<anonymous> (packages/terra-functional-testing/tests/jest/commands/utils/describeViewports.test.js:21:29)
```

benbcai (approved):

PR #505, opened 2020-12-10
[terra-functional-testing] Add themes locales and form factor options to the test runner

StephenEsser:
### Summary

This PR adds `themes`, `locales`,  and `form factors` as options to the test runner. The test runner will initiate a new process for each permutation of these options. The process are run sequentially.

Important note: When are now using [yargs](https://github.com/yargs/yargs) as our command line parser. Yargs has first class support for array arguments, however, array arguments are delimited by spaces. This is different than our previous array implementation. 

```
// Previous format
terra -- wdio --formFactors=['tiny','huge'] --locales=['en','fr']

// New format
terra -- wdio --formFactors tiny huge --locales en fr
```

All tests are run with a theme. By default the theme is `terra-default-theme`.

@cerner/terra


mjhenkes (approved):

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
```diff
@@ -2,15 +2,52 @@
...
+| locales     | array  | false    | ['en']                  | A list of language locales for the test run.                                                  |
+| themes      | array  | false    | ['terra-default-theme'] | A list of themes for the test run.                                                            |
+
+The following example will run the test suite a total of four times. Once for each permutation of the specified locales and form factors.
+
+package.json
+```json
+{
+  "scripts": {
+    "wdio": "npm run terra -- wdio --locales en fr --formFactors tiny huge"
```
emilyrohrbough, line 34: If passed like the previous implementation, will it still run as expected?

StephenEsser: No, the old way would not work. yargs interprets arrays by arguments delimited by spaces:

https://github.com/yargs/yargs/blob/fd66d8e95f14302a376b7030e8d8dae869af

We discussed this early in the WDIO sync and decided to move forward with how yargs interprets arrays. We could probably do some custom parsing like we're doing today and opt out of the yargs array support, it would just depend on how much value we thought that would have. Array data structures don't really "exist" on the command line. Our current custom strategy mutates a single string argument and coerces it into an array.

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -7,12 +7,26 @@ const cli = {
     yargs.options({
       c: {
         type: 'string',
-        alias: 'configuration',
+        alias: 'config',
         describe: 'A file path to the test runner configuration.',
       },
+      formFactors: {
+        type: 'array',
+        describe: 'A list of form factors for the test run.',
```
benbcai, line 15: Do we want to set a default formFactor of `huge`?

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
```diff
@@ -2,15 +2,52 @@
...
+Tests can be executed in any of the following supported form factors:
+
+| Size     | Width | Height |
+|----------|-------|--------|
+| tiny     | 470   | 768    |
+| small    | 622   | 768    |
+| medium   | 838   | 768    |
+| large    | 1000  | 768    |
+| huge     | 1300  | 768    |
+| enormous | 1500  | 768    |
```
benbcai, line 50: 👍  for formatting.

packages/terra-functional-testing/src/terra-cli/wdio/index.js
StephenEsser: No, we're unfortunately in a bit of on odd situation with the form factor. Defaulting to `huge` here will set the `process.env.FORM_FACTOR` to `huge` which will filter all the `describeViewports` tests. Setting no default here will default to huge in the service's before hook, but allow other form factors in `describeViewports`

benbcai (approved):

PR #508, opened 2020-12-14
[terra-functional-testing] Wait for docker stack deployment is ready.

benbcai:
### Summary
Fix the follow error that occurs sporadically when executing two or more wdio test runs back to back. 

`webdriver: Request failed with status 500 due to Error forwarding the new session cannot find : Capabilities {browserName: chrome}`

### Additional Details
There are times where the docker stack deployment for wdio is not completely ready before wdio session begins. The changes made in here are to add a 5 second wait and check for the wdio docker services and network are created before moving forward. Tested many times using a wait time of 1 second, 2 seconds and 3 seconds and error still occurs although less frequent when the wait time is slightly longer. Tests consistently are passing when waiting for 5 seconds.

@cerner/terra


packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
```diff
@@ -66,6 +66,12 @@ class SeleniumDockerService {
 
     await exec(`TERRA_SELENIUM_DOCKER_VERSION=${this.version} docker stack deploy -c ${composeFilePath} wdio`);
 
+    // There are occasions where the docker stack deployment is not completely ready when running back-to-back wdio test runs.
+    // As a result, the test run is unable to start and receives the 'Error forwarding the new session cannot find : Capabilities {browserName: chrome}' error.
+    // This waits for 5 seconds and ensure the services and network are created and ready.
+    await this.wait(5000);
```
StephenEsser: :(

packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
```diff
@@ -66,6 +66,12 @@ class SeleniumDockerService {
...
+    await this.waitForServiceCreation();
+    await this.waitForNetworkCreation();
```
StephenEsser: Were these two service/network waits tested with and without the time wait above? Are all three waits necessary to get reliable test runs or is the 5 second wait enough? These are probably fine to add either way, but I am curious if they play a role in increasing the reliability of avoiding the session error or if the timed wait is the main contributing factor

benbcai: The services and network get created fairly quickly...less than a second from my testing. Unfortunately, just checking the services and network without the wait still gives the error. Adding just the wait alone gives a reliable test run but I thought it doesn't hurt to also verify that the services and network are created. I looked at the [docker stack deploy doc](https://docs.docker.com/engine/reference/commandline/stack_deploy/#examples) but don't see what else I could check other than ensuring that the services and network are created.

packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
emilyrohrbough: Is it possible we need to "beef" up the removal of the network and service? RemoveStack only wait for the tear down of the wdio service/network if the stack is found. I think it's robust enough it could always be called:

https://github.com/cerner/terra-toolkit/blob/128260b396da3fda433e552a52724de35669f99f/packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js#L89-L91

benbcai: One additional thing we could check as part of the stack removal is to ensure we can no longer connect to http://localhost:4444/wd/hub/status. Kind of similar to [waitForNetworkReady](https://github.com/cerner/terra-toolkit/blob/128260b396da3fda433e552a52724de35669f99f/packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js#L203) but resolve if we can no longer connect to it. However, I don't think doing so adds anything to fix this error.

StephenEsser (approved):

packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
mjhenkes: Is this wait is only required when running wdio multiple times sequentially. I wonder if we're punishing everyone for a case most common with ourselves. Should we instead inject the wait in the script?

benbcai: Wouldn't injecting the wait in the wdio script would only fix it for us but not for consumers? They would also need to add the wait themselves in their script.

ryanthemanuel: I think @mjhenkes meant the travis script in the .travis.yml file or the script in our package.json

mjhenkes (approved):

PR #510, opened 2020-12-16
[terra-functional-testing] Add additional options to the test runner cli

StephenEsser:
### Summary

This PR adds the `hostname`, `port`, `baseUrl`, `suite`, and `spec` options to the test runner cli. 

These options are a direct pass-through to the [WDIO CLI Options](https://webdriver.io/docs/clioptions.html)

@cerner/terra


packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
```diff
@@ -73,13 +79,19 @@ class TestRunner {
...
+   * @param {array} options.suite - Overrides specs and runs only the defined suites.
+   * @param {array} options.spec - A list of spec file paths.
    */
   static async start(options) {
     const {
       config,
       formFactors = [],
       locales,
       themes,
+      ...launcherOptions // hostname, port, baseUrl, suite, and spec
```
StephenEsser, line 94: This is a sneaky way to pass the wdio-cli options without defining keys. We don't want to deconstruct the keys otherwise we have to filter them back out. Deconstructing leads to undefined values. 
```
// We don't want to pass undefined keys
{ host: undefined } 
```

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -24,6 +24,26 @@ const cli = {
         describe: 'A list of themes for the test run.',
         default: ['terra-default-theme'],
       },
+      hostname: {
+        type: 'string',
+        describe: 'Automation driver host address.',
```
emilyrohrbough, line 29: This is http://0.0.0.0 by default for the Launcher right?

StephenEsser: The Launcher/WebdriverIO config default is `localhost`:

https://webdriver.io/docs/options.html

https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/config/wdio.conf.js#L109

emilyrohrbough: Should this be documented this so it's not hidden knowledge? I assume we didn't add defaults so the launcher defaults will be used

StephenEsser: I think this would be fine to document in the table on the doc page. We'll want to be careful with documentation since these defaults get applied at different areas in the tech stack.

The CLI has no defaults for these values since they get applied internally by the launcher and WDIO config.

Added port and hostname defaults to the documentation table: 

https://github.com/cerner/terra-toolkit/pull/510/commits/21856f02c15398745f1174afe06a3f15cd331654

emilyrohrbough: Why do you think we need to be careful about communication the behavior of the script? Consumers won't and shouldn't know or care of the implementation details of the script. If we are worried about being too reliant on the defaults changing, the CLI should provide it's own defaults IMO.

StephenEsser: We need to be careful because these defaults are not applied at this level. There is not actually a default at this level, but there is a default hostname that is applied through the tech stack. Documenting this behavior is fine and more thorough information can be added to our doc pages, but I don't think we should try to squeeze it all into the cli descriptions.

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -24,6 +24,26 @@ const cli = {
...
+      },
+      baseUrl: {
+        type: 'string',
+        describe: 'The base url.',
+      },
+      suite: {
+        type: 'array',
+        describe: 'Overrides specs and runs only the defined suites.',
+      },
+      spec: {
```
benbcai, line 43: Should this be `specs` with an `s` that matches [here](https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/config/wdio.conf.js#L42)?

StephenEsser: No, this tricked me too when I first added it because it's an `array` type but singular flag. I tested this locally and you can add an array of spec file paths:

Checkout the CLI options here:

https://webdriver.io/docs/clioptions.html

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -24,6 +24,26 @@ const cli = {
...
+      },
+      port: {
+        type: 'number',
+        describe: 'Automation driver port.',
+      },
+      baseUrl: {
+        type: 'string',
+        describe: 'The base url.',
+      },
+      suite: {
```
benbcai, line 39: How does an example of a suite look like? Is it just an array of the wdio test file names?

packages/terra-functional-testing/tests/jest/terra-cli/wdio/test-runner.test.js
```diff
@@ -118,5 +117,25 @@ describe('Test Runner', () => {
...
+
+    it('should initiate a test runner with custom hostname and port options', async () => {
+      jest.spyOn(TestRunner, 'run').mockImplementation(() => Promise.resolve());
+
+      await TestRunner.start({
+        config: '/path',
+        locales: ['en'],
+        themes: ['terra-default-theme'],
+        hostname: 'hostname',
+        port: 8080,
```
benbcai, line 129: Any reason why `baseUrl`, `suite` and `spec` are not tested here or in a separate test?

StephenEsser: I just chose a few options to verify they were being passed through. I went ahead and updated the test to include all of the options:

https://github.com/cerner/terra-toolkit/pull/510/commits/7b5e61c181f16d2f4624066fedcdf940076d04d8

packages/terra-functional-testing/src/terra-cli/wdio/index.js
StephenEsser: Suites get defined in the config. Suites are groups of specs. So it would be an array of suite names:

https://webdriver.io/docs/organizingsuites.html#grouping-test-specs

packages/terra-functional-testing/src/terra-cli/wdio/index.js
benbcai: Ah, I was looking at https://webdriver.io/docs/options.html#specs.

packages/terra-functional-testing/tests/jest/terra-cli/wdio/test-runner.test.js
benbcai: Not really. There are several options added but saw only a couple are included here.

benbcai (approved):

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
```diff
@@ -102,6 +114,7 @@ class TestRunner {
             formFactor,
```
yuderekyu, line 114: late to the party, but dang, i like the test runner architecture here. i'll rearrange the full stack test runner to mirror this format. it's much cleaner

mjhenkes (approved):

PR #519, opened 2021-01-07
[terra-functional-testing] Custom event dispatcher

### LINKED ISSUE #516: [terra-functional-testing] Create custom event dispatch helper
# Feature Request
Add a helper that dispatches custom events within WDIO tests. This will be used to verify terra-application's locale event listener works [[1](https://github.com/cerner/terra-application/blob/main/tests/wdio/application-base/private/test-overrides-spec.js#L2-L11)] and by the internal full-stack service to trigger a locale override for deployed terra applications. Consumers could also use this to test their own custom events.

Clone of https://github.com/cerner/terra-toolkit-boneyard/issues/3
There's a boneyard branch with this change; as a part of upgrading full-stack-testing to wdio 6, we need to move this logic into terra-functional-testing.

```js
const dispatchCustomEvent = (name, metaData) => {
  try {
    global.browser.execute((eventName, eventMetaData) => {
      /* If IE support is removed, convert below to use event constructors. */
      const event = document.createEvent('Event');
      event.initEvent(eventName, true, true);
      event.metaData = eventMetaData;
      window.dispatchEvent(event);
    }, name, metaData);
  } catch (error) {
    throw new Error(error);
  }
};

export default dispatchCustomEvent;
```

yuderekyu:
### Summary

Create a custom event dispatcher function that is able to dispatch a specified event with metadata. Used to override locales in full stack testing. Closes #516 
### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-functional-testing/src/commands/utils/dispatchCustomEvent.js
+++ b/packages/terra-functional-testing/src/commands/utils/dispatchCustomEvent.js
@@ -1,4 +1,10 @@
-const dispatchCustomEvent = (name, metaData) => {
+/**
+ * Utility to send a custom event containing metadata.
+ * @param {string} options.name - name of event
+ * @param {object} options.metaData - metadata pertaining to event
+ */
+const dispatchCustomEvent = (options) => {
+  const { name, metaData } = options;
   try {
     global.browser.execute((eventName, eventMetaData) => {
       /* If IE support is removed, convert below to use event constructors. */
--- a/packages/terra-functional-testing/src/commands/utils/dispatchCustomEvent.js
+++ b/packages/terra-functional-testing/src/commands/utils/dispatchCustomEvent.js
@@ -1,0 +1,15 @@
+const dispatchCustomEvent = (name, metaData) => {
+  try {
+    global.browser.execute((eventName, eventMetaData) => {
+      /* If IE support is removed, convert below to use event constructors. */
+      const event = document.createEvent('Event');
+      event.initEvent(eventName, true, true);
+      event.metaData = eventMetaData;
+      window.dispatchEvent(event);
+    }, name, metaData);
+  } catch (error) {
+    throw new Error(error);
+  }
+};
+
+export default dispatchCustomEvent;
--- a/packages/terra-functional-testing/src/commands/utils/dispatchCustomEvent.js
+++ b/packages/terra-functional-testing/src/commands/utils/dispatchCustomEvent.js
@@ -8,8 +8,8 @@ const dispatchCustomEvent = (name, metaData) => {
       window.dispatchEvent(event);
     }, name, metaData);
   } catch (error) {
-    throw new Error(error);
+    throw new Error(`dispatchCustomEvent failed: ${error}`);
   }
 };
 
-export default dispatchCustomEvent;
+module.exports = dispatchCustomEvent;
--- a/packages/terra-functional-testing/src/services/wdio-terra-service.js
+++ b/packages/terra-functional-testing/src/services/wdio-terra-service.js
@@ -98,6 +98,12 @@ class TerraService {
   }
 
   afterCommand(commandName, _args, _result, error) {
+    // Expose flag to disable hidden input caret function for full stack testing; IE cannot handle multiple async afterCommand requests.
+    const { disabledTerraServiceAfterCommand } = this.serviceOptions;
+    if (disabledTerraServiceAfterCommand) {
+      return;
+    }
+
     if ((commandName === 'refresh' || commandName === 'url') && !error) {
       try {
         // This is only meant as a convenience so failure is not particularly concerning.
--- a/packages/terra-functional-testing/src/services/wdio-terra-service.js
+++ b/packages/terra-functional-testing/src/services/wdio-terra-service.js
@@ -4,7 +4,6 @@ const { accessibility, element } = require('../commands/validates');
 const { toBeAccessible } = require('../commands/expect');
 const {
   describeViewports,
-  dispatchCustomEvent,
   getViewports,
   hideInputCaret,
   setViewport,
@@ -61,7 +60,6 @@ class TerraService {
     // Setup and expose global utility functions.
     global.Terra.viewports = getViewports;
     global.Terra.hideInputCaret = hideInputCaret;
-    global.Terra.dispatchCustomEvent = dispatchCustomEvent;
 
     // Setup and expose the validates utility functions.
     global.Terra.validates = { accessibility, element };
--- a/packages/terra-functional-testing/src/services/wdio-terra-service.js
+++ b/packages/terra-functional-testing/src/services/wdio-terra-service.js
@@ -98,12 +98,6 @@ class TerraService {
   }
 
   afterCommand(commandName, _args, _result, error) {
-    // Expose flag to disable hidden input caret function for full stack testing; IE cannot handle multiple async afterCommand requests.
-    const { disabledTerraServiceAfterCommand } = this.serviceOptions;
-    if (disabledTerraServiceAfterCommand) {
-      return;
-    }
-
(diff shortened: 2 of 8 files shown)


packages/terra-functional-testing/src/commands/utils/dispatchCustomEvent.js
```diff
@@ -0,0 +1,15 @@
+const dispatchCustomEvent = (name, metaData) => {
```
yuderekyu: i wasn't sure where to put this file. opted to put this in `commands/utils`, even though it is not attached to the terra global. open to suggestions here

StephenEsser: My memory is a little fuzzy on the details/use-cases of this utility function. From a fresh perspective it seems a little strange to have this independent utility function that's not used/exposed from this project. Should this be attached to a global/browser action? Here are a few of the possible approaches, I'm mainly just curious to get thoughts on which one we think makes the most sense

Custom Browser Command:

https://webdriver.io/docs/browserobject.html

```js
browser.dispatchEvent();
```

Custom Page Object:

https://webdriver.io/docs/pageobjects.html

```js
EventPage.dispatchEvent();
```

Bound to the Terra global similar to other utility functions:

```js
Terra.dispatchEvent();
```

An independent utility function:

```js
const dispatchCustomEvent = require('@cerner/terra-function-testing/lib/commands/utils/dispatchCustomEvent');

dispatchCustomEvent();
```

StephenEsser: I think I lean towards the `Terra.dispatchEvent` because we'll own that namespace and it is provided globally to the test run. It wouldn't require a custom import into each spec that needs it, but if it should only be used by in the full stack project (consumers aren't expected to dispatch custom events) then it might make sense to keep it independent

yuderekyu: good points - I agree that placing this function here without being exposed is strange. I don't see consumers calling this themselves; full stack uses this function in its after command hook to send a locale override event, to override a terra application's locale. In that sense, I'd lean towards keeping it independent.

Do we want the minimal speed benefit of being tied to open source release process? we could move this back into full stack testing, it's small. atm, the only open source use case is for terra-application to import this for testing purposes.

anyone else have thoughts?

yuderekyu: attaching this to the terra global; based on offline discussion
https://github.com/cerner/terra-toolkit/pull/519/commits/9d100d6f5bb139845718b9f6dcc072af1eef24c4

packages/terra-functional-testing/src/commands/utils/dispatchCustomEvent.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Utility to send a custom event containing metadata.
+ * @param {string} options.name - name of event
+ * @param {object} options.metaData - metadata pertaining to event
+ */
+const dispatchCustomEvent = (options) => {
```
StephenEsser, line 6: This function might benefit from splitting the params:

```js
dispatchCustomEvent(name, metadata);
```

It simplifies the following:

```js
  const event = {
    name: 'applicationBase.testOverride',
    metaData: { locale: 'ar' }
  };
  
  Terra.dispatchCustomEvent(event);
  ```
  vs
  ```js
  Terra.dispatchCustomEvent('applicationBase.testOverride',  { locale: 'ar' } );
  ```
  
That said, this isn't a necessary change. If there is a chance there will be any additional options in the future the current API makes that easier to integrate.

File: packages/terra-functional-testing/src/services/wdio-terra-service.js [hunk 3]
```diff
@@ -58,6 +59,7 @@ class TerraService {
     global.expect.extend({ toBeAccessible });
 
     // Setup and expose global utility functions.
+    global.Terra.dispatchCustomEvent = dispatchCustomEvent;
```
StephenEsser, line 62: I'd personally probably rename this to `dispatchEvent`, but these are custom events so the name is descriptive as is without changes.

StephenEsser (approved):

packages/terra-functional-testing/src/commands/utils/dispatchCustomEvent.js
yuderekyu: do we want explicit named parameters or save a couple lines? i do like knowing what each parameter explicitly is. overall, i feel meh either way

packages/terra-functional-testing/src/services/wdio-terra-service.js
yuderekyu: yeah i think that's confusing. if someone wants to dispatch a regular event, they should just use the first class web api: `eventTarget.dispatchEvent`. the convenience of using this utility is firing a non pollyfilled custom event that's compatible with ie

packages/terra-functional-testing/src/commands/utils/dispatchCustomEvent.js
StephenEsser: I'd say keep it as is if there's not a strong argument to change it. I'm a little unsure about `metadata` vs `metaData`, but that's just a nit pick. Either one is probably technically correct.

> do we want explicit named parameters or save a couple lines? i do like knowing what each parameter explicitly is.

I think there are use-cases for both scenarios, but I think we should be cautious about falling into the mindset that each function we create needs an object API instead of params.

packages/terra-functional-testing/src/services/wdio-terra-service.js
StephenEsser: There are [CustomEvents](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent) in JavaScript so it makes both function names slightly confusing. I don't have any objections to keeping this as `dispatchCustomEvent`, it's bound to the `Terra` global so either function name is namespace safe

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
```diff
@@ -159,3 +159,19 @@ it('should hide the caret', () => {
...
+Fire a custom event with `metadata`. Use this to manually update a deployed terra application's locale, or fire updates for a custom event listener. This method accepts an object containing the event `name` (string) and event `metaData` (object).
+
+```js
+it('should update an application's locale, () => {
+  const event = {
+    name: 'applicationBase.testOverride',
+    metaData: { locale: 'ar' }
+  };
+
+  Terra.dispatchCustomEvent(event);
```
StephenEsser, line 175: After reading this documentation it makes me wonder if there would be any benefit to exposing an explicit function that switches the locale. It would abstract away the `dispatchCustomEvent`. (Or creating both)

```js
Terra.setLocale('en');
```

packages/terra-functional-testing/src/commands/utils/dispatchCustomEvent.js
yuderekyu: good points - unless people have strong opinions, i'll follow the capitalization we've established in framework/application and keep it `metaData`

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
yuderekyu: I like that idea, seeing how this utility was created for that purpose. we could expose a `setLocale` only? if someone requests `dispatchCustomEvent`, we could expose that later

yuderekyu:
closing in favor of stephen's api suggestion https://github.com/cerner/terra-toolkit/issues/523

PR #535, opened 2021-01-19
[terra-functional-testing] Add custom spec reporter

### LINKED ISSUE #527: [terra-functional-testing] Add custom spec reporter
# Story

## Description

As a consumer of terra I want to create a [custom reporter](https://webdriver.io/docs/customreporter.html) that logs test output to a JSON file.

The test output should include:

- Capabilities of the test run
- Locale of the test run
- Theme of the test run
- Form factor of the test run (if applicable)
- Date/time of spec start and end times
- Meta data for each test
  - Title of the test
  - Start/end time
  - Duration
  - State of the test (pass, fail)
  - Error output (if applicable)

## Motivation

So that downstream build processes can consume and analyze collective test output.

## Definition of Done

We'll know we're done when a custom reporter and documentation has been merged to main.

StephenEsser:
### Summary

This PR adds a [custom](https://webdriver.io/docs/customreporter.html) spec reporter that logs the spec results to an output JSON file.

Example output:

https://gist.github.com/StephenEsser/6aa28588b17c9e313307133324221647

Resolves https://github.com/cerner/terra-toolkit/issues/527

### Additional Details

This PR will be converted from a draft after tests and documentation have been added. 

There were breaking changes regarding reporters in WDIO 5 and 6 to help accommodate running tests in parallel. Part of those changes included changing how the spec reporter is initialized. Previously, in WDIO 4, a single spec reporter was created for the entire test run. Starting in WDIO 5 reporters have been updated to create a reporter per each spec file. 

This new reporter strategy requires writing a JSON file per spec file and merging those files together in the onComplete hook of the test run. The [wdio-json-reporter](https://webdriver.io/docs/docs/wdio-json-reporter.html#result-files) provides a good explanation of the changes and demonstrates merging individual files together. 

The lifecycle of this custom reporter looks like this:

- In the onPrepare hook remove any stale result files
- Write a JSON report to file for each spec file
- In the onComplete hook merge those spec result files together and remove the individual files

@cerner/terra


packages/terra-functional-testing/src/reporters/spec-reporter/clean-results.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param {string} outputDir - The output directory to clean.
+ */
+const cleanResults = (outputDir) => {
+  const directory = outputDir || getOutputDir();
+
+  if (!fs.existsSync(directory)) {
+    return;
+  }
+
+  const files = fs.readdirSync(directory);
```
emilyrohrbough: Can we make this async since it's just deleting files?

StephenEsser: I had written this async originally, but there didn't seem to be many advantages so I ended up swapping it back. At this point in the lifecycle there is only one process and it has to wait for these files to be deleted before moving on to the next lifecycle hook. I'm a little on the fence, I think async in general is better practice , but there are some code complexity trade offs (the hooks will need to await before proceeding)

packages/terra-functional-testing/src/config/wdio.conf.js
```diff
@@ -157,12 +159,43 @@ exports.config = {
...
+   * Gets executed once before all workers get launched.
+   * @param {Object} config - wdio configuration object.
+   * @param {Array.<Object>} capabilities - List of capabilities details.
+   */
+  onPrepare(config) {
+    const { reporterOptions = {} } = config;
+    const { outputDir } = reporterOptions;
+
+    // Remove previous reporter results.
+    cleanResults(outputDir);
```
emilyrohrbough: Can this be moved into the reporter's initialization? or add a `onRunnerStart` listener to do this?

packages/terra-functional-testing/src/config/wdio.conf.js
```diff
@@ -157,12 +159,43 @@ exports.config = {
...
+   * @param {Object} config - wdio configuration object.
+   * @param {Array.<Object>} capabilities - List of capabilities details.
+   * @param {<Object>} results - Object containing test results.
+   */
+  onComplete(_exitCode, config) {
+    const { reporterOptions = {} } = config;
+    const { outputDir } = reporterOptions;
+
+    // Merge reporter results.
+    mergeResults(outputDir, {
```
emilyrohrbough: Could you add an event listen for `onRunnerEnd`  to manage this would needing to add it in the wdio config's onComplete hook?

packages/terra-functional-testing/src/config/wdio.conf.js
StephenEsser: No, a reporter is initialized for each spec file so it would be invoked multiple times (once for each spec file in a project). It needs to happen once before the worker processes start. This utility removes all the previous spec result files. This isn't always necessary as the onComplete hook will try to cleanup stale reporters results, but to be safe and ensure a clean test run the files are cleaned before starting

packages/terra-functional-testing/src/config/wdio.conf.js
StephenEsser: No, because of how WDIO 6 initializes the reporters this work needs to happen in the configuration lifecycle hooks. In WDIO 6 a reporter is created per spec file. So any work done in a reporters `onRunnerEnd` lifecycle hooks runs per spec file, not per test run. So if there are 10 spec files, `onRunnerEnd` will be run 10 times with 10 different instances of a reporter.

The [json reporter](https://webdriver.io/docs/docs/wdio-json-reporter.html#result-files) has a pretty good description of this:

> With WDIO v5, reporting has moved from a centralized process to one that is handled by each of the "sessions" spun up for parallel test execution.
This change helped reduce the amount of chatter during WDIO test execution and thus improved performance. The downside is we are no longer able to get a single report for ALL test execution. Consider the following:

> 2 suites of tests configured to run in 2 browsers:

> WDIO v4: 1 json file with execution results
> WDIO v5: 4 json files with execution results

packages/terra-functional-testing/src/config/wdio.conf.js
mjhenkes: This would add complexity, but could you use a singleton in the reporter to cleanResults for a dir? It might be cleaner from a config perspective, but might run into concurrency issues.

packages/terra-functional-testing/src/reporters/spec-reporter/get-output-dir.js
```diff
@@ -0,0 +1,18 @@
...
+const path = require('path');
+
+/**
+ * Determines the output directory for reporter results.
+ * @returns {string} - An output directory.
+ */
+const getOutputDir = () => {
+  const cwd = process.cwd();
+
+  if (fs.existsSync(path.join(cwd, 'test'))) {
```
mjhenkes: are we for sure keeping the tests/test thing? IMO i'd default one and let people config if absolutely necessary

packages/terra-functional-testing/src/reporters/spec-reporter/merge-results.js
```diff
@@ -0,0 +1,118 @@
...
+    browser,
+    formFactor,
+    locale,
+    packageName,
+    theme,
+  } = options;
+
+  const keys = [locale, theme, formFactor, browser, packageName].filter(key => !!key);
+
+  return `results-${keys.join('-')}.json`;
```
mjhenkes, line 39: since this doesn't need to be human readable, you could hash the keys string: https://nodejs.org/docs/v0.6.18/api/crypto.html#crypto_crypto_createhash_algorithm

or just use a uuid

mjhenkes: Should we make the 'results' key more unique to this reporter?

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-reporters/SpecReporter.tool.mdx
```diff
@@ -0,0 +1,170 @@
+# Spec Reporter
+
+The Spec Reporter writes test results to file in JSON format.
+
+## Installation
+
+The spec reporter should be installed as a development dependency.
```
mjhenkes: The spec reporter is part of the default wdio config. Should we document it as such? Do we ever expect people to use it by itself?

mjhenkes: Or maybe we focus the documentation on the default config and have this doc in an advanced section since i don't think the majority of our users will need it?

StephenEsser: There's a note below:

> Note: The spec reporter is enabled by default in the `terra-functional-testing` wdio configuration. If you are using/extending the default `wdio.conf.js` it is not necessary to reimport the reporter.

But I had similar concerns about the documentation. Distinguishing between a stand alone reporter and how to interact with the reporter when using the test runner and extending the wdio.confg.js.

I think we can either have two different pages of documentation or we rework this one to be from the perspective of the test runner.

I'll take a swing at the latter and add an advance section at the bottom for a stand-alone reporter

packages/terra-functional-testing/src/reporters/spec-reporter/get-output-dir.js
StephenEsser: Removed dynamic test directory logic. The function returns `tests` now. A user can configure this using the `reporterOptions` in the wdio config.

https://github.com/cerner/terra-toolkit/pull/535/commits/88f0eac64b0f4c17fc2d2da031653fd0be66b18c

packages/terra-functional-testing/src/reporters/spec-reporter/merge-results.js
StephenEsser: I don't have a strong opinion. The current naming convention is aligning with the WDIO 4 spec reporter. I don't think the file names matter too much other than someone looking through the files manually (which might be the case on Jenkins artifacts). I'll defer to @ryanthemanuel 

If we do use a UUID we will need to clear the entire output directory before starting the test run to ensure all stale reports are cleared

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-reporters/SpecReporter.tool.mdx
StephenEsser: I [reorganized](https://github.com/cerner/terra-toolkit/pull/535/commits/269b822d40792f8ccfc0bf7c8c35656298bb4345) the documentation, but then decided to just go ahead and remove the advanced section all together:

https://github.com/cerner/terra-toolkit/pull/535/commits/9db62fa3c3ead3ec0e28016e13a05b22c77486b2

packages/terra-functional-testing/src/reporters/spec-reporter/merge-results.js
mjhenkes: I'm thinking `terra-spec-reporter-<uuid>`. so you have a key you can use to clear, it won't conflict with other possible future reporters and the name *may* be shorter than if you use the full text.

ryanthemanuel: When we run the reporter, we'll want to be able to "load all files" and merge them together.  I think we could do that if they're all in the same directory or something even with uuids. The thing to remember is the flow in jenkins will be: spawn off all of our different possible combinations of locales, form factors, etc. these will have their own files that then need to be archived off for use later and *cannot* overwrite each other. I then think it is ok for orion-release-site to read them all in and merge them together. There will need to be enough data contained in the file to make sure that we know exactly what the contents were testing though.

packages/terra-functional-testing/src/config/wdio.conf.js
StephenEsser: I spiked into the singleton idea a little more, I think for now my vote is not to implement the singleton. I believe there could be some concurrency issues that would be challenging to debug and ensure consistency.

mjhenkes (approved):

packages/terra-functional-testing/src/reporters/spec-reporter/wdio-spec-reporter.js
```diff
@@ -0,0 +1,147 @@
...
+    // The root suite is always at index 0 and is guaranteed to exist.
+    const rootSuite = this.currentSuites[0];
+    const results = SpecReporter.formatResults(rootSuite, runner);
+
+    SpecReporter.writeResults(runner, results);
+  }
+
+  /**
+   * Formats the test runner spec results.
+   * @param {SuiteStats} runner - The root suite object.
```
benbcai:
```suggestion
   * @param {SuiteStats} rootSuite - The root suite object.
```

packages/terra-functional-testing/src/reporters/spec-reporter/wdio-spec-reporter.js
```diff
@@ -0,0 +1,147 @@
...
+  static formatTests(tests) {
+    return tests.map((test) => {
+      const {
+        title,
+        fullTitle,
+        duration,
+        start,
+        end,
+        state,
+        error,
```
benbcai, line 98: The [error message/stack](https://gist.github.com/StephenEsser/6aa28588b17c9e313307133324221647#file-results-en-terra-default-theme-chrome-terra-functional-testing-json-L90-L91) in the sample output seems hard to read. Is it just a bad example or there is some formatting issue?

StephenEsser: It's formatted chalk'ed output so most of what you're seeing are special unicode characters. For now I think this is serialized as expected assuming it will be displayed in a terminal output. It would need to be sanitized if displayed outside of the terminal/outside of an environment that supports the characters 

Edit:

For reference. Here's the output of console logging the values linked above to the terminal:

```
Message:

 expect(received).toEqual(expected) // deep equality

Expected: false
Received: true


Stack:

Error: expect(received).toEqual(expected) // deep equality

Expected: false
Received: true
    at Context.<anonymous> (/Users/se028266/Github/terra/terra-toolkit/packages/terra-functional-testing/tests/wdio/terra-validates-spec.js:33:18)
```

Picture to demo colors:

image

benbcai (approved):

packages/terra-functional-testing/src/reporters/spec-reporter/wdio-spec-reporter.js
StephenEsser: Updated:

https://github.com/cerner/terra-toolkit/pull/535/commits/706d9de5b0cbbc5eb544214af903a2e32c0dc9ae

ryanthemanuel (approved):

PR #536, opened 2021-01-20
Added browser to mainFields

### LINKED ISSUE #522: [webpack-config-terra] Investigate adding 'browser' to the webpack modules list
# Feature Request

## Description

Webpack has a field ([resolve.mainFields](https://webpack.js.org/configuration/resolve/#resolvemainfields)) that allows you to specify the precedence of which fields to look for in the package.json file for an entry point into packages. Currently, our webpack configuration limits this field to only looking at [main](https://github.com/cerner/terra-toolkit/blob/main/packages/webpack-config-terra/src/webpack.config.js#L230). 

I would like to investigate adding 'browser' to the beginning of the list so that it takes precedence over 'main'. This can be accomplished by building a test tgz of our the webpack-config-terra package, adding it to a project's base directory, and then referencing that in those package's package.json and verifying that there are no problems. We could even issue a PR into those repos with that change to verify that the builds are still succeeding. I'd like to do this for our open source repos as well as some of our internal MPage Integration repos. Then, we can track any problems that we might run into on this issue and then we can decide if we want to make that change.

One main reason I want to look into this, is I've run into a few packages recently where I think 'browser' is more appropriate than 'main'. One example is newer versions of the [uuid package](https://github.com/uuidjs/uuid/blob/master/package.json#L33). These versions will actually break in older browsers due to its use of more modern syntax (e.g. const) in the 'main' package' where it uses older syntax (e.g. var) in the 'browser' package.

pranav300:
### Summary

This PR adds `browser` to `resolve.mainFields` in webpack.

closes #522 

These changes have been tested on Framework.

@cerner/terra

### DIFF
--- a/packages/webpack-config-terra/CHANGELOG.md
+++ b/packages/webpack-config-terra/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 * Added
-  * Added `browser` to webpack module list.
+  * Added `browser` to `resolve.mainFields` in webpack module list.
 
 ## 1.2.0 - (January 5, 2021)
 
--- a/packages/webpack-config-terra/CHANGELOG.md
+++ b/packages/webpack-config-terra/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added `browser` to webpack module list.
+
 ## 1.2.0 - (January 5, 2021)
 
 * Changed
--- a/packages/webpack-config-terra/src/webpack.config.js
+++ b/packages/webpack-config-terra/src/webpack.config.js
@@ -227,7 +227,7 @@ const defaultWebpackConfig = (env = {}, argv = {}, options = {}) => {
     resolve: {
       extensions: ['.js', '.jsx'],
       modules: getResolveModules(env),
-      mainFields: ['main'],
+      mainFields: ['browser', 'main'],
     },
     output: {
       filename: `${filename}.js`,


packages/webpack-config-terra/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added `browser` to webpack module list.
```
emilyrohrbough: Can you expand on this changelog on how this changes the configuration?

packages/webpack-config-terra/src/webpack.config.js
```diff
@@ -227,7 +227,7 @@ const defaultWebpackConfig = (env = {}, argv = {}, options = {}) => {
     resolve: {
       extensions: ['.js', '.jsx'],
       modules: getResolveModules(env),
-      mainFields: ['main'],
+      mainFields: ['browser', 'main'],
```
StephenEsser, line 230: I think these changes make sense but we should confirm this is passive (it's probably not technically passive, but might still be okay pending issues we see), in our repositories. Do you have branches/draft PRs that consume this change to validate it doesn't break the builds? (terra-framework, terra-application, etc...)

pranav300: Yes, this change has been validated in Framework [PR 1353](https://github.com/cerner/terra-framework/pull/1353).

mjhenkes (approved):

StephenEsser (approved):

ryanthemanuel:
I think we're still validating some repos internally.  Added the do not merge label until that is done. @pranav300 let us know when you think everything looks good.

pranav300:
I think we are good to go.
As all the open-source repos have been tested. And draft PR's are out:
1. [terra-application](https://github.com/cerner/terra-application/pull/129)
2. [terra-ui](https://github.com/cerner/terra-ui/pull/327)
3. [terra-framework](https://github.com/cerner/terra-framework/pull/1353)

Note: `terra-dev-site` could not be upgraded as the `postcss-rtl` doesn't support `postcss v8.x.x`. And we need terra-toolkit-boneyard config in dev-site export config. So, I guess we will have to wait for this [PR](https://github.com/vkalinichev/postcss-rtl/pull/73) to get merged or use it the way @mhenkes has done it in `webpack-config-terra`. 
Any thoughts on this? @ryanthemanuel @mjhenkes 

And regarding workflow-server and dynamic-component-catalog-toolkit, I have raised draft PRs for upgrading them to webpack-config -terra.

ryanthemanuel (approved):

PR #545, opened 2021-01-25
[terra-functional-testing] Add screenshot captures

### LINKED ISSUE #471: [terra-functional-testing] Simplify Screenshot Naming
# Feature Request

## Description
Fix screenshot naming to remove naming after describe block and always require a name to be provided.

### Current Behavior
```js
// my-awesome-test-spec.js
describe('My Test', () => {
  it('does something grand', () => {
     Terra.validates.element();
  });

  it('does something grand x2', () => {
     Terra.validates.element('x2');
  });
})
```
results in:
```
wdio/__snapshots__/reference/en/chrome_huge/my-awesome-test-spec/My_Test['default'].png
wdio/__snapshots__/reference/en/chrome_huge/my-awesome-test-spec/My_Test['x2'].png
```

### Updated Behavior
```diff
// my-awesome-test-spec.js
describe('My Test', () => {
  it('does something grand', () => {
-    Terra.validates.element();
+    Terra.validates.element('default');
  });

  it('does something grand x2', () => {
     Terra.validates.element('x2');
  });
})
```
results in:
```
wdio/__snapshots__/reference/en/chrome_huge/my-awesome-test-spec/default.png
wdio/__snapshots__/reference/en/chrome_huge/my-awesome-test-spec/x2.png
```

### LINKED ISSUE #526: [terra-functional-testing] Add command to capture screenshot
# Story

## Description

As a consumer of Terra I want a command to capture screenshot at any point during a wdio test step. This is a story for feature #401.

## Motivation

So that the captured screenshots can be used as new reference screenshots or compared with existing reference screenshots to ensure the they match visually.

## Definition of Done

We'll know we're done when 
- A PR is issued for creating the screenshot command to capture and generate screenshots used either as a reference or to match with existing screenshots.
- Documentations for the screenshot command are created.
- Tests are written to test the screenshot command.

benbcai:
### Summary

Added the `Terra.validates.screenshot` command to use the visual regression service for capturing and comparing screenshots.

Closes #526 
Closes #471 

### Additional Details
- Added `Terra.validates.screenshot` command
- Updated `Terra.validates.element` command to include capturing and comparing screenshots.
- Both `Terra.validates.screenshot` and `Terra.validates.element` now require a screenshot name.
- The title in the `describe` block will no longer be prepended to the screenshot file name.
- Added `ignoreComparisonResults` CLI option to ignore screenshot comparison. The option will allow screenshots to be captured and tests will pass even if there are mismatches.
- A folder for the `terra-default-theme` theme will be created to store the screenshots similar to other themes.
- The `Terra.validates.screenshot` command will no longer accept a list of viewports to capture screenshots.
- Aggregating the terra service options in TerraService is moved from the `onPrepare` hook to the `beforeSession` hook. The reason being that `onPrepare` gets executed only once and the local `this.serviceOptions` variable gets reset in the constructor by each test.

@cerner/terra


packages/terra-functional-testing/src/commands/validates/screenshot.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * The screenshot comparisons.  It will capture screenshots of a specified element
+ * and assert the screenshot comparison results are within the mismatch tolerance or are an exact match
+ *
+ * This should be used within an `it` block.
+ *
+ * @param {[name, options]} [args] - the list of test arguments to parse.
+ * @param {string} [testName] - the required test case name.
+ * @param {Object} [options] - the visual regression test options. Options include viewports and selector
```
mjhenkes:
```suggestion
 * @param {Object} [options] - the visual regression test options. Options include mismatchTolerance and selector
```

packages/terra-functional-testing/src/services/wdio-visual-regression-service/VisualRegressionLauncher.js
```diff
@@ -1,12 +1,10 @@
 /* global browser */
-import _ from 'lodash';
-import { parse as parsePlatform } from 'platform';
-
-import { LocalCompare } from './compare';
-import makeElementScreenshot from './modules/makeElementScreenshot';
-
-import getUserAgent from './scripts/getUserAgent';
-import getTerraFormFactor from './modules/getTerraFormFactor';
+const _ = require('lodash');
```
mjhenkes: instead of pulling in all of lodash, could we just pull in what we need?

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
```diff
@@ -126,13 +127,48 @@ it('should override a rule configuration', () => {
...
+
+The screenshot naming pattern is as follows:
+
+```js
+baseScreenshotDir/test_spec_directory/test_spec_name/__snapshots__/(reference|latest|diff)/theme/locale/browserName_terraViewportName/screenshot-name.png
+```
+
+The screenshot assertion accepts an optional second argument with the following keys.
+  * `selector` - the element selector to use to capture the element for screenshot comparison. The default selector is `data-terra-test-content`.
+  * `misMatchTolerance` - the number between 0 and 100 that defines the degree of mismatch to consider two images as identical. Increasing this value will decrease level of test coverage.
```
mjhenkes:
```suggestion
  * `mismatchTolerance` - the number between 0 and 100 that defines the degree of mismatch to consider two images as identical. Increasing this value will decrease level of test coverage.
```

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
```diff
@@ -126,13 +127,48 @@ it('should override a rule configuration', () => {
...
+
+```js
+it('should validate the screenshot with options', () => {
+  browser.url('/testing/route/');
+
+  const options = {
+    selector: '#element-id',
+    misMatchTolerance: 10,
```
mjhenkes:
```suggestion
    mismatchTolerance: 10,
```

mjhenkes (approved):
Looks good to me with a couple of minor updates.

packages/terra-functional-testing/src/commands/validates/screenshot.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+const screenshot = (testName, options = {}) => {
+  if (!testName || typeof testName !== 'string' || testName.length === 0) {
+    throw new Error('[terra-functional-testing:screenshot] Terra.validate.screenshot requires a unique test name as the first argument.');
+  }
+
+  const { selector } = options;
+  const wrappedOptions = {
+    name: testName,
+    ...options,
```
emilyrohrbough, line 20: would this only be misMatchTolerance? passed in `....options`?

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -16,25 +16,23 @@ class TerraService {
...
   /**
-   * Service hook executed before worker processes are launched.
+   * Service hook executed prior to initializing the webdriver session.
    * @param {object} config - The WebdriverIO configuration object.
    */
-  onPrepare(config) {
-    const { serviceOptions } = config;
+  beforeSession(config) {
+    global.Terra = {};
+    const { serviceOptions, ignoreComparisonResults } = config;
```
emilyrohrbough: where is `ignoreComparisonResults` coming from?

packages/terra-functional-testing/src/services/wdio-visual-regression-service/VisualRegressionLauncher.js
benbcai: Updated to use only needed lodash modules: https://github.com/cerner/terra-toolkit/pull/545/commits/90db6b082ead33e7514b4178e5d509cf8f1cf39c

packages/terra-functional-testing/src/services/wdio-visual-regression-service/VisualRegressionLauncher.js
```diff
@@ -29,16 +28,17 @@ class VisualRegressionLauncher {
...
-    const { name, version, ua } = parsePlatform(userAgent);
+    const { name, version, ua } = parse(userAgent);
 
     this.context = {
       browserInfo: {
         name,
         version,
         userAgent: ua,
       },
+      desiredCapabilities: capabilities,
```
emilyrohrbough, line 30: Since you added `desiredCapabilities` can we remove the `browserInfo` from the context?

I thought browserInfo.name was the only info being used in the screenshot naming pattern & you replaced that reference: 
https://github.com/cerner/terra-toolkit/pull/545/files#diff-dea71618479daf3d0e21b24c6f7201de0dc18481e558caf0c3aee745a434af06L90

packages/terra-functional-testing/src/services/wdio-visual-regression-service/VisualRegressionLauncher.js
```diff
@@ -104,14 +104,16 @@ class VisualRegressionLauncher {
 
       const screenshotContext = {
         browserInfo: this.context.browserInfo,
+        desiredCapabilities: this.context.desiredCapabilities,
         suite: this.currentSuite,
         test: this.currentTest,
         meta: {
           currentFormFactor,
         },
+        options,
```
emilyrohrbough, line 102: these options are passed to the screenshot command on line 116, do they need to be added to the screenshot context as well?

packages/terra-functional-testing/src/services/wdio-visual-regression-service/methods/BaseCompare.js
```diff
@@ -69,25 +69,25 @@ export default class BaseCompare {
    */
   getScreenshotName(context) {
     const { name } = (context.options || {});
-    const parentName = this.createTestName(context.test.parent);
```
emilyrohrbough, line 72: 🎉

packages/terra-functional-testing/src/services/wdio-visual-regression-service/methods/BaseCompare.js
```diff
@@ -69,25 +69,25 @@ export default class BaseCompare {
...
     const testName = this.createTestName(name || context.test.title);
```
emilyrohrbough: Since we are requiring the name to be provided, I think we can remove reference to context.test.title as well

packages/terra-functional-testing/src/services/wdio-visual-regression-service/modules/getTerraFormFactor.js
```diff
@@ -1,20 +1,45 @@
...
     const [formFactor, size] = viewports[index];
 
-    if (viewpointWidth <= size.width) {
+    if (currentWidth <= size.width) {
       return formFactor;
     }
   }
 
-  return 'enormous';
+  return 'huge';
```
emilyrohrbough: The largest viewport size is enormous without a max width. I think this should still be returned form factor based on the comparison logic

packages/terra-functional-testing/src/config/wdio.conf.js
```diff
@@ -137,6 +138,11 @@ exports.config = {
       ...WDIO_INTERNAL_PORT && { port: WDIO_INTERNAL_PORT },
       ...fs.existsSync(defaultWebpackPath) && { webpackConfig: defaultWebpackPath },
     }],
+    [VisualRegressionLauncher, {
+      baseScreenshotDir: process.cwd(),
```
emilyrohrbough: Do we need to maintain the baseScreenshotDir override that existed in the last version?

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
benbcai: Update misMatch to mismatch. https://github.com/cerner/terra-toolkit/pull/545/commits/509b510544d67acf189d8d55a18aa91a27ad3219

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
benbcai: Update misMatch to mismatch. https://github.com/cerner/terra-toolkit/pull/545/commits/509b510544d67acf189d8d55a18aa91a27ad3219

packages/terra-functional-testing/src/services/wdio-visual-regression-service/VisualRegressionLauncher.js
benbcai: The reason I added `desireCapabilities` is to get the lower case browser name (e.g. `chrome`) instead of `Chrome` from `browserInfo.name`. Yes, you're right. `browserInfo` was needed only for the browserName so we no longer need to add `browserInfo` to the context. I also removed `getUserAgent` and the `platform` dependency since they were only used to get the user agent data for `browserInfo`.

https://github.com/cerner/terra-toolkit/pull/545/commits/9933d1afc047d7a96ebbc454f3b716a0b8a9a420

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: `ignoreComparisonResults` used to be set as an ENV in wdio4. It is now being set as a CLI option: https://github.com/cerner/terra-toolkit/pull/545/files#diff-b821c3a083e050c54034354280dac343f0b8d9ad66c460cad80a99ec7ebea640

packages/terra-functional-testing/src/commands/validates/screenshot.js
benbcai: It can be `mismatchTolerance` and/or `selector`: https://github.com/cerner/terra-toolkit/pull/545/files#diff-0aa7dac6eaa779fb29a0b1ac99068b18cbf2c0a6ecbc1db503237212266a8f68R23

In wdio4 we used to also allow an array of viewports but we decided to no longer support that.

packages/terra-functional-testing/src/services/wdio-visual-regression-service/VisualRegressionLauncher.js
benbcai: The reason I added options to the context is to get `options.name` to use for the screenshot name. It currently creates the screenshot name using the context in BaseCompare: https://github.com/cerner/terra-toolkit/pull/545/files#diff-dea71618479daf3d0e21b24c6f7201de0dc18481e558caf0c3aee745a434af06R71

packages/terra-functional-testing/src/services/wdio-visual-regression-service/methods/BaseCompare.js
benbcai: `context.test.title` is no longer referenced: https://github.com/cerner/terra-toolkit/pull/545/commits/b85221c60247a8a943bdf916e2fb00df6cee06f7

packages/terra-functional-testing/src/services/wdio-visual-regression-service/modules/getTerraFormFactor.js
benbcai: That makes sense to me. Changed it back to `enormous`: https://github.com/cerner/terra-toolkit/commit/b85221c60247a8a943bdf916e2fb00df6cee06f7

My initial thinking was to default it to `huge` just like what we're doing [here](https://github.com/cerner/terra-toolkit/pull/545/files#diff-a6ba3c1d16c08e58e7ee3994f3ab541b63b2a3a8dafb0de9daed2c8090c948d9R94).

packages/terra-functional-testing/src/config/wdio.conf.js
benbcai: Ahh...I didn't know we currently allow this to be overridden [here](https://github.com/cerner/terra-toolkit-boneyard/blob/ea4a5fc159f8e5f71911077ef613259694940e47/config/wdio/visualRegressionConf.js#L69). Looks like this is currently only used in our theme repos. We can talk during our next sync to decided whether or not to allow the override.

packages/terra-functional-testing/src/commands/validates/screenshot.js
emilyrohrbough: ahh I was looking at the code below on line 24 which passes the selector directly to checkElement

packages/terra-functional-testing/src/config/wdio.conf.js
emilyrohrbough: We are also using this to run grid tests on Terra internally

packages/terra-functional-testing/src/services/wdio-terra-service.js
emilyrohrbough: Ahh yeah I had saw it in the CLI but didn't realize it was passed through to the service via the wdio CLI

packages/terra-functional-testing/src/services/wdio-visual-regression-service/VisualRegressionLauncher.js
emilyrohrbough: Gottchya that makes sense. Could we just pass the name? no preference there
Can we add `name` to the docs?

packages/terra-functional-testing/src/commands/expect/toMatchReference.js
```diff
@@ -0,0 +1,51 @@
...
+ * @param {boolean} screenshot.isExactSameImage - If the latest screenshot matched the reference screenshot exactly, i.e. 0% mismatch.
+ * @param {Number} screenshot.misMatchPercentage - The mismatch percentage when comparing the latest screenshot to the reference screenshot.
+ * @returns {String} - A message describing the status of the screenshot.
+ */
+function getComparisonResult(screenshot) {
+  const { misMatchPercentage, isSameDimensions } = screenshot;
+  const relevantInformation = {};
+
+  if (!isSameDimensions) {
+    relevantInformation.isSameDimensions = isSameDimensions;
```
StephenEsser: Curious about this logic assignment. Assign isSameDimensions only if `false`? True doesn't getting added into the property list?

ryanthemanuel (approved):

packages/terra-functional-testing/src/commands/expect/toMatchReference.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  relevantInformation.misMatchPercentage = misMatchPercentage;
+
+  return `${JSON.stringify(relevantInformation, null, 2)}`;
+}
```
StephenEsser: nit: (These aren't necessary changes, just suggestions)

This function is heavy for what is essentially just formatting a string. We could make this more human friendly by formatting the expect message sentence.

Mismatch:

> expected the screenshot to be within the mismatch tolerance, but received a mismatch difference of 0.23%.

Dimension Changes

> expected the screenshot to match reference, but received a screenshot with different dimensions

Edit:

If a diff screenshot is generated we could also include that in the message. For some reason a lot of users don't know the diffs exist.

packages/terra-functional-testing/src/commands/expect/toMatchReference.js
```diff
@@ -0,0 +1,51 @@
...
+ * the mismatch tolerance and are the same size.
+ *
+ * @param {Object} screenshot - The comparison result
+ * @param {boolean} screenshot.isSameDimensions - If the latest screenshot was the same size as the reference screenshot.
+ * @param {boolean} screenshot.isWithinMismatchTolerance - If the latest screenshot was within the mismatch tolerance.
+ * @param {boolean} screenshot.isExactSameImage - If the latest screenshot matched the reference screenshot exactly, i.e. 0% mismatch.
+ * @param {Number} screenshot.misMatchPercentage - The mismatch percentage when comparing the latest screenshot to the reference screenshot.
+ * @returns {Object} - An object that indicates if the assertion passed or failed with a message.
+ */
+function toMatchReference(receivedScreenshot) {
```
StephenEsser: nit: Param documentation doesn't align. `screenshot`/`receivedScreenshot `. 

I'm on the fence about documenting values that aren't used, but it doesn't hurt, it's just a little confusing at first. Several values aren't used in some of these functions.

packages/terra-functional-testing/src/commands/validates/screenshot.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * The screenshot comparisons.  It will capture screenshots of a specified element
+ * and assert the screenshot comparison results are within the mismatch tolerance or are an exact match
+ *
+ * This should be used within an `it` block.
+ *
+ * @param {[name, options]} [args] - the list of test arguments to parse.
```
StephenEsser: nit:  Is this param doc intentional? I don't think I understand what it's defining here

StephenEsser (approved):

emilyrohrbough (changes requested):
Requesting `ignoreComparisonResults` be removed bases on yesterday's meeting. Adding change request to prevent accidental merge before this is updated.

packages/terra-functional-testing/src/commands/expect/toMatchReference.js
emilyrohrbough: >  Several values aren't used in some of these functions.

Which values? Do you mean to call out `isExactSameImage`?

If there are that aren't being used in our screenshot assertion, we should remove them from the comparison results that are generated by the screenshot strategy.

packages/terra-functional-testing/src/config/wdio.conf.js
emilyrohrbough: discussed yesterday that we don't want to implement this override. The repos that need that use-case can provide custom configuration.

packages/terra-functional-testing/src/commands/expect/toMatchReference.js
StephenEsser: Anywhere the `screenshot` object is referenced every property of that object is documented, which sounds good in practice, but is slightly confusing when looking at the purpose of the function. 

`getComparisonResult ` for example documents 4 properties (5 counting the root object), but only uses `misMatchPercentage` and `isSameDimensions`.

I don't think it's inherently bad to keep the documentation

packages/terra-functional-testing/src/services/wdio-visual-regression-service/VisualRegressionLauncher.js
benbcai: Added `name` to docs. https://github.com/cerner/terra-toolkit/pull/545/commits/9933873194e2d7e6d395853f313cdd4cf89d1212

packages/terra-functional-testing/src/config/wdio.conf.js
benbcai: Removed configuration option for baseScreenshotDir. https://github.com/cerner/terra-toolkit/pull/545/commits/9933873194e2d7e6d395853f313cdd4cf89d1212

packages/terra-functional-testing/src/commands/expect/toMatchReference.js
benbcai: Restructured the error message to make it fore human friendly to read based on the suggestion below. https://github.com/cerner/terra-toolkit/pull/545/commits/9933873194e2d7e6d395853f313cdd4cf89d1212

packages/terra-functional-testing/src/commands/expect/toMatchReference.js
benbcai: Restructured the error message to make it more human friendly to read. The `getComparisonResult` method is no longer needed.
https://github.com/cerner/terra-toolkit/pull/545/commits/9933873194e2d7e6d395853f313cdd4cf89d1212

packages/terra-functional-testing/src/commands/expect/toMatchReference.js
benbcai: Updated to only listed out the properties that are used in the method. `isExactSameImage` is not being used so it is removed here and in BaseCompare.js.

https://github.com/cerner/terra-toolkit/pull/545/commits/9933873194e2d7e6d395853f313cdd4cf89d1212

packages/terra-functional-testing/src/commands/validates/screenshot.js
benbcai: This doc is extra and not needed. Removed. https://github.com/cerner/terra-toolkit/commit/9933873194e2d7e6d395853f313cdd4cf89d1212

benbcai:
> Requesting `ignoreComparisonResults` be removed bases on yesterday's meeting. Adding change request to prevent accidental merge before this is updated.

The configuration option for `ignoreComparisonResults` is now removed.
https://github.com/cerner/terra-toolkit/commit/9933873194e2d7e6d395853f313cdd4cf89d1212

emilyrohrbough (approved):

PR #548, opened 2021-01-26
Add $ to globals

jmsv6d:
### Summary

Added $ to list of globals so we don't have to define it in each of our wdio spec files.

### Additional Details

@cerner/terra

### DIFF
--- a/packages/stylelint-config-terra/CHANGELOG.md
+++ b/packages/stylelint-config-terra/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 ## Unreleased
 
-* Changed
-  * Added $ to globals
-
 ## 4.0.0 - (January 5, 2021)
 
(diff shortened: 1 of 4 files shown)


packages/stylelint-config-terra/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Added $ to globals
```
StephenEsser: Since this package is disconnected from the WebDriver infrastructure I think we should include what we're adding the global for since it's just a magic wildcard here

StephenEsser (approved):

yuderekyu (approved):

benbcai (approved):

mjhenkes (approved):

PR #568, opened 2021-02-05
[terra-functional-testing] auto update screenshots

### LINKED ISSUE #530: [terra-functional-testing] Add CLI Option to Auto-Update Screenshots
# Feature Request

## Description

When running my wdio testing, it would be handy to have the wdio runner automatically update the wdio screenshots so I don't have to either:

1. run tests 2x. First to find test failures, second to validate which screenshots were updates
2. delete all reference screenshots and check-in all new screenshots that might have been within the comparison threshold.

My thought is that the screenshots should be updated only if the comparison checks failed because the latest screenshot was outside of mismatch tolerance or the image size differed. 

From a reporting aspect, it could be helpful to report the comparison results even when the reference screenshot is updated.

The VR command/expects command used in the tests would want to check if the comparison results returned `screenshotWasUpdated` to continue/pass.

Here is a branch with the changes I am envisioning: https://github.com/cerner/terra-toolkit/compare/vr-update-screenshots

benbcai:
### Summary
Add the `updateScreenshots` CLI option to automatically update/replace the reference screenshot with latest screenshot if there is a mismatch.

Closes #530 

### Additional Details
- Example command: `npm run terra wdio --updateScreenshots`
- When this option is provided and the reference screenshot matches the latest screenshot, no replacement takes place and the test will pass as usual.
- When this option is provided and the reference screenshot does not matches the latest screenshot, the reference screenshot will be replaced with the latest screenshot and the test step _will_ pass.
- Many of the files are just for updated the case for primitive types on documentations.

@cerner/terra

### DIFF
--- a/packages/terra-functional-testing/src/commands/expect/toMatchReference.js
+++ b/packages/terra-functional-testing/src/commands/expect/toMatchReference.js
@@ -3,11 +3,11 @@
  * the mismatch tolerance and are the same size.
  *
  * @param {Object} screenshot - The comparison result.
- * @param {Boolean} screenshot.isNewScreenshot - If the latest screenshot was a new screenshot that has no reference screenshot.
- * @param {Boolean} screenshot.isSameDimensions - If the latest screenshot was the same size as the reference screenshot.
- * @param {Boolean} screenshot.isWithinMismatchTolerance - If the latest screenshot was within the mismatch tolerance.
- * @param {Number} screenshot.misMatchPercentage - The mismatch percentage when comparing the latest screenshot to the reference screenshot.
- * @param {Boolean} screenshot.screenshotWasUpdated - If the reference screenshot was updated with the latest captured screenshot.
+ * @param {boolean} screenshot.isNewScreenshot - If the latest screenshot was a new screenshot that has no reference screenshot.
+ * @param {boolean} screenshot.isSameDimensions - If the latest screenshot was the same size as the reference screenshot.
+ * @param {boolean} screenshot.isWithinMismatchTolerance - If the latest screenshot was within the mismatch tolerance.
+ * @param {number} screenshot.misMatchPercentage - The mismatch percentage when comparing the latest screenshot to the reference screenshot.
+ * @param {boolean} screenshot.screenshotWasUpdated - If the reference screenshot was updated with the latest captured screenshot.
  * @returns {Object} - An object that indicates if the assertion passed or failed with a message.
  */
 function toMatchReference(screenshot) {
--- a/packages/terra-functional-testing/src/commands/expect/toMatchReference.js
+++ b/packages/terra-functional-testing/src/commands/expect/toMatchReference.js
@@ -3,10 +3,11 @@
  * the mismatch tolerance and are the same size.
  *
  * @param {Object} screenshot - The comparison result.
- * @param {boolean} screenshot.isNewScreenshot - If the latest screenshot was a new screenshot that has no reference screenshot.
- * @param {boolean} screenshot.isSameDimensions - If the latest screenshot was the same size as the reference screenshot.
- * @param {boolean} screenshot.isWithinMismatchTolerance - If the latest screenshot was within the mismatch tolerance.
+ * @param {Boolean} screenshot.isNewScreenshot - If the latest screenshot was a new screenshot that has no reference screenshot.
+ * @param {Boolean} screenshot.isSameDimensions - If the latest screenshot was the same size as the reference screenshot.
+ * @param {Boolean} screenshot.isWithinMismatchTolerance - If the latest screenshot was within the mismatch tolerance.
  * @param {Number} screenshot.misMatchPercentage - The mismatch percentage when comparing the latest screenshot to the reference screenshot.
+ * @param {Boolean} screenshot.screenshotWasUpdated - If the reference screenshot was updated with the latest captured screenshot.
  * @returns {Object} - An object that indicates if the assertion passed or failed with a message.
  */
 function toMatchReference(screenshot) {
@@ -15,11 +16,12 @@ function toMatchReference(screenshot) {
     isSameDimensions,
     isWithinMismatchTolerance,
     misMatchPercentage,
+    screenshotWasUpdated,
   } = screenshot;
 
   // Validate the screenshot is the same size for the case when the new screenshot matches 100% but is 'n' pixel taller due to new content.
   // For example: the latest screenshot of a list has two more items than the reference screenshot so it is 60 pixels taller. That should fail.
-  const imagesMatch = isNewScreenshot || (isSameDimensions && isWithinMismatchTolerance);
+  const imagesMatch = isNewScreenshot || screenshotWasUpdated || (isSameDimensions && isWithinMismatchTolerance);
   let message = '';
 
   if (!isSameDimensions) {
--- a/packages/terra-functional-testing/src/commands/validates/screenshot.js
+++ b/packages/terra-functional-testing/src/commands/validates/screenshot.js
@@ -17,6 +17,7 @@ const screenshot = (testName, options = {}) => {
   const { selector } = options;
   const wrappedOptions = {
     name: testName,
+    updateScreenshots: global.Terra.serviceOptions.updateScreenshots,
     ...options,
   };
 
--- a/packages/terra-functional-testing/src/commands/validates/screenshot.js
+++ b/packages/terra-functional-testing/src/commands/validates/screenshot.js
@@ -16,9 +16,9 @@ const screenshot = (testName, options = {}) => {
 
   const { selector } = options;
   const wrappedOptions = {
-    name: testName,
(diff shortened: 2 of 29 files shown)


mjhenkes:
>Example command: npm run terra -- wdio --updateScreenshots

I would have thought the command would be `npm run terra wdio --updateScreenshots.`

Thoughts on adding a `--u` shortcut (cribbing from jest)
`npm run terra wdio --u`

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
```diff
@@ -97,7 +99,7 @@ class TestRunner {
       gridUrl,
       locales,
       themes,
-      ...launcherOptions // hostname, port, baseUrl, suite, spec, and keepAliveSeleniumDockerService
+      ...launcherOptions // hostname, port, baseUrl, suite, spec, updateScreenshots, and keepAliveSeleniumDockerService
```
StephenEsser: If possible, I think we should avoid flattening our custom configuration options onto the wdio config. A collision is probably unlikely, but I think we should have some distinction between our keys and WebDriverIO's keys similar to the `serviceOptions` key we added. Maybe `launcherOptions` or `customLauncherOptions` or `terraLauncherOptions`. What are your thoughts?

packages/terra-functional-testing/src/commands/expect/toMatchReference.js
```diff
@@ -3,10 +3,11 @@
...
  *
  * @param {Object} screenshot - The comparison result.
- * @param {boolean} screenshot.isNewScreenshot - If the latest screenshot was a new screenshot that has no reference screenshot.
- * @param {boolean} screenshot.isSameDimensions - If the latest screenshot was the same size as the reference screenshot.
- * @param {boolean} screenshot.isWithinMismatchTolerance - If the latest screenshot was within the mismatch tolerance.
+ * @param {Boolean} screenshot.isNewScreenshot - If the latest screenshot was a new screenshot that has no reference screenshot.
+ * @param {Boolean} screenshot.isSameDimensions - If the latest screenshot was the same size as the reference screenshot.
+ * @param {Boolean} screenshot.isWithinMismatchTolerance - If the latest screenshot was within the mismatch tolerance.
  * @param {Number} screenshot.misMatchPercentage - The mismatch percentage when comparing the latest screenshot to the reference screenshot.
+ * @param {Boolean} screenshot.screenshotWasUpdated - If the reference screenshot was updated with the latest captured screenshot.
```
StephenEsser: nit: How do you feel about lowercasing primitives? I've been following the examples on the JSDoc page which lowercase primitive types and capitalize object types:

https://jsdoc.app/tags-param.html

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -57,6 +57,11 @@ const cli = {
         type: 'array',
         describe: 'A list of spec file paths.',
       },
+      updateScreenshots: {
```
StephenEsser: How closely do we want to follow jest's cli options? Our screenshots are nested inside of `__snapshots__`. So this could be a 1-1 command `--updateSnapshot --u`. Although for consumers I think `updateScreenshot(s)` might be more intuitive. I've noticed WebDriverIO and jest both use singular option names, I'm not sure why. 

https://jestjs.io/docs/en/cli#--updatesnapshot

I'm not strongly opinionated towards either option, but I wanted to have a small discussion before we settled on the final commit changes.

benbcai:
> > Example command: npm run terra -- wdio --updateScreenshots
> 
> I would have thought the command would be `npm run terra wdio --updateScreenshots.`
> 
> Thoughts on adding a `--u` shortcut (cribbing from jest)
> `npm run terra wdio --u`

Yes, it should just be `npm run terra wdio --updateScreenshots.` without the `--` before `wdio`. The extra `--` is needed for mono repos, which I miscopied from package.json.

I added the `u` alias (for now). We can talk in the sync to see if anyone thinks we shold not keep it and if we should name it `updateScreenshots`, `updateScreenshot`, or `updateSnapshot`.

https://github.com/cerner/terra-toolkit/pull/568/commits/07340103010232d3e8f52accdb9e9f65c44e0371#diff-b821c3a083e050c54034354280dac343f0b8d9ad66c460cad80a99ec7ebea640R62

packages/terra-functional-testing/src/commands/expect/toMatchReference.js
benbcai: I updated these because other places are using the uppercase but looks like I followed the bad examples. Updated all the primitive types to use lowercase: https://github.com/cerner/terra-toolkit/pull/568/commits/07340103010232d3e8f52accdb9e9f65c44e0371

packages/terra-functional-testing/src/terra-cli/wdio/index.js
benbcai: I added the `u` alias for now. We can discuss in our sync to see what everyone thinks about the alias and naming. https://github.com/cerner/terra-toolkit/pull/568/commits/07340103010232d3e8f52accdb9e9f65c44e0371

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
benbcai: Yeah, I think it does make sense to separate them out. Is this what you're thinking (also need to not blow away the global Terra object [here](https://github.com/cerner/terra-toolkit/blob/440f788697e86c4c4aca1630d134ea1942948a13/packages/terra-functional-testing/src/services/wdio-terra-service.js#L23)) or something more sophisticated?

Screen Shot 2021-02-08 at 4 54 07 PM

StephenEsser: I was thinking of passing the options through onto the config, but scoped into a launcherOptions variable

image

or

image

This would add `launcherOptions` onto the `wdio.conf.js` but in a variable scoped to our custom options being passed through. I free handed this without testing it so it might need tweaked a little.

image

benbcai: Gotcha. I tweaked things around just a bit: https://github.com/cerner/terra-toolkit/pull/568/commits/304bb561772b452a6d5949267e03938b35c0e64c

packages/terra-functional-testing/src/commands/validates/screenshot.js
```diff
@@ -17,6 +17,7 @@ const screenshot = (testName, options = {}) => {
   const { selector } = options;
   const wrappedOptions = {
     name: testName,
+    updateScreenshots: global.Terra.serviceOptions.updateScreenshots,
     ...options,
```
StephenEsser: It might be worth spreading options first; otherwise it creates an unintended API.

```js
// Always update the screenshot.
Terra.validates.screenshot('name', { updateScreenshots: true }); // name too, but that's less concerning
```

benbcai: Good call. https://github.com/cerner/terra-toolkit/pull/568/commits/99e1b636372468bd4b470719b2c9af02185f6ba0

mjhenkes (approved):

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
```diff
@@ -18,20 +18,21 @@ A test runner is provided for local development and should be invoked using the
...
+| hostname                       |       | string  | false    | localhost               | Automation driver host address.                                                               |
+| port                           |       | number  | false    | 4444                    | Automation driver port.                                                                       |
+| baseUrl                        |       | string  | false    |                         | The base url.                                                                                 |
+| suite                          |       | array   | false    |                         | Overrides specs and runs only the defined suites.                                             |
+| spec                           |       | array   | false    |                         | A list of spec file paths.                                                                    |
+| updateScreenshots              | u     | boolean | false    | false                   | Updates all reference screenshots with the latest screenshots.                                |
```
StephenEsser: After our own confusion these might be better to list in a single column with the flags:

| Option       | Type   | Default | Description                                   |
|--------------|--------|---------|-----------------------------------------------|
| --config, -c | string |         | A file path to the test runner configuration. |

I removed the `Required` column in this example, I'm not sure how much value it's providing since no options are required

StephenEsser (approved):

packages/terra-functional-testing/src/terra-dev-site/terra-functional-testing/wdio-testing/About.tool.mdx
benbcai: Added the aliases to the `Options` column and removed the `Alias` and `Required` columns. https://github.com/cerner/terra-toolkit/pull/568/commits/aff1cf1ec2f36c43efc3b6c8fe31743a4092fa01

PR #572, opened 2021-02-11
[terra-functional-testing] Fix locale and theme

jmsv6d:
### Summary

While running tests with the cli, if passed an array of locales the tests were only being run against the first locale provided. This is because the `AssetServerService` defined in the `wdio.conf.js` is instantiated with it's options once at the start of the test run. Subsequent changes to the options do not get passed down. So the locale passed to the WebpackServer was only ever the first locale (this would be the same with themes). Updated to check for env first before the config

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-functional-testing/src/webpack-server/webpack-server.js
+++ b/packages/terra-functional-testing/src/webpack-server/webpack-server.js
@@ -23,9 +23,11 @@ class WebpackServer {
 
     // eslint-disable-next-line global-require, import/no-dynamic-require
     const config = require(webpackConfig);
+    const defaultLocale = process.env.LOCALE || locale;
+    const defaultTheme = process.env.THEME || theme;
 
     if (typeof config === 'function') {
-      return config({ ...locale && { defaultLocale: locale }, theme }, { p: true });
+      return config({ ...defaultLocale && { defaultLocale }, theme: defaultTheme }, { p: true });
     }
 
     return config;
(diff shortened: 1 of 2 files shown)


packages/terra-functional-testing/src/webpack-server/webpack-server.js
```diff
@@ -23,9 +23,11 @@ class WebpackServer {
 
     // eslint-disable-next-line global-require, import/no-dynamic-require
     const config = require(webpackConfig);
+    const defaultLocale = process.env.LOCALE || locale;
+    const defaultTheme = process.env.THEME || theme;
```
mjhenkes, line 27: Could the env be applied before theme and locale are passed to config? Ideally we'd contain checking the env to one location. I think? 🤷‍♂️

StephenEsser:
I think there's more going on here that we should investigate. I've added debug logs to each of the constructors and the values are being passed; the assert service constructor gets instantiated with the correct options, but in the wrong order. WebDriver might be caching the initial run's launcher options somehow.

StephenEsser:
I've logged an issue so we can get feedback from the WebDriverIO maintainers to see if they might have recommendations or suggestions for the behavior we're seeing or our implementation approach. This might be something with node caching the file and not WebDriverIO

https://github.com/webdriverio/webdriverio/issues/6411

benbcai:
The change on #583 should resolve this issue and more generally all config options. I think this can be closed.

mjhenkes:
Closing in favor of #583

PR #582, opened 2021-02-21
[terra-functional-testing] Convert docker service from docker swarm to docker-compose

StephenEsser:
### Summary

This PR converts the selenium docker service to utilize docker-compose instead of docker swarm. This delegates the majority of service and network orchestration to docker-compose. Manually managing the swarm was introducing a lot of complexity and didn't utilize any of the significant swarm features. After testing both the docker swarm and docker-compose I believe most of the instability we were experiencing was directly related to orchestrating the docker swarm. 

These changes have been tested locally and on travis and have shown a significant improvement in performance and stability. To continue testing stability this PR enables firefox  as part of the wdio-site test run and removes the keep alive flag to test that composing up and down operates as expected.

With these changes we could probably remove the `keepAliveSeleniumDockerService` flag. I decided to keep it as an option for consumers to slightly increase (by decreasing) local testing runtime, but it should not be "needed" anymore when running commands back to back. docker-compose starts and stops very fast (after the first-time initial use/installation). 

Changes:

- Replaced docker swarm stack, service, and network orchestration with docker-compose
- Replaced [swarm specific](https://github.com/SeleniumHQ/docker-selenium/tree/selenium-3#version-3-with-swarm-support) docker-compose configuration with [version 3 docker-compose](https://github.com/SeleniumHQ/docker-selenium/tree/selenium-3#version-3)
- Added a [healthcheck](https://github.com/SeleniumHQ/docker-selenium/tree/selenium-3#adding-a-healthcheck-to-the-grid) to the docker-compose using the built-in image check-grid.sh
- Added logic to wait and poll the selenium grid health check before starting tests

### Additional Details

Because we previously initialized the docker swarm on all machines running wdio you may see a warning after switching over. I don't think the warning has any issues, but you can get rid of it by leaving the swarm. `docker swarm leave --force`

I think we should consider removing the following CLI options until we encounter a necessary use-case. The options can be added passively, but cannot be taken away passively.

- port
- hostname
- baseUrl

Verification PRs:

- https://github.com/cerner/terra-toolkit/pull/582 (Surprise)
- https://github.com/cerner/terra-framework/pull/1374
- https://github.com/cerner/terra-core/pull/3359
- https://github.com/cerner/terra-clinical/pull/746

@cerner/terra


packages/terra-functional-testing/src/docker/docker-compose.yml
```diff
@@ -1,10 +1,16 @@
-version: '3.7'
```
StephenEsser, line 1: This file has been converted from the [docker-swarm](https://github.com/SeleniumHQ/docker-selenium/tree/selenium-3#version-3-with-swarm-support) example to the [docker-compose example](https://github.com/SeleniumHQ/docker-selenium/tree/selenium-3#version-3) as recommended by docker-selenium. A [healthcheck](https://github.com/SeleniumHQ/docker-selenium/tree/selenium-3#adding-a-healthcheck-to-the-grid) was also added

mjhenkes (approved):
This all looks much simpler. Good Call.

packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
```diff
@@ -130,85 +70,36 @@ class SeleniumDockerService {
...
+    logger.info('Waiting for the docker selenium hub to become ready...');
 
-    await this.pollCommand(`curl -sSL http://${this.host}:${this.port}/wd/hub/status`, (result) => (
+    await this.pollCommand("docker inspect --format='{{json .State.Health.Status}}' selenium-hub", (result) => (
       new Promise((resolve, reject) => {
         const { stdout } = result;
-        const { value } = JSON.parse(stdout);
 
-        if (value.ready) {
+        if (stdout && stdout.trim() === '"healthy"') {
```
benbcai, line 102: Can just the single quote be used here or does it also need the double quotes?

StephenEsser: It needs the double quotes, the stdout pipes the values in as a quoted string from a JSON value. I decided to compare against that rather than manipulate the string (other than the trim, which is removing empty spaces and new line characters).

benbcai: Follow-up question: Since we're doing the health check on the grid once every 15 seconds, does that mean this may need to wait up to 15 seconds to get the up-to-date status? Otherwise, wouldn't the status be outdated? Or does executing `docker-compose -f` updates the health check status right away. Is that how it works?

StephenEsser: The healthcheck in the dockerfile will perform the check once every 15 seconds. When the compose file launches the status will be "starting" and it will immediately perform a healthcheck. If that healthcheck doesn't succeed it keeps trying for the number of `retries` every 15 seconds until it appears as healthy. If it grid was already up (say using the keep alive flag or the test run was hard killed) the grid will be healthy immediately because it is already running. The docker service is currently polling that health check every two seconds for up to two minutes. I don't think we have to worry about outdated statuses here, the tests won't launch unless a successful healthy check succeeds

Edit:

We can tweak the healthcheck retries, intervals, and polling times. As they are now they seem to be significantly more stable than the swarm network orchestration. We'll need to account for a grid launching the first time and a grid that is already active (kept alive)

benbcai: Thanks for the explanation!

benbcai (approved):

PR #583, opened 2021-02-23
[terra-functional-testing] Remove ENVs to use CLI options

### LINKED ISSUE #573: [terra-functional-testing] Update runner to not set environment variables
# Story

## Description

As a terra developer, I want to ensure that we do not set environment variables based on command line parameters

benbcai:
### Summary
Remove usage of environment variables in favor or CLI options.

Closes #573 

### Additional Details
1. Added new CLI options to replace existing [ENVs](https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/config/wdio.conf.js#L15-L25).
2. The idea here is to allow all appropriate CLI options to be passed down to the test runner to override those in the wdio config. Given the issue presented on #572 and https://github.com/webdriverio/webdriverio/issues/6411 where the options passed to the service are cached and unreliable should they get changed dynamically, those service options are merged with the CLI options passed to the test runner with the CLI options have precedence.

@cerner/terra


packages/terra-functional-testing/src/services/wdio-visual-regression-service/VisualRegressionLauncher.js
```diff
@@ -11,8 +11,24 @@ class VisualRegressionLauncher {
...
+     * Always use the launcher options from the test runner if available before using
+     * the options passed thru the service in wdio.conf.js. The reason is because the
+     * service options set using env is cached and are unreliable if these options
+     * are changed dynamically by the test runner.
+     *
+     * Reference: https://github.com/webdriverio/webdriverio/issues/6411
+     */
+    const visualRegressionOptions = {
+      ...(locale || options.locale) && { locale: locale || options.locale },
+      ...(theme || options.theme) && { theme: theme || options.theme },
```
emilyrohrbough: Should this also be passing the `updateScreenshots` config value?

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
```diff
@@ -29,39 +31,69 @@ class TestRunner {
...
-      if (formFactor) {
-        process.env.FORM_FACTOR = formFactor;
-      }
+      const launcherOptions = {
+        ...cliOptions,
+        baseUrl: `http://${externalHost || getIpAddress()}:${externalPort || 8080}`,
+        capabilities: getCapabilities(browsers, !!gridUrl),
+        hostname: gridUrl || hostname || 'localhost',
+        port: gridUrl ? 80 : 4444,
+        launcherOptions: {
```
StephenEsser: Thinking about this a little more, I'm wondering if it would be better to define the services here. Services here would overwrite the default services in the wdio config and we could remove the custom property merges in the services constructors. We'd still probably need the launcher options for the onComplete hook to pass down test run configuration the reporter's merge function. But we wouldn't need to do anything custom in our services, they would receive the test run options normally. 

Caveats (with either solution) is that Jenkins will need to continue to run the `wdio` command. It would not be able to use the test runner cli directly while it is still using envs. We might be able to make Jenkins changes will wdio v7 to remove all envs. 

Edit:

Something like this:

```js
  services: [
    [TerraService, {
      ...formFactor && { formFactor },
      ...theme && { theme },
    }],
    [AssetServerService, {
      ...site && { site },
      ...locale && { locale },
      ...theme && { theme },
      ...port && { port },
      // Webpack path 
    }],
    [VisualRegressionLauncher, {
      ...locale && { locale },
      ...theme && { theme },
    }],
    // Do not add the docker service if disabled.
    ...(disableSeleniumService || gridUrl ? [] : [[SeleniumDockerService]]),
  ],
  ```

File: packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js [hunk 3]
```diff
@@ -22,16 +22,19 @@ class SeleniumDockerService {
...
+    this.disableSeleniumService = disableSeleniumService === true;
     this.keepAliveSeleniumDockerService = keepAliveSeleniumDockerService === true;
 
-    try {
-      // Verify docker is installed.
-      await exec('docker -v');
-      await this.startSeleniumHub();
-    } catch (error) {
-      throw new SevereServiceError(error);
+    if (!this.disableSeleniumService) {
```
benbcai, line 32: This check here and in `onComplete` below is to basically mimic what [this](https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/config/wdio.conf.js#L146) is doing to ignore running this service if `WDIO_DISABLE_SELENIUM_SERVICE` or `SELENIUM_GRID_URL` is set.

I thought about just keeping [this](https://github.com/cerner/terra-toolkit/pull/583/files#diff-f13ec26e86b6e8d765ec9020014d38600b7df8643e37a26d40000d28807e2bfbL110-L112) to continue to use ENV only for this scenario but it is inconsistent with everything else. So I went with this logic here. Any better suggestions?

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
```diff
@@ -29,39 +31,69 @@ class TestRunner {
...
-
-      if (formFactor) {
-        process.env.FORM_FACTOR = formFactor;
-      }
+      const launcherOptions = {
+        ...cliOptions,
+        baseUrl: `http://${externalHost || getIpAddress()}:${externalPort || 8080}`,
+        capabilities: getCapabilities(browsers, !!gridUrl),
+        hostname: gridUrl || hostname || 'localhost',
+        port: gridUrl ? 80 : 4444,
```
benbcai: I want to discuss in our next sync on whether we set the following CLI options as is provided thru the runner or keep these logics in place. If we use these as they are provided, we are giving the responsibilities to the consumers to set these correctly. This goes back to @StephenEsser's comment [here](https://github.com/cerner/terra-toolkit/pull/582#issue-577197761) on whether to keep them.

- baseUrl
- port
- hostname

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -5,6 +5,15 @@ const cli = {
   describe: 'Run wdio tests',
   builder: (yargs) => (
     yargs.options({
+      assetServerPort: {
```
benbcai, line 8: `assetServerPort` is replacing [WDIO_INTERNAL_PORT](https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/config/wdio.conf.js#L24) which is used as the _port_ in AssetServerService. We currently already have a `port` CLI options so naming this `INTERNAL_PORT` seems confusing with the existing `port` option. So the name is changed to `assetServerPort`.

packages/terra-functional-testing/tests/wdio/viewport-helpers-spec.js
```diff
@@ -1,24 +0,0 @@
-Terra.describeViewports('Viewport Helpers', ['small', 'large'], () => {
-  it('creates tiny and huge screens', () => {
-    browser.url('/compare.html');
-    // Terra.validates.screenshot();
```
benbcai, line 4: This was initially created to test capturing screenshots and viewports. We have since added the terra-validates-spec.js wdio test and getViewports.test.js jest tests so this is no longer appropriate or needed.

packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
StephenEsser: If we chose to implement the services in the terra-cli I think this issue would go away:

https://github.com/cerner/terra-toolkit/pull/583#discussion_r581276065

Edit:

With the recent changes to the cli and launcher options this option is still needed.

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
StephenEsser: My vote is to remove these options from the cli. As implemented today I don't think all the necessary configurations are in sync. Specifically regarding the port mapping in the docker-compose file. We could add these options back in the future if we encounter a scenario we need to support.

package.json
```diff
@@ -77,7 +77,7 @@
     "start-heroku": "npm run terra -- express-server --port $PORT --site './build' ",
     "terra": "node ./packages/terra-cli/bin/terra",
     "test": "npm run lint && npm run jest && npm run mocha && npm run wdio",
-    "wdio-site": "SITE=build npm run terra -- wdio --browsers chrome firefox",
+    "wdio-site": "npm run terra -- wdio --site build --browsers chrome firefox --locales en fr --formFactors small large",
```
mjhenkes, line 80: Was the locale and form factor scoping for testing?

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
mjhenkes: I also vote to remove

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
benbcai: I like this idea! I assume with this we will continue to keep how the services are defined as they are now in wdio.conf.

package.json
benbcai: This PR should fix the issue on #572 where only the first locale/theme/formFactor/etc is used if multiple are passed down. This is for making sure that the subsequent values in the options are used as expected. Do you think adding this is unnecessary and creates too many screenshots?

packages/terra-enzyme-intl/tests/jest/__snapshots__/enzyme-intl.test.jsx.snap
```diff
@@ -34,6 +34,7 @@ exports[`mountWithIntl using injectIntl should match the snapshot 1`] = `
     foo="bar"
     intl={
       Object {
+        "__addMessages": [Function],
```
benbcai: This is unrelated and I don't know why this snapshot is updated. I'll investigate.

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
StephenEsser: That sounds right, I think with these changes we won't need to make many modifications to the current wdio.conf. The terra-cli will define any necessary configurations that will take precedence over the wdio.conf, but we can leave the current wdio.conf as a default for Jenkins to use and configure with ENVs. Then we can passively switch Jenkins over to the terra-cli and circle back on a major version bump and delete all ENVs from the default wdio.conf.

mjhenkes: +1 to consolidating the merges

packages/terra-functional-testing/src/services/wdio-visual-regression-service/VisualRegressionLauncher.js
benbcai: Currently, it is being passed down to the [global.browser.checkElement](https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/commands/validates/screenshot.js#L21) command. Yes, it would be more consistent for Visual Regression to get `updateScreenshots` directly from the CLI options.

https://github.com/cerner/terra-toolkit/pull/583/commits/2251ca53df2fa449f3ced6b1ef0d2d8f1b25e385

I will make a second update to pass it to the VR service options.

packages/terra-enzyme-intl/tests/jest/__snapshots__/enzyme-intl.test.jsx.snap
benbcai: My node modules were outdated. clean:install reverted this: https://github.com/cerner/terra-toolkit/pull/583/commits/2251ca53df2fa449f3ced6b1ef0d2d8f1b25e385

package.json
mjhenkes: nah, it's fine

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
benbcai: Added the services to the launcher: https://github.com/cerner/terra-toolkit/pull/583/commits/eb6e22e962d6bb12f197d735445237e6d2aaa5ff

packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
benbcai: Addedthe services to the launcher and reverted these changes since they are no longer needed: https://github.com/cerner/terra-toolkit/pull/583/commits/eb6e22e962d6bb12f197d735445237e6d2aaa5ff

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
benbcai: Removed baseUrl, hostname, and port CLI options: https://github.com/cerner/terra-toolkit/pull/583/commits/70319fe7996b688f517f3cbaf0d2eb7d3947111b

packages/terra-functional-testing/src/commands/validates/screenshot.js
```diff
@@ -18,7 +18,6 @@ const screenshot = (testName, options = {}) => {
   const wrappedOptions = {
     ...options,
     name: testName,
-    updateScreenshots: global.Terra.serviceOptions.updateScreenshots === true,
```
benbcai, line 21: Instead of passing down `updateScreenshots` to the global.browser.checkElement, it is now being read from the [Visual Regression service options](https://github.com/cerner/terra-toolkit/pull/583/files#diff-dea71618479daf3d0e21b24c6f7201de0dc18481e558caf0c3aee745a434af06R17).

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -14,6 +19,19 @@ const cli = {
         alias: 'config',
         describe: 'A file path to the test runner configuration.',
       },
+      disableSeleniumService: {
+        type: 'boolean',
+        describe: 'A flag to disable the selenium docker service.',
+        default: false,
+      },
```
StephenEsser, line 49: I think this is probably fine to keep, but I don't know of any use-cases anymore after updating travis to run against the docker service. The service is automatically disabled if a gridUrl is provided. If you're not running against the grid, you need to run against the selenium docker service. 


Edit:

After the recent changes we need to keep this option to toggle the service on and off.

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -32,30 +50,22 @@ const cli = {
...
+      spec: {
+        type: 'array',
+        describe: 'A list of spec file paths.',
       },
       suite: {
         type: 'array',
         describe: 'Overrides specs and runs only the defined suites.',
       },
-      spec: {
+      themes: {
```
StephenEsser, line 129: I'm starting to understand why the industry convention seems to be singular. I think it's fine to keep it plural, just an observation. I'm starting to see this more often:

```
    "wdio-lowlight": "terra wdio --themes clinical-lowlight-theme",
    "wdio-fusion": "terra wdio --themes orion-fusion-theme",
```

And Jenkins will end up looking like this after we eventually update it to use the cli because it runs each permutation as a different process:

`terra wdio --browsers chrome --formFactors huge --locales en`

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
```diff
@@ -29,39 +35,86 @@ class TestRunner {
...
+        ],
+        launcherOptions: {
+          disableSeleniumService: disableSeleniumService || !!gridUrl,
+          ...assetServerPort && { assetServerPort },
+          ...formFactor && { formFactor },
+          ...keepAliveSeleniumDockerService && { keepAliveSeleniumDockerService },
+          ...locale && { locale },
+          ...site && { site },
+          ...theme && { theme },
+          ...updateScreenshots && { updateScreenshots },
```
StephenEsser: It doesn't hurt anything to expose all of these, but I think limiting it to only what we need might reduce some code here. I think now only the onComplete reads this and it only needs theme, locale, form factor .

Edit:

With the recent launcher option changes we need to expose all of these

StephenEsser (approved):
Looks good to me. I think we could create a `getServices` helper utility similar to getCapabilities that could be shared between the cli and default config, but I think more thought needs to go into it. It's a nice to have, not a need to have.

StephenEsser:
Ah, I think WebDriver is merging the services array instead of overwriting it.

benbcai:
> Ah, I think WebDriver is merging the services array instead of overwriting it.

Yeah, what I'm seeing is that each of the four services get initialized twice. When running `npm run terra -- wdio --site build`, the Launcher would run the ExpressServer because of site while the wdio config would run WebpackServer simultaneously and both are fighting for port 8080.

Looks like WebDriver is using [deepmerge](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-config/src/lib/ConfigParser.ts#L137) and by default [arrays are merged by concatenating](https://github.com/TehShrike/deepmerge#api) them. There is no way for us to pass down the `overwriteMerge` option. I guess adding the services this way is not working as we expect.

benbcai:
The config shows these services are merged:

>   services: [
>     [ [class TerraService], {} ],
>     [ [class AssetServerService], [Object] ],
>     [ [class VisualRegressionLauncher], {} ],
>     [ [class SeleniumDockerService] ],
>     [ [class TerraService], [Object] ],
>     [ [class AssetServerService], [Object] ],
>     [ [class VisualRegressionLauncher], [Object] ],
>     [ [class SeleniumDockerService] ]
>   ],

It also looks like WebDriver only add hooks to services defined in wdio.config but not services passed thru the Launcher. For testing purposes I removed all four services in wdio.config so that it would only use the services passed in the Launcher. What I noticed is that none of the hooks in these services get called. 

If I understand the wdio-cli code correctly it looks like they first [add the config file to the Parser and then merge the overrides](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-cli/src/launcher.ts#L76-L77). But they [add the hooks only to the services](https://github.com/webdriverio/webdriverio/blob/main/packages/wdio-config/src/lib/ConfigParser.ts#L109-L116) defined in the config file. This is done before the overrides are merged so any override services won't have any hooks added. I don't know if this is a bug or a limitation in supporting services passed thru the Launcher. I might log an issue to see what they say. For now we probably would have to go back to what we did before to merge the options in each service constructor.

StephenEsser:
I logged this issue for the services but haven't heard back yet:

https://github.com/webdriverio/webdriverio/issues/6503

 It's a bummer, defining the services in the launcher seemed like a great way to ensure they were initialized correctly with the permutation options. 

> For now we probably would have to go back to what we did before to merge the options in each service constructor.

I think you're right. For testing I looked into doing something like this:

https://github.com/cerner/terra-toolkit/compare/env-to-cli-2...launcher-options

It's the same launcherOptions idea but I removed all the constructor options so we don't have to merge them. At this point I think we're stuck with using the launcherOptions one way or another. Merging the options like we were previously is probably fine too, I was just trying to get the ENVs somewhere we could delete them later.

Edit:

The caveat to the branch above is that both Jenkins and local CI develop must use the terra-cli. That's one thing we should keep in mind when deciding which option we want to go with. With the original launcher option Jenkins needs to run `wdio` directly. If the ENVS are managed inside of the launcher somehow then Jenkins needs to use the terra cli. If the ENVs are managed inside of the wdio.conf Jenkins will need to use `wdio`

```
{
  "test:wdio": "wdio" // vs terra wdio
}
```

StephenEsser (approved):

mjhenkes (approved):

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -5,22 +5,91 @@ const cli = {
...
+
+            return [browsers];
+          }
+
+          return [];
+        },
       },
       c: {
         type: 'string',
         alias: 'config',
```
yuderekyu, line 42: should the alias and object key be switched? it looks like what we have defined results in `--c, -config`

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -30,32 +99,43 @@ const cli = {
...
         type: 'string',
-        describe: 'Automation driver host address.',
-      },
-      port: {
-        type: 'number',
-        describe: 'Automation driver port.',
+        describe: 'A file path to a static directory of assets. When defined, an express server will launch to serve the assets and disable running webpack.',
+        default: () => {
+          if (process.env.SITE) {
+            return [process.env.SITE];
```
yuderekyu: does the site value need to be inside an array?

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -5,22 +5,91 @@ const cli = {
...
+      },
       browsers: {
         type: 'array',
         describe: 'A list of browsers for the test run.',
+        default: () => {
+          let browsers = process.env.BROWSERS;
+          if (browsers) {
+            // Remove the brackets if BROWSERS is set as an array like ['chrome, firefox, ie'] or ['chrome','firefox','ie'] in Jenkinsfile.
+            browsers = browsers.replace('[', '');
+            browsers = browsers.replace(']', '');
```
yuderekyu, line 27: not a big deal, but we could reduce this to one replace call with a regex. if there's a chance there's more than one set of brackets, we could add the `/g` flag
```suggestion
            browsers = browsers.replace('/\[|\]/', '');
```

yuderekyu (approved):
really cool approach!

packages/terra-functional-testing/src/terra-cli/wdio/index.js
benbcai: That is what I thought at first. But basically yargs sets the key names as equivalent such that updates to a key will propagate to aliases and vice-versa.

Optionally .alias() can take an object that maps keys to aliases. Each key of this object should be the canonical version of the option, and each value should be a string or an array of strings.

This is their documentation: https://yargs.js.org/docs/#api-reference-optionskey-opt

packages/terra-functional-testing/src/terra-cli/wdio/index.js
benbcai: Yes, it should be a string. Good catch! https://github.com/cerner/terra-toolkit/pull/583/commits/b0d401b05c22bf675b0eaf05d09b897f3c9e76df

packages/terra-functional-testing/src/terra-cli/wdio/index.js
benbcai: Since we're parsing just the brackets, I thought the code is easier to read/understand this way. The regex is cool but the  escape and pipe characters makes it harder to read.

PR #589, opened 2021-02-26
[terra-functional-testing] Pass theme as `defaultTheme` to webpack-config-terra

benbcai:
### Summary
- webpack-config-terra reads the theme either from `process.env.THEME` or preferably [env.defaultTheme](https://github.com/cerner/terra-toolkit/blob/main/packages/webpack-config-terra/src/webpack.config.js#L50). terra-functional-testing recently removed setting all options in process.env. As a result, testing in other themes are failure and it uncovers that webpack-server should pass down theme as `defaultTheme` instead of `theme` which is not what webpack-config-terra expects. 

-  Updated doc from `getCssProperty` to `getCSSProperty`. It used to be in lowercase but is now upper case: https://v6.webdriver.io/docs/api/element/getCSSProperty.html

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/About.1.tool.mdx
+++ b/packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/About.1.tool.mdx
@@ -196,7 +196,7 @@ it('should hide the caret', () => {
   Terra.hideInputCaret('#inputID');
   const element = browser.element('#inputID');
 
-  expect(element.getCSSProperty('caretColor').value).to.equal('rgba(0,0,0,0)');
+  expect(element.getCSSProperty('caretColor').value).toEqual('rgba(0,0,0,0)');
 });
 ```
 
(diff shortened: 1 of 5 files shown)


packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/About.1.tool.mdx
```diff
@@ -196,7 +196,7 @@ it('should hide the caret', () => {
   Terra.hideInputCaret('#inputID');
   const element = browser.element('#inputID');
 
-  expect(element.getCssProperty('caretColor').value).to.equal('rgba(0,0,0,0)');
+  expect(element.getCSSProperty('caretColor').value).to.equal('rgba(0,0,0,0)');
```
StephenEsser: We'll also want to update the chai `to.equal` to expect's `toEqual` 

```suggestion
  expect(element.getCSSProperty('caretColor').value).toEqual('rgba(0,0,0,0)');
```

StephenEsser (approved):

packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/About.1.tool.mdx
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/589/commits/dc1880f9c4fc5a432c10859b56dd5d749818cfda

mjhenkes (approved):

Issue #595, opened 2021-03-09
[terra-functional-testing] Update documentation


StephenEsser:
# Feature Request

## Description

This issue should be used to do a second pass on the terra-functional-testing documentation. This includes adding, removing, and modifying the existing documentation. A few examples of improvements changes:

- Add documentation regarding the terra-cli dependency requirement in the About page
- Add documentation for configuring the test run service options. We currently have documentation in the individual services but it is a little confusing from a consumers perspective. The services are abstracted away. We should provide examples for configuring the `serviceOptions` in the `wdio.conf.js` file in the `About` section or a new `Configuration` section.
- Modify the Service documentation to be from the consumers perspective. The services should rarely, if ever, be used individually. The documentation should instead explain what the service is doing and how it can be modified in the consumers project. (serviceOptions in the wdio.conf.js).
- Modify the Service documentation to remove examples that pass options through the wdio.conf.js. The services no longer accept options through the constructor; they use the launcherOptions from the cli interface. 
- Miscellaneous cleanup/improvements

PR #598, opened 2021-03-09
[terra-functional-testing] Fix selenium docker version service option

### LINKED ISSUE #596: [terra-functional-testing] Fix selenium docker version service option
# Bug Report

## Description

The selenium docker version option should be deconstructed from the `serviceOptions` in the configuration object. It is currently being deconstructed from the `launcherOptions`. 

https://github.com/cerner/terra-toolkit/blob/28f4e2d9c0339b01db3a70990f66e187e56835c1/packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js#L21

```js
// Deconstruct from the serviceOptions
 const { launcherOptions, serviceOptions } = config;
 const { disableSeleniumService, keepAliveSeleniumDockerService } = launcherOptions || {};
 const { version } = serviceOptions || {};
```

This can be tested by updating the selenium docker version in the serviceOptions override and verifying docker pulls down the expected image.

```js
// wdio.conf.js
const { config } = require('@cerner/terra-functional-testing/lib/config/wdio.conf');

config.serviceOptions = {
  version: 'new version',
};

exports.config = config;
```

StephenEsser:
### Summary

This PR fixes the API for specifying the docker version option. `version` should be deconstructed from `serviceOptions` instead of `launcherOptions`.

Fixes #596.

### Additional Details

Since this wasn't working as expected on the 1.0.0 release we could rename this option if we wanted to. 

@cerner/terra

### DIFF
--- a/packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
+++ b/packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
@@ -17,8 +17,9 @@ class SeleniumDockerService {
    * @param {Object} config - The object containing the wdio configuration and options defined in the terra-cli test runner.
    */
   constructor(_options, _capabilities, config = {}) {
-    const { launcherOptions } = config;
-    const { disableSeleniumService, keepAliveSeleniumDockerService, version } = launcherOptions || {};
+    const { launcherOptions, serviceOptions } = config;
+    const { disableSeleniumService, keepAliveSeleniumDockerService } = launcherOptions || {};
+    const { version } = serviceOptions || {};
 
     this.disableSeleniumService = disableSeleniumService === true;
     this.keepAliveSeleniumDockerService = keepAliveSeleniumDockerService === true;
@@ -92,14 +93,14 @@ class SeleniumDockerService {
 
     await exec(`${envVars}docker-compose -f ${this.getDockerComposeFilePath()} up -d`);
     await this.waitForSeleniumHubReady();
+
+    logger.info('Successfully started the docker selenium hub.');
   }
 
   /**
    * Waits for the docker selenium hub to become healthy.
    */
   async waitForSeleniumHubReady() {
-    logger.info('Waiting for the docker selenium hub to become ready...');
-
     await this.pollCommand("docker inspect --format='{{json .State.Health.Status}}' selenium-hub", (result) => (
       new Promise((resolve, reject) => {
         const { stdout } = result;
--- a/packages/terra-functional-testing/CHANGELOG.md
+++ b/packages/terra-functional-testing/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fixed
+  * Fix docker version service option to be read from serviceOptions instead of launcherOptions
+
 ## 1.0.1 - (March 1, 2021)
 
 * Fixed
--- a/packages/terra-functional-testing/tests/jest/services/wdio-selenium-docker-service.test.js
+++ b/packages/terra-functional-testing/tests/jest/services/wdio-selenium-docker-service.test.js
@@ -12,6 +12,8 @@ const config = {
   launcherOptions: {
     disableSeleniumService: true,
     keepAliveSeleniumDockerService: true,
+  },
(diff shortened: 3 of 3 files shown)


packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
```diff
@@ -17,8 +17,9 @@ class SeleniumDockerService {
    * @param {Object} config - The object containing the wdio configuration and options defined in the terra-cli test runner.
    */
   constructor(_options, _capabilities, config = {}) {
-    const { launcherOptions } = config;
-    const { disableSeleniumService, keepAliveSeleniumDockerService, version } = launcherOptions || {};
+    const { launcherOptions, serviceOptions } = config;
+    const { disableSeleniumService, keepAliveSeleniumDockerService } = launcherOptions || {};
+    const { version } = serviceOptions || {};
```
StephenEsser, line 22: Since this wasn't working as expected on the 1.0.0 release we could rename this option if we wanted to.

packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
```diff
@@ -92,14 +93,14 @@ class SeleniumDockerService {
 
     await exec(`${envVars}docker-compose -f ${this.getDockerComposeFilePath()} up -d`);
     await this.waitForSeleniumHubReady();
+
+    logger.info('Successfully started the docker selenium hub.');
```
StephenEsser, line 97: I added this log so we can see when the docker hub is active. Sometimes in large projects webpack takes a while to complete and because the last log was "starting docker..." it seems like it's waiting on docker, but it's actually waiting on webpack. This just adds verification the docker hub started.

packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
benbcai: We currently [read the selector](https://github.com/cerner/terra-toolkit/blob/28f4e2d9c0339b01db3a70990f66e187e56835c1/packages/terra-functional-testing/src/services/wdio-terra-service.js#L27) override from [serviceOptions](https://engineering.cerner.com/terra-ui/dev_tools/cerner/terra-functional-testing/wdio-services/terra-service#selector) so renaming it would be non-passive. Do consumers ever need to set the version? If so, should we add some documentation to show this is where to specify the version?

benbcai (approved):

packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
StephenEsser: Are you referring to renaming `serviceOptions` or `version`. My comment was regarding the name of `version`.

> Do consumers ever need to set the version?

`version` was added so that consumers could periodically update and test against newer selenium docker images. I don't think anyone is doing this today because we rarely update the internal grid, but as we start to make updates more frequently I think this option will become more adopted for testing. 

version is documented [here](https://engineering.cerner.com/terra-ui/dev_tools/cerner/terra-functional-testing/wdio-services/selenium-docker-service#version), but it's out of date now that we no longer accept constructor options. I'll add a note to the documentation update issue #595

benbcai: Sounds good. I thought you were referring to renaming `serviceOptions` so ignore my comment.

mjhenkes (approved):

packages/terra-functional-testing/src/services/wdio-selenium-docker-service.js
benbcai: Since `version` is no longer provided directly as a service option and is now defined at the flat config level, maybe it makes sense to give it a more descriptive name so that we know what this version refers to. (e.g. `seleniumDockerVersion`, `seleniumDockerImageVersion`, etc).

StephenEsser: Whoops, I think I merged at the same time your comment came through. There's still time to update it before the release if we want to change it

PR #605, opened 2021-03-17
[terra-functional-testing] Invalid selector error message

### LINKED ISSUE #475: [terra-functional-testing] VR Service Meaningful Error Message for Incorrect Element Selector
# Feature Request

## Description
Resolve this https://github.com/zinserjan/wdio-visual-regression-service/issues/74

benbcai:
### Summary
Log an error with a more meaningful message when an incorrect selector is used to capture screenshot.

### Additional Details
- Closes #475 
- The original issue shows this error when an incorrect selector is used. Perhaps this error message was logged when either on Node 8 or an older version of wdio-novus-visual-regression-service.
Screen Shot 2021-03-17 at 1 27 15 PM

This is the error when on Node 10:
Screen Shot 2021-03-17 at 1 54 53 AM

This is the error when on Node 12:
Screen Shot 2021-03-17 at 1 48 09 AM

The above error messages from jimp are slightly different because it depends on when the exception occurs based on the node version.

This is the new error message with this change:
Screen Shot 2021-03-17 at 1 49 40 AM

@cerner/terra

### DIFF
--- a/packages/terra-functional-testing/src/services/wdio-visual-regression-service/modules/makeElementScreenshot.js
+++ b/packages/terra-functional-testing/src/services/wdio-visual-regression-service/modules/makeElementScreenshot.js
@@ -28,6 +28,11 @@ async function makeElementScreenshot(browser, elementSelector, options = {}) {
 
   // get bounding rect of elements
   const boundingRects = await browser.execute(getBoundingRects, elementSelector);
+
+  if (boundingRects.length === 0) {
+    throw new Error(`[wdio-visual-regression-service:makeDocumentScreenshot] Failed to capture the element using the "${elementSelector}" selector. Either update the test document to include this selector or use a different selector that exists on the document.`);
+  }
+
   const boundingRect = groupBoundingRect(boundingRects);
 
   // make screenshot of area
--- a/packages/terra-functional-testing/tests/wdio/terra-validates-spec.js
+++ b/packages/terra-functional-testing/tests/wdio/terra-validates-spec.js
@@ -38,27 +38,15 @@ Terra.describeViewports('Terra.validates', ['small', 'large'], () => {
     });
 
     it('should require a screenshot name', () => {
-      let caughtError;
+      const errorMessage = '[terra-functional-testing:screenshot] Terra.validate.screenshot requires a unique test name as the first argument.';
 
-      try {
-        Terra.validates.screenshot();
-      } catch (error) {
-        caughtError = error;
-      }
-
-      expect(caughtError.message).toEqual('[terra-functional-testing:screenshot] Terra.validate.screenshot requires a unique test name as the first argument.');
+      expect(() => Terra.validates.screenshot()).toThrow(errorMessage);
     });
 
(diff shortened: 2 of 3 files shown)


packages/terra-functional-testing/tests/wdio/terra-validates-spec.js
```diff
@@ -48,6 +48,18 @@ Terra.describeViewports('Terra.validates', ['small', 'large'], () => {
...
+    it('should fail with invalid selector', () => {
+      let caughtError;
+
+      try {
+        Terra.validates.screenshot('invalid selector', { selector: 'invalid-selector' });
+      } catch (error) {
+        caughtError = error;
+      }
+
+      expect(caughtError.message).toEqual('[wdio-visual-regression-service:makeDocumentScreenshot] Failed to capture the element using the "invalid-selector" selector. Either update the test document to include this selector or use a different selector that exists on the document.');
```
yuderekyu: jest has a toThrow method to capture error messages - could that be used here instead of a try/catch? try/catch is a bit clunky

emilyrohrbough (approved):

mjhenkes (approved):

packages/terra-functional-testing/tests/wdio/terra-validates-spec.js
benbcai: .toThrow is certainly cleaner and easier: https://github.com/cerner/terra-toolkit/pull/605/commits/ad8e15eb67ba4838ff642dde18e0590e207c2386

yuderekyu (approved):

packages/terra-functional-testing/src/services/wdio-visual-regression-service/modules/makeElementScreenshot.js
```diff
@@ -28,6 +28,11 @@ async function makeElementScreenshot(browser, elementSelector, options = {}) {
 
   // get bounding rect of elements
   const boundingRects = await browser.execute(getBoundingRects, elementSelector);
+
+  if (boundingRects.length === 0) {
+    throw new Error(`[wdio-visual-regression-service:makeDocumentScreenshot] Failed to capture the element using the "${elementSelector}" selector. Either update the test document to include this selector or use a different selector that exists on the document.`);
```
StephenEsser, line 33: Do we want to use the [terra-cli logger](https://engineering.cerner.com/terra-ui/dev_tools/cerner/terra-cli/logger) here? I'm not sure what our rule of thumb is between throwing errors and using the logger to output the error. Throwing an error might make more sense here to break out of the function

StephenEsser (approved):

packages/terra-functional-testing/src/services/wdio-visual-regression-service/modules/makeElementScreenshot.js
benbcai: We want to fail the test in this situation. Using terra-cli logger to log an error does not fail the test and we would still need to return a value (string) which would need to be handled downstream.

PR #607, opened 2021-03-22
[terra-functional-testing] Visual Regression: endX/endY out of range error.

### LINKED ISSUE #476: [terra-functional-testing] VR Service - fix x-y out of bounds error
# Feature Request

## Description

For history on issue see: https://github.com/zinserjan/wdio-screenshot/issues/87

When pulled into the wdio-novus-visual-regression-service, that we pull forward into the terra-functional-testing, the maintainer has commented this out with a comment reference this issue. [Code Reference](https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/services/wdio-visual-regression-service/utils/strategies/BaseStrategy.js#L15)

We need to evaluate if this "fix" is sufficient or if something else should be done to handle the x-y out of bround error. I feel like the follow should work, but needs some testing to validate:

```js
  setScrollArea(coordinates) {
    let {
      startX, startY, endX, endY,
    } = coordinates;

    const documentWidth = this.screenDimensions.getDocumentWidth();
    const documentHeight = this.screenDimensions.getDocumentHeight();

    if (startX >= documentWidth) {
      startX = documentWidth;
    } else if (startY >= documentHeight) {
      startY = documentHeight;
    } else if (endX > documentWidth) {
      endX = documentWidth;
    } else if (endY > documentHeight) {
      endY = documentWidth;
    }

    this.area = {
      startX,
      startY,
      endX,
      endY,
    };
  }

```

benbcai:
### Summary
Fix endY/endX out of range error when the size of the selector element is larger or extends outside the document size.

### Additional Details
- Closes #476 
- This error typically occurs when the selector element being captured for screenshot extends beyond the size of the current document or test page. One way this can occur is when using the `transform: translate3d` style as seen in [terra-slide-panel](https://github.com/cerner/terra-framework/blob/9e8e5a454ab22cd4feec1d25cf5f3adbe0543e10/packages/terra-slide-panel/src/SlidePanel.module.scss#L60).

@cerner/terra


packages/terra-functional-testing/src/services/wdio-visual-regression-service/utils/strategies/BaseStrategy.js
```diff
@@ -13,25 +13,18 @@ class BaseStrategy {
...
-    // } else if (startY >= documentHeight) {
-    //   throw new Error('startY is out of range');
-    // } else if (endX > documentWidth) {
-    //   throw new Error('endX is out of range');
-    // } else if (endY > documentHeight) {
-    //   throw new Error('endY is out of range');
-    // }
+    const documentWidth = this.screenDimensions.getDocumentWidth();
+    const documentHeight = this.screenDimensions.getDocumentHeight();
+    const adjustedStartX = startX > documentWidth ? documentWidth : startX;
```
emilyrohrbough: The adjustedStart X & Y seem a bit weird. If I'm following this correctly, this is saying the start x is off the screen? Thinking of an example being a closed slide panel, would the resulting image be a 0px by 0px if both start and end 7 values are the doc width?

This could be more of an edge case? if startx === endx || starty === endy?

benbcai: Yes, in the case where startX and StartY are > documentWidth and documentHeight respectively, this means the selector element is 100% off the screen and it would be a 0px image. Would it make more sense to default adjustedStartX/adjustedStartY to 0,0 so that it would capture the full screen?

```
    const adjustedStartX = startX > documentWidth ? 0 : startX;
    const adjustedStartY = startY > documentHeight ? 0 : startY;
```

mjhenkes (approved):

tbiethman (approved):
This has been a pain in the butt for a long time.

packages/terra-functional-testing/src/services/wdio-visual-regression-service/utils/strategies/BaseStrategy.js
```diff
@@ -13,25 +13,21 @@ class BaseStrategy {
...
-    //   throw new Error('endX is out of range');
-    // } else if (endY > documentHeight) {
-    //   throw new Error('endY is out of range');
-    // }
+    // Capture what is visible if the element is rendered partially off the screen.
+    // Capture the entire viewable screen dimension if it is rendered completely off the screen.
+    const adjustedStartX = startX > documentWidth || startX < 0 ? 0 : startX;
+    const adjustedStartY = startY > documentHeight || startY < 0 ? 0 : startY;
+    const adjustedEndX = endX > documentWidth || endX < 0 ? documentWidth : endX;
+    const adjustedEndY = endY > documentHeight || endY < 0 ? documentHeight : endY;
```
StephenEsser, line 28: Is there any benefit to logging a message in the scenario the screen is clipped? 

We can leave this as is and circle back if this causes any confusion. I'm wondering if this might cause screenshots to occasionally change without any logging to indicate it was clipped because it was offscreen.

StephenEsser (approved):

PR #609, opened 2021-03-24
[terra-functional-testing] Remove out-of-bound log message.

benbcai:
### Summary

- Removed log message for out of range elements in screenshot because there are valid cases to have out of range elements.
- Added documentation to describe how visual regression determines screenshot dimensions.

### Additional Details
There are valid scenarios to have elements rendered outside the viewport and we do not want the log message to display for every test with these valid scenarios.

Examples include:
- Any component using the VisuallyHiddenText component that is rendered with a negative margins
- Components using negative values for the `left` style
- Components using negative values in `translate3d` such as `transform: translate3d(-100%, 0, 0)`.

@cerner/terra

### DIFF
--- a/packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/wdio-services.4/VisualRegressionService.tool.mdx
+++ b/packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/wdio-services.4/VisualRegressionService.tool.mdx
@@ -83,10 +83,10 @@ All of these provide options that will help you to capture screenshots in differ
 available:
 
 * **hide** `Object[]`<br />
-  hides all elements queried by all kinds of different [WebdriverIO selector strategies](http://webdriver.io/guide/usage/selectors.html) (via `visibility: hidden`)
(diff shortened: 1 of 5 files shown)


packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/wdio-services.4/VisualRegressionService.tool.mdx
```diff
@@ -83,10 +83,10 @@ All of these provide options that will help you to capture screenshots in differ
 available:
```
benbcai, line 83: Should we remove this VisualRegressionService mdx because everything described here is for internal use?Consumers shouldn't use this directly and this would create more confusion for them.

mjhenkes: @StephenEsser is working on a doc revamp. I believe he address this question in his PR.

mjhenkes (approved):

StephenEsser (approved):

PR #619, opened 2021-03-31
Create rule for minimum versions for the package.json linter

### LINKED ISSUE #556: Create a rule for requiring minimum versions of packages
# Story

## Description

As a terra developer I want a package-json-lint rule that requires minimum versions of packages.

## Motivation

So that I can provide guidance to consumers of terra on which versions of packages should ideally be used

## Definition of Done

We'll know we're done when:

* There is a rule class created that:
  * Has a constructor that allows the specification of packages and version numbers but defaults to the set of packages that are utilizing the new theme strategy (see the end of this issue for specifics)
  * A lint method that takes package lock json data to evaluate that fails if the rules packages and versions are not met and succeeds if they are met 
* The code is merged to main and tests succeed

## Specifics of Default Packages and Versions

```js
{
  '@cerner/terra-docs': '>=1.0.0',
  'terra-abstract-modal': '>=3.25.0',
  'terra-action-footer': '>=2.42.0',
  'terra-action-header': '>=2.43.0',
  'terra-alert': '>=4.29.0',
  'terra-application-header-layout': '>=3.28.0',
  'terra-application-links': '>=6.34.0',
  'terra-application-name': '>=3.30.0',
  'terra-application-navigation': '>=1.37.0',
  'terra-application-utility': '>=2.35.0',
  'terra-application': '>=1.19.0',
  'terra-avatar': '>=3.3.0',
  'terra-badge': '>=3.35.0',
  'terra-brand-footer': '>=2.24.0',
  'terra-button-group': '>=3.39.0',
  'terra-button': '>=3.36.0',
  'terra-card': '>=3.27.0',
  'terra-cell-grid': '>=1.5.0',
  'terra-clinical-data-grid': '>=2.25.0',
  'terra-clinical-detail-view': '>=3.20.0',
  'terra-clinical-header': '>=3.16.0',
  'terra-clinical-item-display': '>=3.18.0',
  'terra-clinical-item-view': '>=3.19.0',
  'terra-clinical-label-value-view': '>=3.20.0',
  'terra-clinical-onset-picker': '>=4.21.0',
  'terra-collapsible-menu-view': '>=6.34.0',
  'terra-date-input': '>=1.14.0',
  'terra-date-picker': '>=4.38.0',
  'terra-date-time-picker': '>=4.38.0',
  'terra-demographics-banner': '>=3.37.0',
  'terra-dialog-modal': '>=3.38.0',
  'terra-dialog': '>=2.42.0',
  'terra-divider': '>=3.27.0',
  'terra-dropdown-button': '>=1.14.0',
  'terra-form-checkbox': '>=4.3.0',
  'terra-form-field': '>=4.3.0',
  'terra-form-fieldset': '>=2.42.0',
  'terra-form-input': '>=3.5.0',
  'terra-form-radio': '>=4.5.0',
  'terra-form-select': '>=6.6.0',
  'terra-form-textarea': '>=4.5.0',
  'terra-grid': '>=6.21.0',
  'terra-html-table': '>=1.6.0',
  'terra-hyperlink': '>=2.34.0',
  'terra-icon': '>=3.32.0',
  'terra-image': '>=3.28.0',
  'terra-layout': '>=4.24.0',
  'terra-list': '>=4.31.0',
  'terra-menu': '>=6.34.0',
  'terra-modal-manager': '>=6.34.0',
  'terra-navigation-side-menu': '>=2.31.0',
  'terra-notification-dialog': '>=3.35.0',
  'terra-overlay': '>=3.49.0',
  'terra-paginator': '>=2.51.0',
  'terra-popup': '>=6.35.0',
  'terra-profile-image': '>=3.30.0',
  'terra-progress-bar': '>=4.23.0',
  'terra-search-field': '>=3.51.0',
  'terra-section-header': '>=2.37.0',
  'terra-show-hide': '>=2.35.0',
  'terra-signature': '>=2.30.0',
  'terra-slide-group': '>=4.21.0',
  'terra-slide-panel': '>=3.27.0',
  'terra-spacer': '>=3.40.0',
  'terra-status-view': '>=4.27.0',
  'terra-switch': '>=1.0.0',
  'terra-table': '>=4.8.0',
  'terra-tabs': '>=6.35.0',
  'terra-tag': '>=2.35.0',
  'terra-text': '>=4.31.0',
  'terra-time-input': '>=4.29.0',
  'terra-toolbar': '>=1.8.0'
}
```

ryanthemanuel:
### Summary

Create rule for minimum versions for the package.json linter. This will handle getting off of versions that have a peer dependency on terra-base and getting off of versions that don't support the theme context.

Currently, this is working with a hard coded config, however MVP for the overall project should allow for teams to override rules as needed.

Fixes #556


packages/package-json-lint/src/lint.js
```diff
@@ -1,7 +1,43 @@
...
+const rules = require('./rules');
+const PackageIssues = require('./issues/PackageIssues');
+const AggregateIssues = require('./issues/AggregateIssues');
+const { getRuleConfig, getConfig } = require('./config');
+const { getPathsForPackages } = require('./utilities');
 
-const logger = new Logger({ prefix: '[terra-functional-testing:wdio]' });
+const lint = ({ packageJsonData, config }) => {
+  const issues = [];
+  Object.entries(config.rules).forEach(([ruleId, ruleInformation]) => {
```
yuderekyu: could we just use filter here, instead of forEach + array.push?

packages/package-json-lint/src/rules/require-no-terra-base-peer-dependency-versions.js
```diff
@@ -0,0 +1,99 @@
...
+  { name: 'terra-toggle-section-header', versionRange: '>2.5.0' },
+  { name: 'terra-toggle', versionRange: '>3.5.0' },
+  { name: 'terra-visually-hidden-text', versionRange: '>2.4.0' },
+];
+
+module.exports = {
+  lint: ({ packageJsonData, ruleConfig }) => {
+    const { dependencies = {} } = packageJsonData;
+    const currentProblems = [];
+    versionSet.forEach(({ name, versionRange }) => {
```
yuderekyu: same filter question here

packages/package-json-lint/src/config/index.js
```diff
@@ -0,0 +1,15 @@
+const getConfig = async () => ({
```
mjhenkes, line 1: I assume the async here is future thinking?

packages/package-json-lint/src/issues/LintIssue.js
```diff
@@ -0,0 +1,28 @@
...
+    this.severity = severity;
+    this.node = node;
+    this.lintMessage = lintMessage;
+  }
+
+  /**
+   * Helper to convert the LintIssue to a printable string
+   * @returns {string} Human-friendly message about the lint issue
+   */
+  toString() {
```
mjhenkes: I think this class could just be a helper function that returns a string, but i don't know if there is a good reason to pick one over the other 🤷‍♂️

packages/package-json-lint/src/lint.js
```diff
@@ -1,7 +1,43 @@
...
+    const rule = rules[ruleId];
+    const ruleConfig = getRuleConfig({ rule, ruleInformation });
+    if (ruleConfig.severity !== 'off') {
+      const issue = rule.lint({ packageJsonData, ruleConfig });
```
mjhenkes: So how you have this designed now you'll analyze the package data once per rule. Depending on how many rules we add it may be more efficient to read the package data once and apply the rules where appropriate. I think this is how other linters are applied. For the two rules we have now your current approach is probably fine, but something to think about for the future.

mjhenkes: For example rules could subscribe to sections of the package.json they care about and as we read through the document we could send the info to each rule.

So a "dependency" rule would get called with (dependency, version, type(dep, peer, dev)) and each rule could evaluate and return its results.

packages/package-json-lint/src/utilities/index.js
```diff
@@ -0,0 +1,24 @@
...
+
+const glob = util.promisify(callbackGlob);
+
+const getIgnorer = async () => {
+  const ignoreFileContents = await fs.readFile(IGNORE_FILE_PATH, 'utf8');
+  return ignore().add(ignoreFileContents);
+};
+
+const getPathsForPackages = async () => {
+  const paths = await glob(path.join(process.cwd(), '**', 'package.json'), { ignore: [path.join('**', 'node_modules', '**')] });
```
mjhenkes: should we always assume a package at root and look for a lerna config to glob for mono repos? looking to avoid unnecessary globs.

packages/package-json-lint/src/config/index.js
ryanthemanuel: Yep. Eventually I'll need to do some file analysis.

packages/package-json-lint/src/lint.js
ryanthemanuel: Yeah I thought about that a little bit. Could make it easier to make a vs code plugin I think. Hmmm.

packages/package-json-lint/src/utilities/index.js
ryanthemanuel: I was wondering if we would want to use this in a server as well.

packages/package-json-lint/src/lint.js
ryanthemanuel: Hmmm. Would it need to be a combo of map and then filter?  Since we're not actually filtering Object.entries(config.rules) but instead transforming it (and then filtering it)

yuderekyu: ohh I see now, good point - we get slightly better readability with map + filter but it's two iterations - doing this in one iteration isn't that hard to read 🤷🏼 could go either way

packages/package-json-lint/src/issues/LintIssue.js
ryanthemanuel: This will be resolved by my big refactoring.

packages/package-json-lint/src/lint.js
ryanthemanuel: This will be fixed by my big refactoring

packages/package-json-lint/src/utilities/index.js
ryanthemanuel: Updated this to cover all scenarios

packages/package-json-lint/package.json
```diff
@@ -30,11 +30,21 @@
...
+    "fs-extra": "^9.0.1",
+    "glob": "^7.1.6",
+    "ignore": "^5.1.8",
+    "log-symbols": "^4.0.0",
+    "semver": "^7.3.4"
+  },
   "peerDependencies": {
-    "@cerner/terra-cli": "^1.0.0"
+    "@cerner/terra-cli": "^1.0.0",
+    "@npmcli/promise-spawn": "^1.2.0"
```
yuderekyu: should this be a dependency instead?

packages/package-json-lint/src/rules/require-theme-context-versions.js
```diff
@@ -0,0 +1,97 @@
...
+  { name: 'terra-switch', versionRange: '>=1.0.0' },
+  { name: 'terra-table', versionRange: '>=4.8.0' },
+  { name: 'terra-tabs', versionRange: '>=6.35.0' },
+  { name: 'terra-tag', versionRange: '>=2.35.0' },
+  { name: 'terra-text', versionRange: '>=4.31.0' },
+  { name: 'terra-time-input', versionRange: '>=4.29.0' },
+  { name: 'terra-toolbar', versionRange: '>=1.8.0' },
+];
+
+module.exports = {
```
yuderekyu, line 78: this seems to be the same logic as the other rule, with the exception of the lint message and id. How about we turn this into a generic minimum version function, then pipe in the message context and id?

```js
// separate function
const minVersionRule = ({ dependencies, lintContext, lintId }} => {
// inject lintContext and lintId for generated strings
}

// usage
module.exports = {
  create: ({ ruleConfig, report }) => ({
    dependencies: (dependencies) => minVersionRule({ dependencies, lintContext: 'theming context', lintId: 'require-theme-context-versions' )};
  });
};
```

packages/package-json-lint/package.json
ryanthemanuel: Good call. Updated

mjhenkes (approved):

packages/package-json-lint/src/rules/utils/require-version-set.js
```diff
@@ -0,0 +1,20 @@
+const semver = require('semver');
```
yuderekyu: nitpick - i actually think this fits fine under the `rules` directory, rather than a `utils` sub-directory. We could just not export it in the index file. What do you think?

yuderekyu (approved):

benbcai (approved):

PR #623, opened 2021-04-08
[terra-functional-testing] Honor terra-theme.config.js file

### LINKED ISSUE #622: [terra-functional-testing] Honor terra-theme.config.js
# Feature Request

## Description

Today when the `terra wdio` command is run, by default screenshots are saved to the `terra-default-theme` folder and `terra-default-theme` is passed into the webpack config as the default theme.

Instead of defaulting the 'terra-default-theme' when no theme is specified we should read from the `terra-theme.config.js`file. If a default is specified there we should use it if not we should use `terra-default-theme`. If the `terra-theme.config.js` is read, we should not pass the default theme to webpack and allow webpack to read the theme config itself. Passing in the default theme removes the other themes and disallows dynamic theme switching.

mjhenkes:
### Summary

This pr updated terra-functional-testing to honor the terra-theme.config.js file's `theme` key for screenshot naming and to not pass in a default theme when the theme config is used.

Closes #622

### Additional Details

@cerner/terra


packages/terra-functional-testing/src/config/utils/getConfigurationOptions.js
```diff
@@ -36,7 +37,8 @@ const getConfigurationOptions = (options) => {
       locale,
       port: assetServerPort,
       site,
-      theme,
+      ...(theme ? { theme } : { theme: getDefaultThemeName() }),
+      overrideTheme: theme,
```
mjhenkes, line 41: override theme is only used in the webpack service. I need to know when a theme is specified in the test runner but for all the other services we always need the theme name.

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -134,7 +134,7 @@ const cli = {
             return [process.env.THEME];
           }
 
-          return ['terra-default-theme'];
+          return [undefined];
```
mjhenkes, line 137: by leaving this an array with undefined we can still use the looping logic in the test runner, but loop once with an undefined theme.

packages/terra-functional-testing/src/config/utils/getDefaultThemeName.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = () => {
+  let themeConfig = {};
+
+  const defaultConfig = path.resolve(process.cwd(), CONFIG);
+  if (fs.existsSync(defaultConfig)) {
+    // eslint-disable-next-line global-require, import/no-dynamic-require
+    themeConfig = require(defaultConfig);
+  }
+  return themeConfig.theme || 'terra-default-theme';
```
mjhenkes, line 14: if theme isn't defined we use  `terra-default-theme` for the theme name

benbcai: I think we can also remove [this default theme](https://github.com/cerner/terra-toolkit/blob/cb967914cd9f14eeb54d7bf9c1847cc6592d366c/packages/terra-functional-testing/src/services/wdio-terra-service.js#L26) in wdio-terra-service. I don't think it was needed without this change since we're were defaulting it in [test runner](https://github.com/cerner/terra-toolkit/pull/623/files#diff-b821c3a083e050c54034354280dac343f0b8d9ad66c460cad80a99ec7ebea640) before.

benbcai (approved):

packages/terra-functional-testing/src/config/utils/getDefaultThemeName.js
mjhenkes: updated in: https://github.com/cerner/terra-toolkit/pull/623/commits/5d59c66b4cea90a2d10813c8c0f8ced6fb30cf64

tbiethman (approved):

ryanthemanuel (approved):

PR #628, opened 2021-04-16
[terra-functional-testing] Clear Screenshots

### LINKED ISSUE #620: [terra-functional-testing] Clean up the latest, diff, and error directories on new test runs
# Story

## Description

As a consumer of terra, I would like for my `latest`, `diff`, and `error` directories to be deleted at the start of test runs

## Motivation

So that I will know that the screenshots I am looking at in those three directories are from the latest test run rather than a previous test run

## Definition of Done

We'll know we're done when the directories are cleared out at the start of the test runs and

* Tests are written and passing
* The code is merged to main

benbcai:
### Summary
Deletes the `diff`, `error`, and `latest` screenshot directories, if they exist, before each wdio test run.

Closes #620 

### Additional Details

- The [cleanScreenshots.js](https://github.com/cerner/terra-toolkit/blob/clear-snapshots/packages/terra-functional-testing/src/commands/utils/cleanScreenshots.js) is mostly copied from [clean-screenshots.js](https://github.com/cerner/terra-toolkit-boneyard/blob/main/scripts/wdio/clean-screenshots.js) in terra-toolkit-boneyard with a few tweaks added to look for the screenshots directories directly in the `tests/wdio` file path for both monorepo and non monorepo projects instead of searching in all directories including the node-modules folder.

@cerner/terra

### DIFF
--- a/packages/terra-functional-testing/src/commands/utils/cleanScreenshots.js
+++ b/packages/terra-functional-testing/src/commands/utils/cleanScreenshots.js
@@ -13,6 +13,7 @@ async function cleanScreenshots() {
   const isMonoRepo = fs.existsSync(monoRepoPath);
   let patterns = [];
 
+  // Check whether or not it a monorepo and then get the paths to the snapshot directories.
   if (isMonoRepo) {
     const packageNames = fs.readdirSync(monoRepoPath);
 
@@ -29,11 +30,12 @@ async function cleanScreenshots() {
     ];
   }
 
-  let screenshotDirectories = [];
-  patterns.forEach((pattern) => {
-    screenshotDirectories = screenshotDirectories.concat(glob.sync(pattern));
-  });
+  // Determine the existing snapshot directories.
+  const screenshotDirectories = patterns.filter((pattern) => (
+    glob.sync(pattern).length > 0
+  ));
 
+  // Delete the existing snapshot directories.
   screenshotDirectories.forEach((dir) => {
     if (isDirectory(dir)) {
       fs.removeSync(dir);
--- a/packages/terra-functional-testing/src/commands/utils/cleanScreenshots.js
+++ b/packages/terra-functional-testing/src/commands/utils/cleanScreenshots.js
@@ -1,0 +1,46 @@
+const fs = require('fs-extra');
+const glob = require('glob');
+const path = require('path');
+const { Logger } = require('@cerner/terra-cli');
+
+const logger = new Logger({ prefix: '[terra-functional-testing:clean-snapshots]' });
+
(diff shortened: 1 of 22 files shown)


packages/terra-functional-testing/src/commands/utils/cleanScreenshots.js
```diff
@@ -0,0 +1,46 @@
...
+    });
+  } else {
+    patterns = [
+      `${process.cwd()}/tests/wdio/__snapshots__/diff`,
+      `${process.cwd()}/tests/wdio/__snapshots__/error`,
+      `${process.cwd()}/tests/wdio/__snapshots__/latest`,
+    ];
+  }
+
+  let screenshotDirectories = [];
```
yuderekyu: what are your thoughts being more declarative and using `map` instead of `forEach` + `concat`
```suggestion
  const screenshotDirectories = patterns.map(pattern => glob.sync(pattern));
```

packages/terra-functional-testing/src/commands/utils/cleanScreenshots.js
```diff
@@ -0,0 +1,46 @@
...
+  screenshotDirectories.forEach((dir) => {
+    if (isDirectory(dir)) {
+      fs.removeSync(dir);
+    }
+  });
+
+  logger.info('Cleaned wdio snapshots directories.');
+}
+
+module.exports = cleanScreenshots;
```
yuderekyu, line 48: we should add this file to the `utils/index.js` so we have a single entry point for import/export and update our imports in other files to reflect that change

packages/terra-functional-testing/src/commands/utils/cleanScreenshots.js
benbcai: I ended up using `filter` since it needs to filter any directories that do not exist: https://github.com/cerner/terra-toolkit/pull/628/commits/29ec637ee165c09c0e227a686256a230f46e080b#diff-024b00fe17f65e235595743f7f75f67609cd5f7fafff46c24cf90251c08d011eR34

packages/terra-functional-testing/src/commands/utils/cleanScreenshots.js
benbcai: Updated: https://github.com/cerner/terra-toolkit/pull/628/commits/29ec637ee165c09c0e227a686256a230f46e080b

yuderekyu (approved):

ShettyAkarsh (approved):

jmsv6d (approved):

mjhenkes (approved):

PR #646, opened 2021-04-30
[terra-functional-testing] Add useSeleniumStandalonService option

### LINKED ISSUE #637: [terra-functional-testing] Handle Jenkins setting to disable selenium docker service
# Story

## Description
As a maintainer of terra-functional-testing, I want to read the env variable set by Jenkins that indicated the selenium grid is not used for testing

## Motivation
So that I know whether or not there is a need to spin up the local selenium docker server to run the tests.

## Definition of Done
We'll know we're done when terra-functional-testing can handle this configuration similar to [terra-toolkit-boneyard](https://github.com/cerner/terra-toolkit-boneyard/blob/main/config/wdio/selenium.config.js#L83-L97).

## Additional Details
When the `useSeleniumGrid` option in Jenkinsfile is set to false to indicate not to use the selenium grid for testing, we need a new env variable that terra-functional-testing can read to determine to disable the selenium docker service.

benbcai:
### Summary
Added `useSeleniumStandalonService` option to determine whether to use the standalone-chrome or the the selenium docker service when not using the selenium grid.

Closes #637 

### Additional Details

- When the Selenium Grid is not used (`gridUrl` is not provided) to run wdio tests, Jenkins will now set the `USE_SELENIUM_STANDALONE_SERVICE` env variable to allow us to determine whether to use standalone-chrome as the hostname or start the selenium docker service. This is similar to how the `CI` env is used in [terra-toolkit-boneyard](https://github.com/cerner/terra-toolkit-boneyard/blob/main/config/wdio/selenium.config.js#L74-L97).
- I didn't include `useSeleniumStandalonService` in the [CLI options documentation](https://engineering.cerner.com/terra-ui/dev_tools/cerner/terra-functional-testing/about#options) because consumers should not need to know about or set this locally since it should only be used by Jenkins.

@cerner/terra

### DIFF
--- a/packages/terra-functional-testing/CHANGELOG.md
+++ b/packages/terra-functional-testing/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added useSeleniumStandalonService option for using the standalone-chrome host instead of the selenium docker service when building in Jenkins.
+
 ## 1.2.0 - (April 23, 2021)
 
 * Added
--- a/packages/terra-functional-testing/CHANGELOG.md
+++ b/packages/terra-functional-testing/CHANGELOG.md
@@ -6,7 +6,7 @@
   * Update specPath in BaseCompare to replace `node_modules` with `tests/wdio`.
 
 * Added
-  * Added useSeleniumStandalonService option for using the standalone-chrome host instead of the selenium docker service when building in Jenkins.
+  * Added useSeleniumStandaloneService option for using the standalone-chrome host instead of the selenium docker service when building in Jenkins.
 
 ## 1.2.0 - (April 23, 2021)
 
--- a/packages/terra-functional-testing/src/config/utils/getConfigurationOptions.js
+++ b/packages/terra-functional-testing/src/config/utils/getConfigurationOptions.js
@@ -22,15 +22,16 @@ const getConfigurationOptions = (options) => {
     suite,
     theme,
     updateScreenshots,
+    useSeleniumStandalonService,
   } = options;
 
   return {
     baseUrl: `http://${externalHost || getIpAddress()}:${externalPort || 8080}`,
     capabilities: getCapabilities(browsers, !!gridUrl),
-    hostname: gridUrl || 'localhost',
+    hostname: gridUrl || (useSeleniumStandalonService ? 'standalone-chrome' : 'localhost'),
     port: gridUrl ? 80 : 4444,
     launcherOptions: {
-      disableSeleniumService: disableSeleniumService || !!gridUrl,
+      disableSeleniumService: disableSeleniumService || useSeleniumStandalonService || !!gridUrl,
       formFactor,
       gridUrl,
       keepAliveSeleniumDockerService,
--- a/packages/terra-functional-testing/src/config/utils/getConfigurationOptions.js
+++ b/packages/terra-functional-testing/src/config/utils/getConfigurationOptions.js
@@ -22,16 +22,16 @@ const getConfigurationOptions = (options) => {
     suite,
     theme,
     updateScreenshots,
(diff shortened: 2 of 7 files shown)


packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -143,6 +143,11 @@ const cli = {
         describe: 'Whether or not to automatically update all reference screenshots with the latest screenshots.',
         default: false,
       },
+      useSeleniumStandalonService: {
```
benbcai: I'm open to suggestions if there is a better name to use.

ShettyAkarsh (approved):

packages/terra-functional-testing/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 * Fixed
   * Update specPath in BaseCompare to replace `node_modules` with `tests/wdio`.
 
+* Added
+  * Added useSeleniumStandalonService option for using the standalone-chrome host instead of the selenium docker service when building in Jenkins.
```
mjhenkes:
```suggestion
  * Added useSeleniumStandaloneService option for using the standalone-chrome host instead of the selenium docker service when building in Jenkins.
```

packages/terra-functional-testing/src/config/utils/getConfigurationOptions.js
```diff
@@ -22,15 +22,16 @@ const getConfigurationOptions = (options) => {
     suite,
     theme,
     updateScreenshots,
+    useSeleniumStandalonService,
```
mjhenkes:
```suggestion
    useSeleniumStandaloneService,
```

packages/terra-functional-testing/src/terra-cli/wdio/index.js
mjhenkes: seems good enough to me
```suggestion
      useSeleniumStandaloneService: {
```

mjhenkes (approved):
+1 with updated spelling

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -143,6 +143,11 @@ const cli = {
...
+        type: 'boolean',
+        describe: 'A flag to use the selenium standalone service instead of the selenium docker service.',
+        default: () => process.env.USE_SELENIUM_STANDALONE_SERVICE === 'true',
```
yuderekyu: can we just check the existence of the env?

```suggestion
        default: process.env.USE_SELENIUM_STANDALONE_SERVICE
```

yuderekyu (approved):

packages/terra-functional-testing/CHANGELOG.md
benbcai: Good catch. Updated all.
https://github.com/cerner/terra-toolkit/pull/646/commits/ee6355ab0f9f8b345db66ce3900222f0f3d87731

packages/terra-functional-testing/src/terra-cli/wdio/index.js
benbcai: envs are always a string and the type here needs to be a boolean. Otherwise, it would default to a string of 'true'.

yuderekyu: boooo

yuderekyu: do we still need to execute this as a function though?
```suggestion
        default: process.env.USE_SELENIUM_STANDALONE_SERVICE === 'true',
```

benbcai: You win!

PR #655, opened 2021-05-17
Add CLI option to disable screenshot mismatch

### LINKED ISSUE #632: [terra-functional-testing] Add a new CLI option that toggles runtime snapshot mismatches on and off
# Story

## Description

As a consumer of terra, I'd like to have an option that will allow for functional testing to be run without failing on snapshot mismatches.

## Motivation

So that I can run tests in a CI setting and have the build go through even with snapshot mismatches. This will allow us to report back on a given PR that there are failures and allow for a team to approve/deny those mismatches.

## Definition of Done

We'll know we're done when the command line option exists that allows mismatches to occur and the build to succeed and this code is merged to main.

BenBoersma:
### Summary

Adding an CLI flag to disable screenshot mismatch failure.
Closes #632 

### Additional Details

@cerner/terra


packages/terra-functional-testing/src/commands/expect/toMatchReference.js
```diff
@@ -36,7 +37,7 @@ function toMatchReference(screenshot) {
   }
 
   return {
-    pass: imagesMatch === true,
+    pass: disableSnapshotMismatch ? true : imagesMatch === true,
```
yuderekyu: my vote is to use the logical OR operator to make this less verbose
```suggestion
    pass: disableSnapshotMismatch || imagesMatch,
```

packages/terra-functional-testing/src/commands/validates/screenshot.js
```diff
@@ -19,8 +19,8 @@ const screenshot = (testName, options = {}) => {
     ...options,
     name: testName,
   };
-
   const screenshotResult = global.browser.checkElement(selector || global.Terra.serviceOptions.selector, wrappedOptions);
```
yuderekyu: personal preference, but we could be more declarative by returning a new object instead of appending a new property

```suggestion
  const checkElementResult = global.browser.checkElement(selector || global.Terra.serviceOptions.selector, wrappedOptions);
  const disableSnapshotMismatch = global.Terra.serviceOptions.disableSnapshotMismatch;
  const result = {
    ...checkElementResult,
    disableSnapshotMismatch,
  };
  
  global.expect(result).toMatchReference();
```
```

File: packages/terra-functional-testing/src/terra-cli/wdio/index.js [hunk 3]
```diff
@@ -47,6 +47,11 @@ const cli = {
         describe: 'A flag to disable the selenium docker service.',
         default: false,
       },
+      disableSnapshotMismatch: {
```
benbcai: We use the word "screenshot" instead of "snapshot" in `updateScreenshots`. I think we should be consistent and also use "screenshot" here. I also think we should use the word "ignore" instead of "disable" since we're not disabling the comparisons...we are only ignoring.

Name suggestions:
`ignoreScreenshotMistmatch`
`ignoreScreenshotComparison`

mjhenkes: +1 @benbcai Either of your suggested names would work for me.

packages/terra-functional-testing/src/commands/validates/screenshot.js
benbcai: We should just be able to check for `global.Terra.serviceOptions.disableSnapshotMismatch` directly in [toMatchReference](https://github.com/cerner/terra-toolkit/pull/655/files#diff-1372ed3653dff72eae9a8aa8b3969b4e50473f645766c7c2f232298e514d0518R40) instead of adding it to the result and then pull it out later.

packages/terra-functional-testing/src/commands/expect/toMatchReference.js
BenBoersma: Changed here https://github.com/cerner/terra-toolkit/pull/655/commits/8571fc4e7f0de6b5aa3178eb5c70084c3fa80539#diff-1372ed3653dff72eae9a8aa8b3969b4e50473f645766c7c2f232298e514d0518R39

packages/terra-functional-testing/src/commands/validates/screenshot.js
BenBoersma: Changed here https://github.com/cerner/terra-toolkit/pull/655/commits/8571fc4e7f0de6b5aa3178eb5c70084c3fa80539#diff-1372ed3653dff72eae9a8aa8b3969b4e50473f645766c7c2f232298e514d0518R39

packages/terra-functional-testing/src/terra-cli/wdio/index.js
BenBoersma: Renamed to `ignoreScreenshotMistmatch`, also updated description to say screenshot
https://github.com/cerner/terra-toolkit/pull/655/commits/8571fc4e7f0de6b5aa3178eb5c70084c3fa80539#diff-b821c3a083e050c54034354280dac343f0b8d9ad66c460cad80a99ec7ebea640R50

yuderekyu: mistmatch isn't a word - let's update this to `ignoreScreenshotMismatch`

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -47,6 +47,11 @@ const cli = {
         describe: 'A flag to disable the selenium docker service.',
         default: false,
       },
+      ignoreScreenshotMistmatch: {
+        type: 'boolean',
+        describe: 'A flag to disable failing on screenshot mismatch.',
```
benbcai:
```suggestion
        describe: 'A flag to ignore screenshot mismatch.',
```

packages/terra-functional-testing/tests/wdio/terra-validates-spec.js
```diff
@@ -37,6 +37,15 @@ Terra.describeViewports('Terra.validates', ['small', 'large'], () => {
       Terra.validates.screenshot('validates-screenshot');
     });
 
+    it('should pass if ignoreScreenshotMistmatch is true even when mismatched', () => {
+      const oldIgnoreScreenshotMistmatch = Terra.serviceOptions.ignoreScreenshotMistmatch;
+      Terra.serviceOptions.ignoreScreenshotMistmatch = true;
+      browser.url('/insufficient-color-contrast.html');
+
+      Terra.validates.screenshot('validates-mismatch');
```
benbcai: I think this test is misleading and not testing what we think. There is no mismatch when comparing this screenshot and test will pass even when ignoreScreenshotMismatch was set to false because it should always match. I wonder if we really need this test since it is already being covered in jest and this is not generating a mismatch screenshot.

packages/terra-functional-testing/src/terra-cli/wdio/index.js
benbcai: Sorry, it was my typo.

packages/terra-functional-testing/tests/wdio/terra-validates-spec.js
BenBoersma: I copied a different test image and saved it as the name for this test specifically to make sure there was a mismatch (couldn't find a way to test this without doing it that way). 

I do think I agree that we might not need it with jest, however we may want it for the next step of this series of changes, where whatever reporter we end up using/making reports that there are mismatches.

benbcai: Ah I see. However, that different test image used here will get deleted when ever the `updateScreenshots` option is used to regenerate all screenshots.

Instead of using a different image for comparison, another option could be to use the same screenshot name as a previous test so no new screenshots are needed. But this too also seems a bit hacky. My suggestion is to leave out such test for now and add one as needed if/when the reporter is used for reporting these mismatches.

packages/terra-functional-testing/src/terra-cli/wdio/index.js
yuderekyu: eh, we can also consider this a lesson of not blindly applying a change :)

packages/terra-functional-testing/tests/jest/commands/expect/toMatchReference.test.js
```diff
@@ -1,6 +1,14 @@
 const toMatchReference = require('../../../../src/commands/expect/toMatchReference');
 
 describe('toMatchReference', () => {
+  beforeEach(() => {
```
yuderekyu: In my opinion, we should make each unit test self contained by removing this setup from the beforeEach, and inserting it into every test  - there will be duplication, but each test will be extremely clear on the setup and assertions. Otherwise, we need to remember that `ignoreScreenshotMismatch` is false by default, which is defined all the way at the top of the file. What do you think?

packages/terra-functional-testing/src/terra-cli/wdio/index.js
BenBoersma: yep, renamed here https://github.com/cerner/terra-toolkit/pull/655/commits/9b9e06d9786ebf0471e3f3e0f7c2df80639a004c#diff-5121be21da689ee0a1bd5a529bbd578dc2bde92ce93b497f118e58290be63dc1R6

packages/terra-functional-testing/tests/wdio/terra-validates-spec.js
BenBoersma: Removed teh wdio test and screenshots here https://github.com/cerner/terra-toolkit/pull/655/commits/9b9e06d9786ebf0471e3f3e0f7c2df80639a004c#diff-c942bb45306f90f41bcaffbfcfecbcde32862b69523089e67cec126b30328451L40

packages/terra-functional-testing/tests/jest/commands/expect/toMatchReference.test.js
BenBoersma: I think we're better off with this defined in the beforeEach. While setting up each test individually could be clearer, as far as I am aware the current planned design for `ignoreScreenshotMismatch` is pretty simple and only really effects one specific code decision point. If this was handing off an object or something capable of more variance I feel like setting it up in each unit test could be a better solution. We only are ever going to need it to be `true` instead of `false`  in the one example jest test that has been added and in the case where it is `true` it changes a single logic point. If this touched on more logic or was planned to touch more logic (within this test file) I could see the need to make more of a case by case basis similar to how the `receivedScreenshot` object is handled in the rest of the file. Also not repeating code is always nice.

yuderekyu: Sure, let's follow your outline then - if we only need to test a single use case where it's set to false, and the rest of use cases are true, how about using `beforeAll`? 

Rather than re-setting a global value to false for every test, we'd use `beforeAll` to default it to false, then set and reset it within the unit test that needs it to be true. This eliminates the redundant setup that executes for every single block. What do you think?

BenBoersma: That works for me. Made the changes here https://github.com/cerner/terra-toolkit/pull/655/commits/ac8bd511f4e95ae21e42259858cebb7bc8e4ca95

yuderekyu (approved):

benbcai (approved):

mjhenkes:
Update docs in terra-toolkit-docs to include the new cli option.

tbiethman (approved):

BenBoersma:
> Update docs in terra-toolkit-docs to include the new cli option.

Updated here https://github.com/cerner/terra-toolkit/pull/655/commits/105e5ecde503e0891ba85d9f38c6d85d9603ea8a

mjhenkes (approved):

packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/wdio-services.4/VisualRegressionService.tool.mdx
```diff
@@ -4,7 +4,7 @@ import { Notice } from '@cerner/terra-docs';
 
 Visual regression testing for WebdriverIO.
 
-Based on the work of Jan-André Zinser on [wdio-visual-regression-service](https://github.com/zinserjan/wdio-visual-regression-service) and [wdio-screenshot](https://github.com/zinserjan/wdio-screenshot). The visual regression service performs image comparison evaluations to determine mismatch tolerances. Screenshots exceeding the mismatch tolerance will fail the test run.
+Based on the work of Jan-André Zinser on [wdio-visual-regression-service](https://github.com/zinserjan/wdio-visual-regression-service) and [wdio-screenshot](https://github.com/zinserjan/wdio-screenshot). The visual regression service performs image comparison evaluations to determine mismatch tolerances. Screenshots exceeding the mismatch tolerance will fail the test run. Screenshot failure due to mismatch tolerance can be disabled with the `--ignoreScreenshotMismatch` option.
```
benbcai: nit: Should we clarify that consumers generally should not use --ignoreScreenshotMismatch just to get a test run to pass. The intention of this option is for a dry run or for CI so consumers don't abuse this option.

BenBoersma: As per offline discussion, reworded the sentence a here https://github.com/cerner/terra-toolkit/pull/655/commits/7b22a7c62d5b4a236bb47def840c973e0e08289b

PR #658, opened 2021-05-20
[terra-functional-testing] Add configuration to retrieve and unzip screenshots from registry

### LINKED ISSUE #631: [terra-functional-testing] Add a configuration option that accepts a url to a screenshot registry for the project
# Story

## Description

As a consumer of terra, I want our functional testing configuration to be able to accept a url for a location from which to download screenshots

## Motivation

So that we can rely on screenshots that are not located in github

## Definition of Done

We'll know we're done when functional testing accepts a CLI option that:
* Indicates a zip file to be downloaded before the testing process starts
* When passed the option, downloads that zip file, and unzips it

An example of what the url would look would be: http://repo.nightly.cerner.corp/nexus/content/sites/ion-nightly-site/com.cerner.orion/orion-release-site/master/screenshots.zip

benbcai:
### Summary
Added configuration to retrieve and unzip screenshots from registry.

Closes #631 

### Additional Details
- Added 'screenshotUrl` CLI option that takes a url to a registry to download the screenshots.
- If the 'screenshotUrl` option is not provided, it will then check the `screenshots` config in wdio.conf.js for the url.
- If a url is provided, it will download and unzip the screenshots into the  `terra-wdio-screenshots` folder in the current working directory.

@cerner/terra


packages/terra-functional-testing/src/commands/utils/downloadScreenshots.js
```diff
@@ -0,0 +1,46 @@
...
+ * @param {string} screenshotUrl - The url to the repository that stores the screenshots.
+ */
+async function downloadScreenshots(screenshotUrl) {
+  if (!screenshotUrl) {
+    logger.warn('A url to the screenshot registry is required to download screenshots.');
+    return;
+  }
+
+  await new Promise((resolve, reject) => {
+    axios({
```
yuderekyu: does axios support async/await? not a big deal, but we could switch wholesale to that, instead of mixing in `then` statements

packages/terra-functional-testing/package.json
```diff
@@ -54,9 +54,11 @@
     "@wdio/spec-reporter": "^6.0.16",
     "@wdio/sync": "^6.1.14",
     "axe-core": "4.0.2",
+    "axios": "^0.18.1",
```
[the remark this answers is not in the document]
benbcai: I think I can use [https.get](https://nodejs.org/docs/latest-v12.x/api/https.html#https_https_get_options_callback). I'll explore that.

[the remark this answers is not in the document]
yuderekyu: [`node-fetch`](https://github.com/node-fetch/node-fetch) and [`got`](https://github.com/sindresorhus/got) are good alternatives too. I know we've used node-fetch for some internal projects

packages/terra-functional-testing/src/commands/utils/downloadScreenshots.js
benbcai: I use `then` to be consistent with our other projects and also was following the [axios API example](https://github.com/axios/axios#axios-api).

yuderekyu: well, the axios api also shows an example using async/await 🤷🏼 https://github.com/axios/axios#example

packages/terra-functional-testing/package.json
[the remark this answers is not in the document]
benbcai: Updated to use native Node https: https://github.com/cerner/terra-toolkit/pull/658/commits/ce267b6c469523fe48fe90bd79823b4b6fcca5c1

mjhenkes:
Can the screenshot URL be set in the wdio config too? I think it should be able to be.

Also update docs in terra-toolkit-docs to include the new option.

benbcai:
> Can the screenshot URL be set in the wdio config too? I think it should be able to be.
> 
> Also update docs in terra-toolkit-docs to include the new option.

Yes, the url can also be set in wdio.config.js. [Here](https://github.com/cerner/terra-toolkit/pull/658/files#diff-f13ec26e86b6e8d765ec9020014d38600b7df8643e37a26d40000d28807e2bfbR141-R150) is where it reads the config if the CLI option is not provided.

```
screenshot: {
   url: 
}
```

I intentionally didn't add any doc for this because this feature is not yet usable. My thought was to add full documentation once it is feature complete. Do you think I should at least add the CLI option for now or just wait until everything is ready?

mjhenkes (approved):
@benbcai waiting for everything to be ready seems like a good plan. Thanks for the explanation.

tbiethman (approved):

packages/terra-functional-testing/src/commands/utils/downloadScreenshots.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+async function downloadScreenshots(screenshotUrl) {
+  if (!screenshotUrl) {
+    logger.error('A url to the screenshot registry is required to download screenshots.');
+    return;
+  }
+
+  await new Promise((resolve, reject) => {
+    https.get(screenshotUrl)
+      .on('response', async (response) => {
```
yuderekyu, line 21: I'm not very familiar with the https api - do we need to handle a situation where an error occurs before the `response` event is triggered?

benbcai: It should always return a response with a statusCode which will indicate whether or not it succeeded or not.

yuderekyu (approved):

Issue #634, opened 2021-04-20
[terra-functional-testing] Update the report writer to write the contents to a file


ryanthemanuel:
# Story

## Description

As a consumer of terra, I want to have my output written to a file (similar to what was done with the reporter in our old toolkit https://github.com/cerner/terra-toolkit-boneyard/blob/main/reporters/wdio/TerraWDIOSpecReporter.js)

## Motivation

So that this report can be utilized by a release site similarly to how the old toolkit's report was utilized

## Definition of Done

We'll know we're done when you can run wdio tests and the report can be pulled into our internal release site exactly the same way that things were done when toolkit was used.

PR #662, opened 2021-05-26
Added FileOutputReporter reporter

sirishaNekkanti:
### Summary
Wdio File Output Reporter is a reporter that logs wdio test output to separate files based on locale, theme, and form-factor
addresses:  https://github.com/cerner/terra-toolkit/issues/634


package.json
```diff
@@ -68,7 +68,7 @@
     "heroku-prebuild": "npx lerna init",
     "jest": "jest",
     "lint:js": "eslint --ext .js,.jsx .",
-    "lint": "npm run lint:js",
+    "lint": "npm run lint:js -- --fix",
```
yuderekyu: I'm not sure if we want to default to automatically fixing lints; my thinking is to leave this be so developers can see what errors occur and then decide what action to take. I'm curious what others think

package.json
```diff
@@ -118,5 +118,8 @@
     "webpack": "^4.43.0",
     "webpack-cli": "^3.3.12",
     "webpack-dev-server": "^3.11.0"
+  },
+  "dependencies": {
+    "strip-ansi": "^6.0.0"
```
yuderekyu: is this dependency necessary for the whole monorepo? we should scope this down to the package that uses this

packages/terra-functional-testing/src/reporters/fileOutput-reporter/index.js
```diff
@@ -0,0 +1,220 @@
...
+
+        if (!this.resultJsonObject.output[this.moduleName]) {
+          this.resultJsonObject.output[this.moduleName] = [];
+        }
+        let readableMessage = this.getMessage(runner);
+        const readableArrayMessage = readableMessage.toString().split(',');
+        for (let i = 0; i < readableArrayMessage.length; i + 1) {
+          readableArrayMessage[i] = stripAnsi(`${readableArrayMessage[i]}${endOfLine}`);
+        }
+        readableMessage = readableArrayMessage.join('');
```
yuderekyu: how about being declarative and using map:
```suggestion
        const messages = readableArrayMessage.map((message) => stripAnsi(`${message}${endOfLine}`).join('');
```

packages/terra-functional-testing/src/reporters/fileOutput-reporter/index.js
```diff
@@ -0,0 +1,220 @@
...
+    this.hasResultsDir();
+  }
+
+  /**
+   * Sets results directory for the test run. Uses the wdio reporterOptions.outputDir if set, otherwise
+   * it outputs to tests?/wdio/reports.
+   * @return null;
+   */
+  setResultsDir() {
+    const { reporterOptions } = this.options;
```
yuderekyu: can you walk through how `this.options` is being propagated and set? Is it coming from the spec reporter that we're extending? 

I tried to dig into this, and my understanding is when we specify what reporters to use inside configuration, we can pass an option as the second argument, and that options object will propagate into this reporter. Is that totally wrong? 

```js
// wdio config
reporters: [[fileOutputReporter, { dir: 'full-stack' }]],
```

```js
// file output reporter
const { dir } = this.options
```

packages/terra-functional-testing/src/reporters/fileOutput-reporter/index.js
```diff
@@ -0,0 +1,220 @@
...
+    this.printReport(runner);
+  }
+
+  getMessage(runner) {
+    const preface = `[${this.getEnviromentCombo(
+      runner.capabilities,
+      false,
+      runner.isMultiremote,
+    ).trim()} #${runner.cid}]`;
+    const divider = '------------------------------------------------------------------';
```
yuderekyu: I vote we define the preface and divider variables right before they are used. That way it's easier to keep track of the scope of variables

package.json
sirishaNekkanti: I deleted it.

package.json
sirishaNekkanti: I moved the dependency to terra-functional-testing

packages/terra-functional-testing/src/reporters/fileOutput-reporter/index.js
sirishaNekkanti: yes you are right options come from the configurations, in this case we are not using them so, I removed that section

package.json
sirishaNekkanti: https://github.com/cerner/terra-toolkit/pull/662/files#diff-72c2847b50d4152bd910ef25c33dafbd31a04d507329a333036e6226e5e8ef71R77

packages/terra-functional-testing/src/reporters/fileOutput-reporter/index.js
sirishaNekkanti: https://github.com/cerner/terra-toolkit/pull/662/files#diff-5aca2141ee738df380d2de0da498a0a17727d935d42de74bbbb00a3046b1febeR41-R47

packages/terra-functional-testing/src/reporters/fileOutput-reporter/index.js
sirishaNekkanti: https://github.com/cerner/terra-toolkit/pull/662/files#diff-5aca2141ee738df380d2de0da498a0a17727d935d42de74bbbb00a3046b1febeR145-R147

packages/terra-functional-testing/src/reporters/fileOutput-reporter/index.js
sirishaNekkanti: https://github.com/cerner/terra-toolkit/pull/662/files#diff-5aca2141ee738df380d2de0da498a0a17727d935d42de74bbbb00a3046b1febeR168-R169

packages/jest-config-terra/src/reporters/verbose-reporter/TerraVerboseReporter.js
```diff
@@ -0,0 +1,116 @@
...
+  log(message) {
+    const readableMessage = `${stripAnsi(message)}${endOfLine}`;
+    this.setTestModule(readableMessage);
+    const { moduleName } = this;
+    if (!this.results.output[moduleName]) {
+      this.results.output[moduleName] = [];
+    }
+    if (!this.unformattedResult[moduleName]) {
+      this.unformattedResult[moduleName] = [];
+    }
```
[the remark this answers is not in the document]
sirishaNekkanti: it seems like the jest reporter files are added to this pr. I will open a new clean pr in few minutes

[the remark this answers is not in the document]
yuderekyu: can you close this one then?

Issue #611, opened 2021-03-25
[terra-functional-testing] Improve selenium grid connection logging


benbcai:
# Story

## Description

As a maintainer of Terra I want to log a more meaningful message when there are selenium grid connection issues

## Motivation

So that when such error occurs, the error message will be helpful for troubleshooting.

## Definition of Done

We'll know we're done when a PR with the change is merged.

## Additional Details.

#602 is the spike to investigate how best to detect such connection issues.

PR #674, opened 2021-06-18
[terra-functional-testing] Wdio7 Upgrade

jmsv6d:
### Summary

Upgrade to wdio7, there were a few changes to the reporter implementation that was causing our reporters to fail.
1) The @wdio/reporter class now checks for existence of a write stream in the options object before assigning the stdout write stream. This is created by wdio when initiating reporters for the test run, but was missing in our jest tests.
2) The implementation of the runners object was changed to have the key `config` storing the `options` object instead of the `config`. This seems like a bug and logging an issue to wdio, will post here when it's logged. I found a workaround for now by getting the config from the browser object instead.

### Additional Details

@cerner/terra


tbiethman (approved):

jmsv6d:
terra-framework test PR: https://github.com/cerner/terra-framework/pull/1423

packages/terra-functional-testing/package.json
```diff
@@ -47,12 +47,12 @@
   "dependencies": {
```
benbcai, line 47: Should we bump the supported Node versions [here](https://github.com/cerner/terra-toolkit/pull/674/files#diff-72c2847b50d4152bd910ef25c33dafbd31a04d507329a333036e6226e5e8ef71R10) to include 14?

`"node": ">=10 <15"`

packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/upgrade-guides.5/version-2-upgrade-guide.2.tool.md
```diff
@@ -0,0 +1,25 @@
+# Terra Functional Testing - Version 2 Upgrade Guide
+
+The only change in this version is the upgrade from wdio v7. This should cause very little disruption to consuming projects.
```
benbcai:
```suggestion
The only change in this version is the upgrade to WebDriverIO v7. This should cause very little disruption to consuming projects.
```

packages/terra-functional-testing/package.json
benbcai: Actually we should exclude Node 10 since V7 [dropped Node 10 support](https://webdriver.io/blog/2021/02/09/webdriverio-v7-released/#drop-node-v10-support).

`"node": ">=12 <15"`

packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/upgrade-guides.5/version-2-upgrade-guide.2.tool.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Breaking Changes
+
+### WDIO v6 to v7
+
+A complete list of breaking changes can be found here:
+
+- [WebdriverIO v7 Breaking Changes](https://github.com/webdriverio/webdriverio/blob/main/CHANGELOG.md#boom-breaking-change)
+- [WebdriverIO v6 to v7 Upgrade Guide](https://github.com/webdriverio/webdriverio/blob/main/CHANGELOG.md#boom-breaking-change)
```
benbcai: This link is the same as above. Did you intend to link to https://webdriver.io/docs/v7-migration?

packages/terra-functional-testing/package.json
jmsv6d: Good point. Updated.

package.json
```diff
@@ -4,7 +4,7 @@
   "version": "5.25.0",
   "description": "Utilities to help when developing terra modules.",
   "engines": {
-    "node": ">=10 <13"
+    "node": ">=12 <15"
```
benbcai: Should this change be made in [package.json](https://github.com/cerner/terra-toolkit/pull/674/files#diff-72c2847b50d4152bd910ef25c33dafbd31a04d507329a333036e6226e5e8ef71) in terra-functional-testing instead of here?

benbcai (approved):

mjhenkes:
We had talked about reverting this change https://github.com/cerner/terra-toolkit/pull/673 and changing the size of test break points as part of v2, is that still something we want to do?

benbcai:
> We had talked about reverting this change #673 and changing the size of test break points as part of v2, is that still something we want to do?

Yes, that is still something we want to do. Should we do that on a separate story to avoid increasing the scope of this? And we should do that before we release 2.0.0.

mjhenkes:
@benbcai, if we don't include that change as part of this story we'll be putting terra-toolkit in an unreleasable state because we wouldn't want to push out the next major version without one of the breaking changes.

benbcai:
@mjhenkes yeah, that is correct. I think we have two options:

1. Reverting #673 and adjusting the size of test break points as part of this PR. This PR is being blocked by the work on #611 so we can't merge into main just yet.
2. Merge this PR into a "V2" branch and then do the revert. Once it is all done, then merge into main. This option would avoid the block from #611.

mjhenkes:
I think either option works but i lean towards option 1 since this is such a small change (relatively)

benbcai:
@mjhenkes Sounds good. Let's go with option 1.

@jmsv6d These should be the additional changes needed:
1. Revert everything on #673 
2. Increase the width of all of these [test viewports](https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/constants/index.js#L2-L25) by 5 pixels. I believe @BenBoersma increased the width for the large viewports from `1000` to `1005` and that was enough to account for the scrollbar in IE. Ben, correct me if I'm wrong but I'll do some additional testing to make sure 5 pixels is enough.

benbcai:
Looks like we need to increase the width of all viewports to at least 8px.

Here are screenshots in IE showing the large viewport increased by 7 px from 1000 to 1007 and still shows the breakpoint at `medium`. When it is increased by 8px to 1008, then it shows the large breakpoint.

viewport-size

packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/upgrade-guides.5/version-2-upgrade-guide.2.tool.md
```diff
@@ -0,0 +1,25 @@
...
+The primary change you will need to make is updating the `@cerner/terra-functional-testing` version in your `package.json`:
+
+```diff
+// package.json
+{
+  "devDependencies": {
++   "@cerner/terra-functional-testing": "^2.0.0",
+-   "@cerner/terra-functional-testing": "^1.0.0"
+  }
+}
```
benbcai, line 31: We should add a note stating that the width of all viewport sizes have increased and will lead to screenshot mismatches. All existing screenshots would need to be regenerated.

benbcai:
After some more testing, I think the only viewport that we need to adjust the width is `large`. The width for all other viewports can stay the same to minimize the number of screenshot regeneration.

These are the breakpoints defined in [terra-breakpoint](https://engineering.cerner.com/terra-ui/components/terra-breakpoints/breakpoints/about#breakpoints).

```
const breakpoints = {
  tiny: 0,
  small: 544,
  medium: 768,
  large: 992,
  huge: 1216,
  enormous: 1440,
};
```

These are the current test [viewport sizes](https://engineering.cerner.com/terra-ui/dev_tools/cerner/terra-functional-testing/about#form-factors):

```
const TERRA_VIEWPORTS = {
  tiny: {
    width: 470,
    height: 768,
  },
  small: {
    width: 622,
    height: 768,
  },
  medium: {
    width: 838,
    height: 768,
  },
  large: {
    width: 1000,
    height: 768,
  },
  huge: {
    width: 1300,
    height: 768,
  },
  enormous: {
    width: 1500,
    height: 768,
  },
};
```

The large viewport is the only one which width is only 8px more than the large breakpoint. All other viewports have a width that is significantly more than the minimum breakpoint width. After running all 6 viewports in chrome, Firefox, and IE using the original viewport width the only viewport that does not render in the expected breakpoint is `large` in IE. The large viewport in chrome and Firefox is correctly rendered in the large breakpoint.

Given this, I suggest that we increase the width only for the large viewport and keep the width for the other viewports unchanged to minimize the impact. Increate the width for the large viewport from 1000 to 1008 appears wide enough but it doesn't hurt to increase it a little more (e.g. 1020) just to give it some additional width to account for some unexpected circumstance.

packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/upgrade-guides.5/version-2-upgrade-guide.2.tool.md
benbcai: We should also update the doc for the viewport sizes [here](https://github.com/cerner/terra-toolkit/blob/main/packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/About.1.tool.mdx#form-factors).

packages/terra-functional-testing/CHANGELOG.md
```diff
@@ -2,11 +2,14 @@
...
 
+* Breaking
+  * Upgraded to wdio v7
+  * Increased large viewport size to `1020`
+
 ## 1.9.0 - (June 18, 2021)
 
 * Fixed
   * Updated screenshot cleaning to match nested snapshot directories
-  * Updated to size the test page to the viewport size instead of the browser's window size to correctly match the corresponding breakpoint. This change may affect the size of existing screenshots, particularly in IE and firefox. The affected screenshots may need to be regenerated.
```
benbcai: I think we should keep the changelog instead of removing it since we did make this change on version 1.9.0 and it would be helpful to let consumers know that 1.9.0 would cause screenshot mismatches.

packages/terra-functional-testing/src/services/wdio-visual-regression-service/modules/getTerraFormFactor.js
```diff
@@ -1,5 +1,26 @@
...
+      screenWidth: Math.max(document.documentElement.clientWidth, window.innerWidth || 0),
+      screenHeight: Math.max(document.documentElement.clientHeight, window.innerHeight || 0),
+    };
+  });
+
+  return {
+    width: resolution.screenWidth,
+    height: resolution.screenHeight,
+  };
+}
```
mjhenkes: Could we just do this?
```suggestion
function getViewportSize() {
  // eslint-disable-next-line prefer-arrow-callback
  return global.browser.getWindowSize();
}
```

benbcai: I think we could and probably should since we use [setWindowSize](https://github.com/cerner/terra-toolkit/pull/674/files#diff-cf781d47655d684c44cf4d2d28591066268e50b8f239a7250e0b0228ded7c9ffR18) to set the window size so calling `getWindowSize` would be more accurate and less code/complex. I also think changing to `getWindowSize` has no impact on what "formFactor" is returned from `getTerraFormFactor` below where this is being used.

If we wanted to we could even get rid of this function and just call `global.browser.getWindowSize().width` directly in [here](https://github.com/cerner/terra-toolkit/pull/674/files#diff-a204538639999723fcde3efa5da7ab19229d476d2d255839f5f9671610cafdadR32) but the proposed change is just fine.

benbcai (approved):

mjhenkes (approved):

PR #679, opened 2021-07-08
[terra-functional-testing] Improve selenium grid logging

### LINKED ISSUE #611: [terra-functional-testing] Improve selenium grid connection logging

BenBoersma:
### Summary

Added a check after compiling to ping a provided gridUrl's status page to see if it is ready. If no gridUrl is provided we just resolve. 
Closes https://github.com/cerner/terra-toolkit/issues/611

### Additional Details

@cerner/terra


packages/terra-functional-testing/src/webpack-server/webpack-server.js
```diff
@@ -11,6 +13,7 @@ class WebpackServer {
     this.config = WebpackServer.config(options);
     this.host = host || '0.0.0.0';
     this.port = port || '8080';
+    this.gridUrl = options.gridUrl ? `http://${options.gridUrl}:80/status` : undefined;
```
benbcai: Should we rename `this.gridUrl` to something like `this.gridStatusUrl` or `this.gridHealthUrl` to avoid confusion with the existing `gridUrl` CLI option that is typically `grid314.test.devcernerpowerchart.com`?

packages/terra-functional-testing/src/webpack-server/webpack-server.js
```diff
@@ -73,8 +76,30 @@ class WebpackServer {
...
+              res.on('end', () => {
+                if (!res.complete) throw new SevereServiceError(`${this.gridUrl} connection was terminated while the message was still being sent`);
+                if (JSON.parse(response).value.ready) {
+                  resolve();
+                } else {
+                  throw new SevereServiceError(`${this.gridUrl} failed to return a ready response. Check to ensure the selenium grid is stable`);
+                }
+              });
+            } else {
+              throw new SevereServiceError(`Url ${this.gridUrl} returns status code of ${statusCode}.`);
```
benbcai: In the case of a failure do we need to call resume() on the response to free up memory as indicated by its [documentation](https://nodejs.org/api/http.html#http_http_get_url_options_callback)?

benbcai: Also for failure cases don't we need to reject the promise?

packages/terra-functional-testing/src/webpack-server/webpack-server.js
```diff
@@ -73,8 +76,30 @@ class WebpackServer {
...
+            if (statusCode >= 200 && statusCode <= 299) {
+              res.on('data', (chunk) => {
+                response += chunk;
+              });
+              res.on('end', () => {
+                if (!res.complete) throw new SevereServiceError(`${this.gridUrl} connection was terminated while the message was still being sent`);
+                if (JSON.parse(response).value.ready) {
+                  resolve();
+                } else {
+                  throw new SevereServiceError(`${this.gridUrl} failed to return a ready response. Check to ensure the selenium grid is stable`);
```
benbcai: I like this message that provides a suggestion to check to make sure the remote selenium grid is stable. For the other failure cases, do we know what may be some of the common causes and provide a suggestion?

For examples:
- For the failure on line 101, is that failure most likely due to not having a valid IP address (WDIO_EXTERNAL_HOST)? If so can we add a suggestion to ensure WDIO_EXTERNAL_HOST is correct/provided.
- For the failure on line 90, when would that occur? Would the failure go away by rerunning the tests again?

BenBoersma: So for the failure on line 101, that usually a failure when we get a timeout response or a connection refused (so far) which is usually due to vpn issues. Would we want something like 

> "Failed to connect to url ${this.gridUrl}, make sure you have access. Error thrown: ${e.message}

As for line 90 I'm not sure exactly when that would  occur but it was present in most of the examples I was looking at. I think it would more likely to only happen in larger responses that would be broken up into multiple responses by `http` which I think theoretically shouldn't be present in a response from the status page of a gridUrl. I'm not sure if the best option is to just remove it or leave as a just in case bit of defensive coding.

packages/terra-functional-testing/src/webpack-server/webpack-server.js
BenBoersma: Changed to `gridStatusUrl` https://github.com/cerner/terra-toolkit/pull/679/commits/900b6e80e6616822114af05e0ef367660049fb53#diff-73928b81969e1fbd0b7fed98d5dc39d99e26fd5d75ace8ef586e3a4b2e25e0ffR15

packages/terra-functional-testing/src/webpack-server/webpack-server.js
BenBoersma: So added resume and rejects here https://github.com/cerner/terra-toolkit/pull/679/commits/900b6e80e6616822114af05e0ef367660049fb53#diff-73928b81969e1fbd0b7fed98d5dc39d99e26fd5d75ace8ef586e3a4b2e25e0ffR101

the `resume()` seems to mostly just free up the memory earlier but using the above documentation and this [tutorial](https://www.digitalocean.com/community/tutorials/how-to-create-an-http-client-with-core-http-in-node-js) it looks like we can just place in the error message with the status code reply and good. Feel free to comment and offer suggestions for better implementation.

packages/terra-functional-testing/src/webpack-server/webpack-server.js
benbcai: Would we get the failure on line 101 if connected to vpn but with an invalid IP address? One common failure is consumers are on the vpn but either didn't provide an IP address or the IP address is invalid. If an invalid IP address can lead to the failure on line 101, it would be helpful to provide a concrete suggestion since consumers may not know what to do/try when they see "make sure you have access".

Something along the line of:

> Failed to connect to url ${this.gridUrl}. Ensure you are connected to the network and/or a valid IP address is provided using WDIO_EXTERNAL_HOST. Error thrown: ${e.message}

I think of these failure messages as a way to inform what went wrong so consumers/we know what to do to attempt to fix it.

BenBoersma: No we wouldn't get that failure. I haven't been able to find a way to successfully test that particular instance of the WDIO_EXTERNAL_HOST being wrong, that also works on Jenkins.

packages/terra-functional-testing/src/webpack-server/webpack-server.js
```diff
@@ -73,8 +75,37 @@ class WebpackServer {
...
+          let response = '';
+          http.get(this.gridStatusUrl, (res) => {
+            const { statusCode } = res;
+            if (statusCode >= 200 && statusCode <= 299) {
+              res.on('data', (chunk) => {
+                response += chunk;
+              });
+              res.on('end', () => {
+                if (!res.complete) {
+                  logger.error(`${this.gridStatusUrl} connection was terminated while the message was still being sent`);
```
benbcai:
```suggestion
                  logger.error(`${this.gridStatusUrl} connection was terminated while the message was still being sent.`);
```

packages/terra-functional-testing/src/webpack-server/webpack-server.js
```diff
@@ -73,8 +75,37 @@ class WebpackServer {
...
+                  reject();
+                }
+                if (JSON.parse(response).value.ready) {
+                  resolve();
+                } else {
+                  logger.error(`${this.gridStatusUrl} failed to return a ready response. Check to ensure the selenium grid is stable`);
```
benbcai:
```suggestion
                  logger.error(`${this.gridStatusUrl} failed to return a ready response. Check to ensure the selenium grid is stable.`);
```

packages/terra-functional-testing/tests/jest/webpack-server/webpack-server.test.js
```diff
@@ -89,6 +103,270 @@ describe('Webpack Server', () => {
...
+      };
+      webpack.mockReturnValue(compiler);
+
+      const server = new WebpackServer();
+      server.gridStatusUrl = '1.1.1.1:8080';
+      await expect(server.start()).rejects.toBeUndefined();
+      expect(mockLoggerInstance.error.mock.calls[0][0]).toEqual(`${server.gridStatusUrl} failed to return a ready response. Check to ensure the selenium grid is stable`);
+      // expect.assertions(2);
+      // return expect(server.start()).rejects.toBe(false);
+      // return server.start().catch(e => expect(e.message).toEqual(`${server.gridStatusUrl} failed to return a ready response. Check to ensure the selenium grid is stable`));
```
benbcai: Are these needed?

BenBoersma: They are not, removed here https://github.com/cerner/terra-dev-site/pull/367

packages/terra-functional-testing/src/webpack-server/webpack-server.js
BenBoersma: Fixed here https://github.com/cerner/terra-toolkit/pull/679/commits/ffe222c2f578f8720b7bf7d64c4137516d1ba731#diff-73928b81969e1fbd0b7fed98d5dc39d99e26fd5d75ace8ef586e3a4b2e25e0ffR90

packages/terra-functional-testing/src/webpack-server/webpack-server.js
BenBoersma: Fixed here https://github.com/cerner/terra-toolkit/pull/679/commits/ffe222c2f578f8720b7bf7d64c4137516d1ba731#diff-73928b81969e1fbd0b7fed98d5dc39d99e26fd5d75ace8ef586e3a4b2e25e0ffR96

benbcai (approved):
I'm ok with these logging messages if we don't have an easy way to determine the common cause to provide better suggestions. We can always improve on these as we go.

mjhenkes (approved):

PR #691, opened 2021-08-07
[terra-functional-testing] Add parameters for locating selenium service

QuasiGnarus:
### Summary
Terra functional testing seems to be designed assuming that selenium is either running in docker selenium hub at localhost:4444 or at http://gridUrl:80. I need to run the docker selenium hub at a specified IP address (still port 4444). In my case Windows is the laptop host OS and I do my terra/orion development in an Ubuntu VM in VirtualBox. It's not possible to run docker inside a VM, so I run docker in a second VM with its own IP address.

I think this is a good approach but I'm open to further discussion. We haven't upgraded to terra-functional-testing v2 yet, so I would like this released for both 1.x and 2.x

### Additional Details

@cerner/terra


packages/terra-functional-testing/src/config/utils/getConfigurationOptions.js
```diff
@@ -29,8 +31,8 @@ const getConfigurationOptions = (options) => {
   return {
     baseUrl: `http://${externalHost || getIpAddress()}:${externalPort || 8080}`,
     capabilities: getCapabilities(browsers, !!gridUrl),
-    hostname: gridUrl || (useSeleniumStandaloneService ? 'standalone-chrome' : 'localhost'),
-    port: gridUrl ? 80 : 4444,
+    hostname: dockerUrl || gridUrl || (useSeleniumStandaloneService ? 'standalone-chrome' : 'localhost'),
+    port: dockerPort || (gridUrl ? 80 : 4444),
```
mjhenkes: Adding a `port` param is reasonable, but i'm not sure what the difference is between the `dockerUrl` and `gridUrl`.

```suggestion
    hostname: gridUrl || (useSeleniumStandaloneService ? 'standalone-chrome' : 'localhost'),
    port: port || (gridUrl ? 80 : 4444),
```

mjhenkes:
Terra-functional-testing v2 is a relatively easy upgrade. I'd rather you take a look at upgrading than patch back to terra-functional-testing v1.

https://engineering.cerner.com/terra-ui/dev_tools/cerner-terra-toolkit-docs/terra-functional-testing/upgrade-guides/version-2-upgrade-guide

packages/terra-functional-testing/src/config/utils/getConfigurationOptions.js
QuasiGnarus: I tried using gridUrl. There's logic elsewhere in the package tied to gridUrl to set http protocol and perform status checks etc, which prevents me from using it.

QuasiGnarus:
> 
> 
> Terra-functional-testing v2 is a relatively easy upgrade. I'd rather you take a look at upgrading than patch back to terra-functional-testing v1.
> 
> https://engineering.cerner.com/terra-ui/dev_tools/cerner-terra-toolkit-docs/terra-functional-testing/upgrade-guides/version-2-upgrade-guide

We're on board with switching to v2. I'm working now on the upgrade. I've rebased from 1.10.0 tag to main branch. The connection to selenium hub is timing out after the upgrade, so there are probably additional changes required.

QuasiGnarus:
> 
> 
> Terra-functional-testing v2 is a relatively easy upgrade. I'd rather you take a look at upgrading than patch back to terra-functional-testing v1.
> 
> https://engineering.cerner.com/terra-ui/dev_tools/cerner-terra-toolkit-docs/terra-functional-testing/upgrade-guides/version-2-upgrade-guide

@mjhenkes We are upgrading our repo to node 12 and terra-functional-testing 2 to make this easier. I've worked out the bugs on my end and rebased to main branch. Regarding gridUrl vs dockerUrl, I agree that it's not readily apparent why there would need to be two separate options for that. Prior to terra-functional-testing it was possible to set the docker IP address directly in the wdio.conf.js config structure. In terra-functional-testing gridUrl would seem at first glance to be doing that, but the business logic tied to gridUrl makes it only useful for connecting to a selenium grid server, not docker selenium hub. Instead of adding more business logic everywhere that gridUrl is mentioned, I think it's cleaner just to add a new parameter that simply sets the IP address with no other logic. 

On that note, I think it might help differentiate gridUrl vs dockerUrl if I renamed the new parameter to be 'dockerHost'. How do you feel about that?

mjhenkes:
@QuasiGnarus, you are indeed correct about the mulit use of gridUrl. How about we go with 'port' and 'seleniumServiceUrl' then we're a bit more specific about how the url is used. I'd like to abstract docker out of the variable names because that the implementation of how we start the selenium service.

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -99,23 +99,23 @@ const cli = {
...
           }
 
           return undefined;
         },
       },
-      dockerPort: {
+      port: {
         type: 'number',
-        describe: 'The port mapping for docker selenium hub.',
+        describe: 'The port mapping for selenium service.',
```
mjhenkes:
```suggestion
        describe: 'The port mapping for selenium service or the external selenium grid.',
```

mjhenkes:
We will also want to update the terra-functional-testing docs: https://github.com/cerner/terra-toolkit/blob/main/packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/About.1.tool.mdx#options

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -99,6 +99,28 @@ const cli = {
...
+
+          return undefined;
+        },
+      },
+      port: {
+        type: 'number',
+        describe: 'The port mapping for the selenium service or the external selenium grid.',
+        default: () => {
+          if (process.env.SELENIUM_SERVICE_PORT) {
+            return process.env.SELENIUM_SERVICE_PORT;
```
QuasiGnarus: @mjhenkes OK with how I've named this env variable?

packages/terra-functional-testing/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Added
+  * Added seleniumServiceUrl parameter for locating selenium service.
+  * Added port parameter for specifying port mapping for the selenium service or the external selenium grid.
```
QuasiGnarus, line 7: @mjhenkes  Ok with how I've adjusted these items?

packages/terra-functional-testing/src/terra-cli/wdio/index.js
mjhenkes: I think we could get away without even offering the env's. We kept them for other variables due to precident in jenkins, but for new params we haven't been adding them.

packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/About.1.tool.mdx
```diff
@@ -46,6 +46,8 @@ package.json
 | --externalPort                   | number  |                                                  | The port mapping from the host to the container.                                                                                          |
 | --formFactors                    | array   |                                                  | A list of form factors for the test run. One of tiny, small, medium, large, huge, or enormous                                             |
 | --gridUrl                        | string  |                                                  | The remote selenium grid address.                                                                                                         |
+| --seleniumServiceUrl             | string  |                                                  | The address for the selenium service.                                                                                                     |
```
mjhenkes: Add the 'localhost' as the default value.

packages/terra-toolkit-docs/src/terra-dev-site/tool/terra-functional-testing/About.1.tool.mdx
```diff
@@ -46,6 +46,8 @@ package.json
...
+| --port                           | number  |                                                  | The port mapping for the selenium service or the external selenium grid.                                                                  |
```
mjhenkes: add 4444 as the default value (yeah it's more complicated than that but hard to explain in a little place.)

packages/terra-functional-testing/CHANGELOG.md
mjhenkes: This looks fine.

packages/terra-functional-testing/src/terra-cli/wdio/index.js
QuasiGnarus: @mjhenkes I prefer to have them because it gives me a way to specify these things in my shell startup script so I don't have to type that parameter every time I run wdio.

mjhenkes: you could alternatively set the parameters on a 'localWindowsTest' npm script

QuasiGnarus: Yeah. I think it would be better to keep user/system-specific IP addresses out of package.json.

QuasiGnarus: I think it's fair to allow the minority of users who require this configuration to have as seamless an experience as mac users etc who run docker on localhost.

mjhenkes (approved):

mjhenkes:
@QuasiGnarus, This looks good to me (with updated tests). Thank you for the contribution. I will want @benbcai to review this as well but that may not be until next week as he's OOO.

QuasiGnarus:
> 
> 
> @QuasiGnarus, This looks good to me (with updated tests). Thank you for the contribution. I will want @benbcai to review this as well but that may not be until next week as he's OOO.

I've updated the index.js wdio snapshot to include the new parameters. Hopefully that fixes the failing check. All the other tests passed without changes. I'm hesitant to add the new parameters in the testrunner.js wdio test (e.g. [here](https://github.com/cerner/terra-toolkit/blob/17b97911b538b50bcd75c46643327dbd29059f71/packages/terra-functional-testing/tests/jest/terra-cli/wdio/test-runner.test.js#L41)) because we would never pass both gridUrl and seleniumServiceUrl together. It would be possible to pass port with gridUrl.

mjhenkes:
@QuasiGnarus I'm fine with just the snapshots updated. It can be hard to test all different variations of terra-functional-testing in an automated fashion.

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -99,6 +99,28 @@ const cli = {
...
+        describe: 'The address for the selenium service.',
+        default: () => {
+          if (process.env.WDIO_SELENIUM_SERVICE_URL) {
+            return process.env.WDIO_SELENIUM_SERVICE_URL;
+          }
+
+          return undefined;
+        },
+      },
+      port: {
```
benbcai: We currently have `externalPort` as an existing option and now the name 'port' seems a little ambiguous. Can we change the name from `port` to `seleniumServicePort` to avoid any confusion and to match the SELENIUM_SERVICE_PORT env variable name?

packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -99,6 +99,28 @@ const cli = {
           return undefined;
         },
       },
+      seleniumServiceUrl: {
```
benbcai: We tend to keep these options alphabetized. Can you put them in alphabetically orders here, [here](https://github.com/cerner/terra-toolkit/pull/691/files#diff-86631c8f3fb52653295ad5b333c6724f3f393a9bec4a477083220d77c6a4c700R19-R20), and [here](https://github.com/cerner/terra-toolkit/pull/691/files#diff-bd6f2cecd6d3c745845a151947012865dd0ad4b869000d4478ffd2bbe28532ebR49-R50)?

packages/terra-functional-testing/src/terra-cli/wdio/index.js
mjhenkes: +1, sorry about that i forgot externalPort existed.

packages/terra-functional-testing/src/terra-cli/wdio/index.js
QuasiGnarus: I just noticed I put seleniumServiceUrl before seleniumServicePort because that seemed like the intuitive order for those. Do you want me to reverse them? Or should I change Url to seleniumServiceHost to be like externalHost?

benbcai (approved):

packages/terra-functional-testing/src/terra-cli/wdio/index.js
benbcai: Sure if you don't mind just to be consistent.

QuasiGnarus: @benbcai reverse them and keep Url or switch to Host?

benbcai: Sorry for my unclear response. Let's keep `seleniumServiceUrl` and just swap the ordering between seleniumServiceUrl and seleniumServicePort.

PR #704, opened 2021-09-01
[package-json-lint] Add rule for hard coded dependencies in module and devModule.

### LINKED ISSUE #697: [package-json-lint] Prevent modules and devModules from hard coding dependencies
# Story

## Description

As a consumer of a module or devModule, I want to make sure that dependencies are not hard coded.

## Motivation

So that we can leave the exact version specification to an application and avoid duplicated dependencies

## Definition of Done

We'll know we're done when the rule is created that prevents versions from being hard coded and the default configuration is specified in package-json-lint-config-terra

## Additional Information

We should only do this on devModules and modules (see #696 for how this will be specified).

pranav300:
### Summary

This PR adds the rule `require-no-hard-coded-dependency-versions`, to not allow any hardcoded dependencies in `module` or `devModule`.

Closes: #697 
### Additional Details

@cerner/terra


package.json
```diff
@@ -33,7 +33,8 @@
     "url": "https://github.com/cerner/terra-toolkit/issues"
   },
   "package-json-lint": {
-    "extends": "./packages/package-json-lint-config-terra/package-json-lint.config.js"
+    "extends": "./packages/package-json-lint-config-terra/package-json-lint.config.js",
+    "projectType": "module"
```
ryanthemanuel: This should be a devModule

packages/package-json-lint-config-terra/package-json-lint.config.js
```diff
@@ -2,5 +2,6 @@ module.exports = {
   rules: {
     'require-no-terra-base-peer-dependency-versions': 'warn',
     'require-theme-context-versions': 'warn',
+    'require-no-hard-coded-dependency-versions': 'warn',
```
ryanthemanuel: I actually think we should start this as an error out of the gate.

package.json
pranav300: Updated here https://github.com/cerner/terra-toolkit/pull/704/commits/cc6d1e0cc26383d90b787febbf2eeec56d41464a.

packages/package-json-lint-config-terra/package-json-lint.config.js
pranav300: Updated here https://github.com/cerner/terra-toolkit/pull/704/commits/cc6d1e0cc26383d90b787febbf2eeec56d41464a.

But now the build will fail due to hardcoded dependencies in these packages:
[eslint-config-terra](https://github.com/cerner/terra-toolkit/blob/main/packages/eslint-config-terra/package.json#L38)
[jest-config-terra](https://github.com/cerner/terra-toolkit/blob/main/packages/jest-config-terra/package.json#L43)
[terra-function-testing](https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/package.json#L56)

ryanthemanuel: Ahh I see. I think in the long run, we'll want consumers to be able to allow for these types of situations. Do you think it would be possible to allow for something like this in the config?

  "package-json-lint": {
    "extends": "package-json-lint.config",
    "projectType": "devModule",
    "rules": {
      "require-no-hard-coded-dependency-versions": ["error", { "allowList": "eslint-plugin-react" }]
    }
  },

pranav300: What I am thinking is, we can change the structure of severity for all the rules to something like this:
```js
"require-no-hard-coded-dependency-versions": {"severityType": "error",  "allowList": ["eslint-plugin-react"] }
```
This will allow us to have more consistent structure for all the rules. What do you think?

ryanthemanuel: Let's go with your suggestion. What I had suggested was something that eslint does, but I think we can go our own way with this.

pranav300: Updated here https://github.com/cerner/terra-toolkit/pull/704/commits/dfc30a5064aee33acfa4abd9f9fe89e768933f93

packages/package-json-lint/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ## 1.1.0 - (August 31, 2021)
 
+* Added
+  * Added new rule `require-no-hard-coded-dependency-versions`.
+
 * Added
   * Added `projectType` to allow users to flag project as `application`, `module` or `devModule`.
```
benbcai:
```suggestion
* Added
  * Added new rule `require-no-hard-coded-dependency-versions`.
  * Added `projectType` to allow users to flag project as `application`, `module` or `devModule`.
```

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
+  create: ({ ruleConfig, projectType, report }) => ({
+    dependencies: (dependencies) => {
+      if (projectType === 'module' || projectType === 'devModule') {
+        const messageString = 'no hard coded dependency';
+        const currentProblems = Object.keys(dependencies).map(dependencyName => {
+          const dependencyVersion = dependencies[dependencyName];
+          if (!dependencyVersion.startsWith('^') && !(ruleConfig.severity.allowList && ruleConfig.severity.allowList.includes(dependencyName))) {
+            return `${dependencyName}@${dependencyVersion} does not satisfy requirement for ${messageString}`;
```
benbcai: Should we update `messageString` to call out which rule it does not satisfy instead of `no hard coded dependency`?

`${dependencyName}@${dependencyVersion} does not satisfy requirement for the require-no-hard-coded-dependency-versions rule;`

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
```diff
@@ -0,0 +1,23 @@
...
+          }
+          return undefined;
+        }).filter(problem => !!problem);
+
+        if (currentProblems.length) {
+          const lintMessage = `The dependencies for this project have hard-coded versions for ${messageString}:\n  ${currentProblems.join('\n  ')}`;
```
benbcai: Similar to the above comment, this message sounds a bit confusing.
`The dependencies for this project have hard-coded versions for no hard coded dependency:`

Should we say something like:
`The dependencies for this project have hard-coded versions that violate the require-no-hard-coded-dependency-versions rule:`

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
+  create: ({ ruleConfig, projectType, report }) => ({
+    dependencies: (dependencies) => {
+      if (projectType === 'module' || projectType === 'devModule') {
+        const messageString = 'no hard coded dependency';
+        const currentProblems = Object.keys(dependencies).map(dependencyName => {
+          const dependencyVersion = dependencies[dependencyName];
+          if (!dependencyVersion.startsWith('^') && !(ruleConfig.severity.allowList && ruleConfig.severity.allowList.includes(dependencyName))) {
```
benbcai: Some project dependencies may look like this that could reference a range, the latest snapshot, or renaming the dependency where `^` is not always the first character. Is the recommendation for teams to just add such dependencies in the allowList or should the rule handle some of these less common cases?

- `"blood-loss-calculator-js": ">= 2.3.0-dev.0 <2.3.0-dev.x"`
- `"ion-provider-relationship-js": "master"`
- `"procedures-js": "dev"`
- `"real-wicg-inert": "npm:wicg-inert@^3.0.2"`

packages/jest-config-terra/package.json
```diff
@@ -40,7 +40,7 @@
     "@jest/reporters": "^25.3.0",
     "babel-jest": "^26.6.3",
     "identity-obj-proxy": "^3.0.0",
-    "jest-environment-jsdom": "26.6.2",
+    "jest-environment-jsdom": "^26.6.2",
```
mjhenkes, line 43: I could not tell you why i locked this down, or if it was a mistake. 🤷‍♂️

packages/package-json-lint/CHANGELOG.md
mjhenkes: The changelog entry should be above in the unreleased section. Probably just an auto merge error.

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
mjhenkes: We should also add documentation for each rule explaining why we set it up and what problems it helps teams avoid.

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
mjhenkes: @ryanthemanuel did you have a suggestion for a more robust version check?

packages/package-json-lint/src/rules/requireVersionSet.js
```diff
@@ -5,7 +5,7 @@ module.exports = ({
 }) => {
   const currentProblems = versionSet.map(({ name, versionRange }) => {
     const dependencyVersion = dependencies[name];
-    if (dependencyVersion && !semver.intersects(dependencyVersion, versionRange)) {
+    if (dependencyVersion && !semver.intersects(dependencyVersion, versionRange) && !(ruleConfig.severity.allowList && ruleConfig.severity.allowList.includes(name))) {
```
mjhenkes, line 8: as we add more rules will this if continue to grow? Is there a way we could clean this up?

packages/package-json-lint/CHANGELOG.md
pranav300: Updated here https://github.com/cerner/terra-toolkit/pull/704/commits/b3446c5a790a1172c0c86da818d7d21076835b53.

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
pranav300: > Should we update messageString to call out which rule it does not satisfy instead of no hard coded dependency?

Updated here https://github.com/cerner/terra-toolkit/pull/704/commits/b3446c5a790a1172c0c86da818d7d21076835b53.

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
pranav300: Updated here https://github.com/cerner/terra-toolkit/pull/704/commits/b3446c5a790a1172c0c86da818d7d21076835b53.

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
pranav300: I have added a compatibility regex array, through which we can check whether the dependency is hardcoded or not. Added here https://github.com/cerner/terra-toolkit/pull/704/commits/b3446c5a790a1172c0c86da818d7d21076835b53.

Do let me know if I missed any scenarios. Thanks

pranav300: I have added an array of regex to check for compatibility versions here https://github.com/cerner/terra-toolkit/pull/704/commits/b3446c5a790a1172c0c86da818d7d21076835b53.

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
pranav300: >We should also add documentation for each rule explaining why we set it up and what problems it helps teams avoid.

This is what I was thinking for the documentation.
Screenshot 2021-09-13 at 3 23 54 PM

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
+  create: ({ ruleConfig, projectType, report }) => ({
+    dependencies: (dependencies) => {
+      const compatibleVersionRegexList = ['dev', 'latest', 'main', 'master', /(\^|<|<=|>|>=|~)\s{0,1}\d/, /^x\.|\.x\.|\.x$/, /[\d.]{0,2}\d\s{0,1}-\s{0,1}[\d.]{0,2}\d/];
+      if (projectType === 'module' || projectType === 'devModule') {
+        const messageString = 'require-no-hard-coded-dependency-versions';
+        const currentProblems = Object.keys(dependencies).map(dependencyName => {
+          const dependencyVersion = dependencies[dependencyName];
+          const isCompatibleVersion = compatibleVersionRegexList.map(versionRegex => (!!dependencyVersion.match(versionRegex))).includes(true);
```
ryanthemanuel: I did a little digging with semver. I wonder if we can say:

const isCompatibleVersion = semver.clean(version) === null;

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
pranav300: Added documentation here https://github.com/cerner/terra-toolkit/pull/704/commits/0095a1f56d6e9489a7db304cdb5ec3013a9917c9

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
benbcai: It doesn't look like [semver.clean](https://github.com/npm/node-semver#clean) can handle ranges so syntax like `">= 2.3.0-dev.0 <2.3.0-dev.x"` or even just `2.3.0-dev.x` would not work so we may still need to handle those separately.

packages/terra-functional-testing/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added hard-coded dependency `inquirer@8.1.3`. As latest i.e. v8.1.4 is using a dependency which is supported bu node 14.
```
ShettyAkarsh:
```suggestion
  * Added hard-coded dependency `inquirer@8.1.3`. As latest i.e. v8.1.4 is using a dependency which is supported by node 14.
```

packages/package-json-lint/src/rules/require-no-hard-coded-dependency-versions.js
ryanthemanuel: I think it’ll handle them fine. We want them to return undefined or null. Basically anything but a solitary version should return null or undefined by the clean function. Thus they will be accepted

pranav300: @benbcai,
`semver.clean` works for `>= 2.3.0-dev.0 <2.3.0-dev.x`. And it will work if it has `< >` symbols. 
It only doesn't work for `2.3.0-dev.x` as this seems to be a fixed version.

If we are sure about it's going to be `dev.x or branch.x`, we can add it as an exception and have a check for it using regex.
What do you think? @ryanthemanuel @benbcai

ryanthemanuel: 2.3.0-dev.x will *only* resolve to a version that is released as 2.3.0-dev.x so this is effectively hard coded. I think it's fine to not allow this.

ryanthemanuel: e.g. 1.0.0-rc.x does not actually resolve to 1.0.0-rc.1 

From https://semver.npmjs.com/

image

benbcai: I confused myself at first but after thinking thru this again, it all makes sense now.

benbcai (approved):

ShettyAkarsh (approved):

ryanthemanuel (approved):

supreethmr (approved):

PR #710, opened 2021-09-28
[package-json-lint] Add new rule require-dependencies-declared-at-appropriate-level

### LINKED ISSUE #695: [package-json-lint] Add a package-json-lint rule for ensuring that development dependencies that are not declared as dependencies
# Story

## Description

As a consumer of a component, I'd like to ensure that development dependencies are not declared as regular dependencies

## Motivation

So that consumers of my component don't need to have any unnecessary peer dependencies met and don't have unnecessary dependencies installed

## Definition of Done

We'll know we're done when the rule is created and defaults are specified in package-json-lint-config-terra

## Additional Notes

This rule should not apply for devModules (see this issue for how that will be specified: #696).

For now we should maintain a list of dependencies that are warranted to be "development dependencies", we should go through all of our components for our dev dependencies and start with that list (things like babel, webpack, terra-dev-site, etc.)

pranav300:
### Summary

This PR adds a new rule require-no-unnecessary-dependency.
This rule doesn't allow unnecessary dependencies/peerDependencies, that can be shifted to devDependendies.

This rule is not applicable on `devModule`.

Closes #695 
### Additional Details

@cerner/terra


packages/package-json-lint/src/rules/require-no-unnecessary-dependency.js
```diff
@@ -0,0 +1,131 @@
+const devDependencySet = [
```
pranav300: I have taken this set from devDependencies in packages in terra-toolkit and terra-core. 
Please let me know if there are any packages that need to be added or removed.
Thanks.

ShettyAkarsh: can we have a separate file for this ? as the list might grow.

packages/package-json-lint/src/rules/require-no-unnecessary-dependency.js
```diff
@@ -0,0 +1,124 @@
...
+  'react-test-renderer',
+  'regenerator-runtime',
+  'shelljs',
+  'sinon',
+  'strip-ansi',
+  'stylelint',
+  'svgo',
+  'terra-application',
+  'terra-button',
+  'terra-disclosure-manager',
```
benbcai: What is the reason that these terra components, `lodash`, `commander`, etc must be a dev dependency? I would think many consumers are currently have these as regular dependencies.

packages/package-json-lint/src/rules/require-no-unnecessary-dependency.js
```diff
@@ -0,0 +1,124 @@
...
+          report({
+            lintId: messageString,
+            severity: ruleConfig.severity.severityType,
+            lintMessage,
+            projectType,
+          });
+        }
+      }
+    },
+    peerDependencies: (peerDependencies) => {
```
benbcai: Both `dependencies` and `peerDependencies` do pretty much the same thing. Is it possible to consolidate their implementations to make it more reusable for both?

packages/package-json-lint/src/rules/require-no-unnecessary-dependency.js
pranav300: I had run a script to get dev and peer dependencies from the mono-repo's. And in that some packages have these as both dependencies and devDependencies. I will go through them once and remove the unneeded dependencies from the set.

Edit: I have removed some of the packages from the set here https://github.com/cerner/terra-toolkit/pull/710/commits/d8f833e18a7d29fcd0498ab3e11edd3a89cb7c9b.

packages/package-json-lint/src/rules/require-no-unnecessary-dependency.js
pranav300: Updated here https://github.com/cerner/terra-toolkit/pull/710/commits/d8f833e18a7d29fcd0498ab3e11edd3a89cb7c9b.

packages/package-json-lint/src/rules/require-no-unnecessary-dependency.js
pranav300: I don't think that is required as later many rules may have large sets. And having a different file for each of them and maintaining them seems unnecessary.

Let me know if you think otherwise?

packages/package-json-lint/src/rules/require-no-unnecessary-dependency.js
```diff
@@ -0,0 +1,104 @@
...
+  '@cerner/terra-aggregate-translations',
+  '@cerner/terra-cli',
+  '@cerner/terra-dev-site',
+  '@cerner/terra-functional-testing',
+  '@cerner/terra-open-source-scripts',
+  '@octokit/core',
+  'bufferutil',
+  'chai',
+  'check-installed-dependencies',
+  'core-js',
```
benbcai: I would think core-js can be a regular dependency. There are several other dependencies here that I'm not familiar with but ones like `glob` and `postcss` should be allowed as a regular dependency right?

packages/package-json-lint-config-terra/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added new rule `require-no-unnecessary-dependency`.
```
ryanthemanuel: Can we rename this rule?  require-dependencies-declared-at-appropriate-level?  Something like that would be a little clearer i think.

packages/package-json-lint/src/rules/require-no-unnecessary-dependency.js
ryanthemanuel: +1 to a separate file, but I'd rather start small and grow it as we run into issues that teams typically have been doing. I think the @babel, @cerner ones below are fine. Then maybe enzyme*, jest, postcss, and webpack*?

pranav300: Created a seperate file `dev-dependency-set.json` here https://github.com/cerner/terra-toolkit/pull/710/commits/71a07f758727181729e63e57410fe1fdef2558e4.

packages/package-json-lint/src/rules/require-no-unnecessary-dependency.js
pranav300: I have updated the set and kept only @babel, @cerner, and jest dependencies.

packages/package-json-lint-config-terra/CHANGELOG.md
pranav300: Renamed the rule here https://github.com/cerner/terra-toolkit/pull/710/commits/71a07f758727181729e63e57410fe1fdef2558e4

ShettyAkarsh (approved):

packages/terra-toolkit-docs/src/terra-dev-site/tool/package-json-lint/Rules.2.tool.mdx
```diff
@@ -6,6 +6,7 @@ import { Badge } from '@cerner/package-json-lint/package.json?dev-site-package';
 
 | Rule Name | Severity Type | Description |
 | --- | --- |--- |
+| __require-dependencies-declared-at-appropriate-level__ | warn| Notifies when there are unnecessary dependencies and peerDependencies being passed in the package.json. Doesn't apply for devModule. |
```
ryanthemanuel:
```suggestion
| __require-dependencies-declared-at-appropriate-level__ | warn| Notifies when there are dependencies declared at an inappropriate level. For example babel compilers and webpack should be devDependencies and not regular dependencies. Doesn't apply for devModule. |
```

packages/package-json-lint/src/rules/require-no-unnecessary-dependency.js
benbcai: Sorry for going back and forth on this list. I think it is safe to keep the ones that we know with certainty that should be a dev dependencies, especially ones used for testing. For example, `chai`, `sinon`, `mocha`, `lerna`, `gh-pages`. I have seen projects adding `wdio-mocha-framework` as a regular dependency so anything `wdio-*` should be a dev dependency.

benbcai: I just saw [Ryan's comment](https://github.com/cerner/terra-toolkit/pull/710#discussion_r724360038) above. We can start small for now and add as go.

benbcai (approved):

ryanthemanuel (approved):

PR #718, opened 2021-10-19
Add Format js linter

### LINKED ISSUE #461: Spike: Add formatjs linter
# Feature Request

## Description
`formatjs` now has a linter available for components that utilize `react-intl`. We should investigate adding this in [eslint-plugin-formatjs](https://formatjs.io/docs/tooling/linter)

saket2403:
### Summary

Add Format js linter
Closed #461 

### Additional Details

@cerner/terra


ShettyAkarsh (approved):

benbcai:
Can #461 now be closed by this PR?

packages/eslint-config-terra/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Add Format-js linter.
```
benbcai, line 6: Does adding this linter potentially cause projects to fail if there are violations? If so, does this require a MVB?

saket2403: After making changes to show warnings for all the rules. 3 rules are showing warnings for all lines of code where we are using translations with `intl.formatMessage`:

- no-id- No Manual IDs can be used. (we are using IDs for the corresponding translation).  
- enforce-default-message
- enforce-description

More info on the rules [here](https://formatjs.io/docs/tooling/linter/). I was thinking if we should disable these 3 rules completely especially the `no-id` rule or is there any alternative to use translations without the ids. As each line of code using translation is throwing all 3 warnings there are about 600 warnings which will be errors when we go for the MVB.
cc @benbcai @ryanthemanuel .

pranav300 (approved):

packages/eslint-config-terra/CHANGELOG.md
ShettyAkarsh: @ryanthemanuel, @benbcai  so should we consider disabling these three rules ? as our code is not complaint for the same.

benbcai: @ShettyAkarsh @saket2403 I'd agree with disabling `no-id` and `enforce-default-message` for the reasons below but I not sure about `enforce-description`. We would want @ryanthemanuel and @tbiethman to weigh in.

- `no-id` - Since the id we use for the translation strings are name-spaced (e.g.; Terra.demographicsBanner.dateOfBirth) we don't expect any id collision so the reasons listed [here](https://formatjs.io/docs/getting-started/message-declaration/) is less applicable to us. Most of our consumers also name-space their ids. For this reason, I think having this rule is more disruptive that the value it provides since almost all of our usage would violate this rule.
- `enforce-default-message` - There is already logic to fall back to the `en` translation so requiring a default message here doesn't make much sense and would warn/error on most if not all usages.
- `enforce-description` - Having a description for the message that provides more context about this message makes sense but the purpose of the context seems to be for the translator as documented [here](https://formatjs.io/docs/react-intl/api/#message-descriptor). But in most cases, the translated string has already been translated with context provided by the team requesting the translations. So I think I'd vote for also removing this rule as it gives more noise than value.

ryanthemanuel: I am fine with disabling all three

benbcai (approved):

supreethmr (approved):

jmsv6d (approved):

PR #724, opened 2021-12-05
[duplicate-package-checker-webpack-plugin] add TypeScript typings

lagunovsky:
### Summary

This PR adds TypeScript typings


benbcai:
@lagunovsky Packages in terra-toolkit and in our other mono repos do not fully support TypeScript and are not ready to do so. The `duplicate-package-checker-webpack-plugin` package is a clone from https://github.com/darrenscerri/duplicate-package-checker-webpack-plugin. If you need TypeScript, I suggest to reach out to the maintainers of the original package.

lagunovsky:
I find this strange because if you don't use TypeScript, the `.d.ts` file doesn't affect anything.

The original repository has been out of support for a long time.

Okay, I'll have to create my own 🙂

PR #726, opened 2022-01-03
bumped package-json-lint

xenoworf:
### Summary
I was seeing errors installing terra-framework:
```
npm WARN @cerner/package-json-lint-config-terra@1.2.0 requires a peer of @cerner/package-json-lint@^0.0.0 but none is installed. You must install peer dependencies yourself.
```
It does not seem right to depend on a 0.0.0 version, so I bumped it to 1.0 This PR should fix that problem, once terra-framework is updated to consume the new version of this project.

### DIFF
--- a/packages/package-json-lint-config-terra/CHANGELOG.md
+++ b/packages/package-json-lint-config-terra/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Changed
+  * Minor dependency bump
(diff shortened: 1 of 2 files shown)


xenoworf:
Looks like 1.x is the first stable documented release, so I see no reason not to bump from v0 to v1 unless CI testing fails.

benbcai (approved):

jmsv6d (approved):

PR #732, opened 2022-02-17
[terra-functional-testing] Add useRemoteScreenshots cli option 

benbcai:
### Summary

Added `useRemoteScreenshots` cli option to allow downloading reference screenshots from a remote site for screenshot comparisons. The default value for `useRemoteScreenshots` is false which will continue to use the project's local reference screenshot like it is works today for passivity.

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-functional-testing/src/terra-cli/wdio/index.js
+++ b/packages/terra-functional-testing/src/terra-cli/wdio/index.js
@@ -184,7 +184,7 @@ const cli = {
         describe: 'Whether or not to automatically update all reference screenshots with the latest screenshots.',
         default: false,
       },
-      useRemoteScreenshots: {
+      useRemoteReferenceScreenshots: {
         type: 'boolean',
         describe: 'A flag to download reference screenshots from a remote site for screenshot comparisons instead of using the local reference screenshots.',
         default: false,
--- a/packages/terra-functional-testing/src/terra-cli/wdio/index.js
+++ b/packages/terra-functional-testing/src/terra-cli/wdio/index.js
@@ -184,6 +184,11 @@ const cli = {
         describe: 'Whether or not to automatically update all reference screenshots with the latest screenshots.',
         default: false,
       },
+      useRemoteScreenshots: {
+        type: 'boolean',
+        describe: 'A flag to download reference screenshots from a remote site for screenshot comparisons instead of using the local reference screenshots.',
+        default: false,
+      },
       useSeleniumStandaloneService: {
         type: 'boolean',
         describe: 'A flag to use the selenium standalone service instead of the selenium docker service.',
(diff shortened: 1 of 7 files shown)


packages/terra-functional-testing/src/terra-cli/wdio/index.js
```diff
@@ -184,6 +184,11 @@ const cli = {
         describe: 'Whether or not to automatically update all reference screenshots with the latest screenshots.',
         default: false,
       },
+      useRemoteScreenshots: {
```
benbcai: I'm open for other name suggestions.

sdadn:
What will happen if `useRemoteScreenshots` is false?

benbcai:
> What will happen if `useRemoteScreenshots` is false?

The thinking is that consumers can opt in for this feature. Therefore, the default will be false, which will behave llke how it does today. See above PR [description](https://github.com/cerner/terra-toolkit/pull/732#issue-1141685460).

packages/terra-functional-testing/src/terra-cli/wdio/index.js
xenoworf: Do you think we'll need to add flags to view remote diffs, or remote latest? If so - I suggest naming it `useRemoteRefScreenshots` so we have room to add those flags later. If not, or if we don't know, this name looks fine to me.

xenoworf (approved):

packages/terra-functional-testing/src/terra-cli/wdio/index.js
benbcai: I don't see us needing additional options for diff/latest screenshots since we are storing only reference screenshots. But I like your suggestion of `useRemoteRefScreenshots` to be more explicit. Do you prefer `useRemoteRefScreenshots` or `useRemoteReferenceScreenshots`?

sdadn: I prefer `useRemoteReferenceScreenshots` with a shorthand flag like `-R` if that's possible

benbcai: I've updated to name to `useRemoteReferenceScreenshots1`: https://github.com/cerner/terra-toolkit/pull/732/commits/aa66ee6480d4eadfde686aea0efbb0c0723c3d50

We previously only created the `u` alias for `updateScreenshots`. The reason was to mimic jest in that you can run  `npm run jest -- -u` given the occasional need to _update_ the jet snapshots on the fly. So there is value in creating the `u` alias so users don't need to type `updateScreenshots` every time in the terminal. 

However, the use case for `useRemoteReferenceScreenshots` is different in that once consumers opt in for this feature by setting it in their wdio test scripts in package.json they should not be flipping this option on and off, hence an alias provides little value. It is expected that when set and opted in they should be discouraged to not be going back to the old way of storing screenshots locally.

sdadn (approved):

BenBoersma (approved):

PR #740, opened 2022-05-05
[terra-functional-testing] Upload and download screenshots

benbcai:
### Summary

- This PR currently doesn't have any new test coverage added. I'm working on that now but wanted to get the PR out earlier for review.
- Added functions to upload and download screenshots from the remote repository.
- Moved reading repository URL configurations from packages.json to wdio.conf.js
- Removed code that deletes reference screenshots directory that was added before because it shouldn't be deleted for local develoment.
- Closes UXPLATFORM-6369

### Additional Details

@cerner/terra


packages/terra-functional-testing/src/config/wdio.conf.js
```diff
@@ -135,6 +135,13 @@ exports.config = {
     ui: 'bdd',
     timeout: 1200000,
   },
+  /**
+   * Defines the defaullt repository id and url used for uploading and downloading reference screenshots from a remote repository.
+   */
+  screenshotsSites: {
+    repositoryId: 'cerner-release-main-site',
```
xenoworf: I don't think we should hard-code Cerner-specific values in our OSS repos. No one outside of the company will be able to use them. How do we typically handle cases like this where we know we're the primary audience and consumer yet the code is in our open source repos?

One idea is to leave these two values (also line 143) undefined here, then later provide those values either in an internal wrapper project. A second option is to provide these values as environmental variables at runtime.

Lastly, if we don't mind putting values like this in our OSS repos then I guess it's fine to keep these as-is. It just seems odd to mix the two worlds. Kinda smells like we are not correctly separating the concerns between our projects.

xenoworf (commented):
Other than the question I left about the mixing internal value into our OSS world, this looks fine. If we can resolve that question then I'd +1 this PR.

packages/terra-functional-testing/src/commands/utils/MemoryStream.js
```diff
@@ -0,0 +1,21 @@
...
+ * MemoryStream is a PassThrough stream that keeps track of the length of the stream
+ */
+class MemoryStream extends stream.PassThrough {
+  constructor(options) {
+    super(options);
+
+    this.length = 0;
+  }
+
+  // eslint-disable-next-line no-underscore-dangle
```
sdadn, line 13: underscores are useful for internal or helper methods, do we want to disable this in the eslint.config instead of a single line disable?

packages/terra-functional-testing/src/config/wdio.conf.js
benbcai: @xenoworf This makes perfect sense. I think we should not provide a default and always require the consumer to provide the repository id and url if using this feature.
https://github.com/cerner/terra-toolkit/pull/740/commits/4b6b71dd14d9a363ef49d27716c7b3cd1fb3c7b4

packages/terra-functional-testing/src/commands/utils/MemoryStream.js
benbcai: `_transforms` is a [stream API](https://nodejs.org/api/stream.html#transform_transformchunk-encoding-callback) so we're not able to modify the name. I think we should still keep this rule enabled since we typically don't recommend using the underscore in our code. This is a copy/paste from orion-releas-site.

BenBoersma (approved):

xenoworf (approved):

PR #755, opened 2022-08-01
Consume git credentials to post PR comment

BenBoersma:
### Summary
 While using remote reference screenshots, if screenshot tests fail, post a comment to the PR for notification about changed screenshots.

### Additional Details

Closes  UXPLATFORM-6628

@cerner/terra


packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -144,7 +148,30 @@ class TerraService {
    */
   async onComplete(_, config) {
     try {
-      if (this.serviceOptions.useRemoteReferenceScreenshots && !this.serviceOptions.buildBranch.match(BUILD_BRANCH.pullRequest) && this.serviceOptions.buildType === BUILD_TYPE.branchEventCause) {
+      if (this.serviceOptions.useRemoteReferenceScreenshots && process.env.SCREENSHOT_MISMATCH_CHECK && this.serviceOptions.buildBranch.match(BUILD_BRANCH.pullRequest)) {
```
benbcai, line 151: Instead of using SCREENSHOT_MISMATCH_CHECK to indicate whether or not there are test failures, a better pattern would be to check using the test results returned by the `onComplete` hook.

See documentation: https://webdriver.io/docs/configurationfile

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
```diff
@@ -86,7 +86,9 @@ class TestRunner {
    * @param {string} options.externalHost - The host address the testing environment is connected to.
    * @param {number} options.externalPort - The port mapping from the host to the container.
    * @param {string} options.formFactors - A list of form factors for the test run.
+   * @param {object} options.gitApiUrl - 
    * @param {string} options.gridUrl - The remote selenium grid address.
+   * @param {string} options.issueNumber -
```
benbcai: Missing description.

Should `gitToken` and `buildUrl` be added for consistency?

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -144,7 +148,30 @@ class TerraService {
...
+        if (!this.serviceOptions.gitToken || !this.serviceOptions.gitApiUrl) {
+          throw new Error('No git token recieved');
+        }
+
+        const packageJson = fs.readJsonSync(path.join(process.cwd(), 'package.json'));
+        const repoUrl = new URL(packageJson.repository.url);
+        const repoName = repoUrl.pathname.match(/[^/]+/g);
+        const octokit = new Octokit({ baseUrl: `${this.serviceOptions.gitApiUrl}`, auth: `${this.serviceOptions.gitToken}` });
+        const message = `:warning: :bangbang: **WDIO MISMATCH** \n\nCheck that screenshot change is intended at: ${this.serviceOptions.buildUrl}`;
```
benbcai: This is what the Consumption Guide says about this alert at the end:

> Pull Request Build - Intentional changes in a project can alter wdio screenshots. Like the branch build, a pull request build will also download the reference screenshots from Nexus to use for comparisons. Unlike the branch build, should screenshot mistmaches occur the pull request build will not fail because the mismatches are expected due to the intentional changes in the pull request. When there are screenshot mismatches, an alert will be displayed in the pull request and to suggest to review the mismatches to verify the mismatches are expected. Latest sccreenshots generated by this type of build will also not be uploaded to Nexus.

Do we need to update the guide to provide better directions/steps to highlight what they need to do? Or perhaps add additional guidance on this message so they know what the expectations are.  
A consumer reading this message probably won't know how to proceed after reviewing the mismatch screenshots.

packages/terra-functional-testing/src/services/wdio-terra-service.js
BenBoersma: Originally I assumed that since we were manually making sure that mismatch failures counted as a success, any passed results would also be a success. Following that I assumed that I would have to search through all the tests' messages to see if the "Screenshot has changed and needs to be reviewed." string was present to detect, which seemed a tad too inefficient.

According to that documentation `onComplete` should run with 4 arguments with the 4th argument being a "results" object. When I console.log that 4th argument I get `undefined` as the answer. 

I also tried using the 1st argument, the exit code, but it reports success even when I purposely put in a failure screenshot but force `toMatchReference` calls to follow the remote reference screenshot logic in it.

packages/terra-functional-testing/src/services/wdio-terra-service.js
BenBoersma: Thinking on it I think a notice of what will happen if mismatches are still there upon merge is needed in the message. I updated the message to as follows but expect we will want to properly word this so I'm open to suggestions.

> :warning: :bangbang: **WDIO MISMATCH** 
> 
> If screenshot change is intended, remote reference screenshots will be updated upon PR merge.
> 
> Check that screenshot change is intended at: {buildURL}

https://github.com/cerner/terra-toolkit/pull/755/commits/627a4865cc4c1e83b773cfe1799d89d092b314b2#diff-a6ba3c1d16c08e58e7ee3994f3ab541b63b2a3a8dafb0de9daed2c8090c948d9R160

packages/terra-functional-testing/src/terra-cli/wdio/test-runner.js
BenBoersma: updated here https://github.com/cerner/terra-toolkit/pull/755/commits/627a4865cc4c1e83b773cfe1799d89d092b314b2#diff-f13ec26e86b6e8d765ec9020014d38600b7df8643e37a26d40000d28807e2bfbR91

packages/terra-functional-testing/src/config/utils/getConfigurationOptions.js
```diff
@@ -12,12 +12,16 @@ const getConfigurationOptions = (options) => {
     browsers,
     buildBranch,
     buildType,
+    buildUrl,
     disableSeleniumService,
     ignoreScreenshotMismatch,
```
sdadn: nit: can you move this to the proper alphabetical location since we're already updating the file?

+1 on maintain alphabetical order for all the other additions

BenBoersma: Reorderd here https://github.com/cerner/terra-toolkit/pull/755/commits/8cf3c9541ebb57fc411545d6df927c6e950b05b8

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: I would think the results should be returned in the the parameter. I think the current implementation would work but just feels a bit hacky. I guess we can keep the current implementation if we don't know why the returned result is not what we expected.

packages/terra-functional-testing/src/services/wdio-terra-service.js
benbcai: Does the buildURL link to the build or to the build artifacts showing the screenshot differences?

Do we think it is more logical to swap the order of these two sentences? It seems more logical to me that we should first tell where to look and then explain what will happen. 

For example:
```
Check that screenshot change is intended at: ${this.serviceOptions.buildUrl}
If screenshot change is intended, remote reference screenshots will be updated upon PR merge.
```

Should we also add something about what to do if the screenshots are not intended after reviewing them and emphasize that they should carefully review the mismatches? Normally when a PR has a successful build, it is easy for consumers to think that everything is good to go. But in this case, they should closely review any unintended mismatches and get those corrected. So I think it's important to call out what to do when there are unintended mismatches.

Basically whenever this alert is displayed, there are two possible actions to take:

1. If the screenshot mismatches are intended, they don't need to do anything with the understand that the latest screenshots will get uploaded up PR merge. This is now called out in the alert message.
2. If the screenshot mismatches are not intended, they should go figure out why the latest screenshots are not what they expect and get that fixed. This is the part that I think we should mention in the alert message. Otherwise, they could easily ignore the unintended mismatches and merge the PR when they should really get those unintended mismatches fixed.

I'm curious to hear what other people think.

sdadn (approved):

packages/terra-functional-testing/src/services/wdio-terra-service.js
BenBoersma: Well the other concern is, aren't we marking the test a pass anyway? So even if we got the results the only way to know if a test failed due to mismatch is to read through all the messages for each test and find at least one that has either our special added message or the mismatch message?

packages/terra-functional-testing/src/services/wdio-terra-service.js
BenBoersma: buildURL will link right to the specific job main page of that branch.

Yea that sounds good to me so how about

> Check that screenshot change is intended at: ${this.serviceOptions.buildUrl}
> If screenshot change is intended, remote reference screenshots will be updated upon PR merge.
> If screenshot change is untended, please fix screenshot issues before PR merge to prevent them from being uploaded.

I'm also curious what people think about the 1 comment per PR thing. Because I am worried about what happens when they get that message early, make a bunch of changes, and a new issue in screenshots is added that they don't catch because the message only posts on the first time a build would fail due to screenshot mismatch? I get we don't want to spam comment on a PR so I definitely think we need to mention we're only posting once per PR with this message whether that be in the message we post or in the documentation for updating.

BenBoersma: New message here https://github.com/cerner/terra-toolkit/pull/755/commits/b157573530a80aae84b5707f582dac14aa87472a#diff-a6ba3c1d16c08e58e7ee3994f3ab541b63b2a3a8dafb0de9daed2c8090c948d9R160

benbcai (approved):

PR #758, opened 2022-08-22
tests for github util command module

xenoworf:
### Summary
I renamed my command util module to `github.js` and added tests.


packages/terra-functional-testing/CHANGELOG.md
```diff
@@ -3,8 +3,8 @@
 ## Unreleased
 
 * Added
-  * Added a util module for calling github's API.
-  * Added a logic to post to the github PR if during a PR build and there are mismatchs.
+  * Added a util module and tests for calling GitHub's API.
+  * Added a logic to post to the GitHub PR if during a PR build and there are mismatches.
```
sdadn:
```suggestion
  * Added a util module and tests for calling GitHub's API.
  * Added logic to post to the GitHub PR if there are mismatches during a PR build.
```

sdadn (approved):

packages/terra-functional-testing/src/commands/utils/github.js
```diff
@@ -0,0 +1,71 @@
...
+ * @typedef PrLookupInfo
+ * @type {object}
+ * @property {string} githubHostname
+ * @property {string} repoOwner
+ * @property {string} repoName
+ * @property {string} prNumber
+ */
+
+/**  Info about a PR, fetched from a github instance. */
+class PrInfo {
```
benbcai: What is the reason for making this a class? Is it because we plan to add functions to it in the future?

xenoworf: No. I made it a class merely so that it could serve as a datatype that consumers can easily know the structure of.

xenoworf: Oh wait: yes kind of :).  I'm going to re-work this class as an adapter - it will take in the raw return JSON from the octokit API but expose only the info we know we need. Like a simplification filter. This will keep the rest of our code from having to know the github ReST API JSON. I'll need to add one for @BenBoersma's code as well.

benbcai: Ok. At first I was thinking that `lookupPR` could just return an object with these attributes instead of creating a class for it.

xenoworf:
Closing this. I'll start a new PR to make it cleaner.

PR #759, opened 2022-08-24
Githubclient

xenoworf:
### Summary
Re-worked the github api wrapper, now called `GitHubRepo`, to handle posting the mismatch warning. The point is that we only implement octokit once in the package. `GitHubRepo` will eventually also be used to detect the branch of the PR - which was why I originally wrote it.

Note: right now the tests are bogus. I am re-factoring them while I await your feedback on


packages/terra-functional-testing/src/commands/utils/GitHubRepo.js
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Information about a PR.
+ * @typedef {Object} Pull
+ * @property {number} number
+ * @property {Object} head - The head branch
+ * @property {string} head.ref
+ * @property {Object} base - The base branch
+ * @property {string} base.ref
+ */
```
sdadn: What is this referring to?

packages/terra-functional-testing/src/commands/utils/GitHubRepo.js
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Create a simplified view of a PR.
+ *
+ * @param {object} json A PR from the ReST API.
+ * @returns {Pull}
+ */
+function filterPull(json) {
```
sdadn: nit: would `filterPullRequest` be a better name? `filterPull` is confusing out of context. `filterPR` also works if you want to minimize characters.

packages/terra-functional-testing/src/commands/utils/GitHubRepo.js
xenoworf: It's the [ref](https://git-scm.com/book/en/v2/Git-Internals-Git-References) of the branch. It's what I will use later to check if the branch is `master` or other.

xenoworf: Fixed in a7cf278f59aa9da516cd4c5da20645f2c7d162cf

packages/terra-functional-testing/src/commands/utils/GitHubRepo.js
xenoworf: Fixed in a7cf278f59aa9da516cd4c5da20645f2c7d162cf.

packages/terra-functional-testing/src/commands/utils/GitHubRepo.js
```diff
@@ -0,0 +1,150 @@
...
+ * @see {@link https://git-scm.com/book/en/v2/Git-Internals-Git-References|Git References} to learn about refs.
+ */
+
+/**
+ * Create a simplified view of a PR.
+ *
+ * @param {object} json A PR from the ReST API.
+ * @returns {PullRequest}
+ */
+function filterPullRequest(json) {
```
sdadn, line 31: `json` looks pretty similar to built in [JSON Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON). I would rename it to something more descriptive.

packages/terra-functional-testing/src/commands/utils/GitHubRepo.js
```diff
@@ -0,0 +1,150 @@
...
+ * Information about an issue's comments.
+ * @typedef {Array} IssueComments
+ * @property {string} body The comment message.
+ * @property {string} issueUrl
+ */
+
+/**
+ * Create a simplified view of issue comments.
+ *
+ * @param {Array} json Comments from the ReST API.
```
sdadn, line 54:
```suggestion
 * @param {Array} JSON Comments from the REST API.
```

packages/terra-functional-testing/src/commands/utils/GitHubRepo.js
```diff
@@ -0,0 +1,150 @@
...
+ * @property {Object} base - The base branch
+ * @property {string} base.ref
+ *
+ * @see {@link https://git-scm.com/book/en/v2/Git-Internals-Git-References|Git References} to learn about refs.
+ */
+
+/**
+ * Create a simplified view of a PR.
+ *
+ * @param {object} json A PR from the ReST API.
```
benbcai, line 28: I think Object is should be capitalized according to https://jsdoc.app/tags-param.html
```suggestion
 * @param {Object} json A PR from the ReST API.
```

packages/terra-functional-testing/src/commands/utils/GitHubRepo.js
```diff
@@ -0,0 +1,150 @@
...
+   */
+  constructor(path, options) {
+    [, this.owner, this.repo] = path.split('/');
+    this.octokitOptions = {};
+    if (options && options.baseUrl) {
+      this.octokitOptions.baseUrl = options.baseUrl;
+    }
+    if (options && options.personalAccessToken) {
+      this.octokitOptions.auth = options.personalAccessToken;
+    }
```
benbcai, line 88: Would it work if this gets condensed to something like this?

```
this.octokitOptions = {
  baseUrl: options.baseUrl,
  auth: options.personalAccessToken
}
```

PR #763, opened 2022-09-06
Fixed a reference bug

xenoworf:
### Summary
The new onComplete logic of terra-wdio-service was allowing some invalid logic: [calling a function with an undefined argument](https://github.com/cerner/terra-toolkit/blob/main/packages/terra-functional-testing/src/services/wdio-terra-service.js#L173-L174). This was masked from our test runs because unhandled promises are not being flagged as failures in our CI setup. Future versions of node will fail these. 

I fixed it by adding "do you have this thing?" to the if-check above the logic.

### Additional Details
We should update our CI system to fail tests due to unhandled promises ASAP. I bet we have more of these latent bugs being emitted only as warnings.

### DIFF
--- a/packages/terra-functional-testing/src/services/wdio-terra-service.js
+++ b/packages/terra-functional-testing/src/services/wdio-terra-service.js
@@ -42,8 +42,8 @@ class TerraService {
    */
   async onPrepare(config) {
     try {
-      if (this.serviceOptions.useRemoteReferenceScreenshots) {
-        const screenshotConfig = config.getRemoteScreenshotConfiguration ? config.getRemoteScreenshotConfiguration(config.screenshotsSites, this.serviceOptions.buildBranch) : {};
+      if (this.serviceOptions.useRemoteReferenceScreenshots && config.getRemoteScreenshotConfiguration) {
+        const screenshotConfig = config.getRemoteScreenshotConfiguration(config.screenshotsSites, this.serviceOptions.buildBranch);
         const screenshotRequestor = new ScreenshotRequestor(screenshotConfig.publishScreenshotConfiguration);
         await screenshotRequestor.download();
       }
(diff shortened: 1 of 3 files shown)


sdadn (approved):

packages/terra-functional-testing/src/services/wdio-terra-service.js
```diff
@@ -169,8 +169,13 @@ class TerraService {
             throw Error(`Error posting issue comment. Status code: ${postCommentResult.status}`);
           }
         }
-      } else if (this.serviceOptions.useRemoteReferenceScreenshots && !this.serviceOptions.buildBranch.match(BUILD_BRANCH.pullRequest) && this.serviceOptions.buildType === BUILD_TYPE.branchEventCause) {
-        const screenshotConfig = config.getRemoteScreenshotConfiguration ? config.getRemoteScreenshotConfiguration(config.screenshotsSites, this.serviceOptions.buildBranch) : {};
+      } else if (
+        this.serviceOptions.useRemoteReferenceScreenshots
+        && !this.serviceOptions.buildBranch.match(BUILD_BRANCH.pullRequest)
+        && this.serviceOptions.buildType === BUILD_TYPE.branchEventCause
+        && config.getRemoteScreenshotConfiguration
```
benbcai, line 176: Can we add the same check in [onPrepare](https://github.com/cerner/terra-toolkit/pull/763/files#diff-a6ba3c1d16c08e58e7ee3994f3ab541b63b2a3a8dafb0de9daed2c8090c948d9R46)?

xenoworf: Done 8f7600c5f6b2388b139c9ff7b3d41d837cf9b1f9.

benbcai (approved):

xenoworf:
This is probably going to get merged with #764.

PR #764, opened 2022-09-07
Determine targeted branch name for pr

xenoworf:
### Summary
When WDIO's onComplete() runs: we will upload the updated screenshots of a merged PR build to a remote repo. Part of this change also changes the unused github util module with an octokit wrapper. The wrapper just takes any thought out of choosing an auth strategy - it requires using PATs.

### Additional Details
I fixed a couple test bugs too while I was in there. I was worried some fails positives or fails would be masked by them.

PR #765, opened 2022-09-28
Determine download branch for remote screenshots - and more - attempt 3

xenoworf:
### Summary
* When using remote screenshots: in onPrepare, if the buildBranch is a PR (follows the `pr-123` pattern), then download reference screenshots from a remote repo given by the name (the ref) of the PR's base branch. If the branch is not a PR (e.g `main` then download the reference screenshots from the remote repo given by the buildBranch itself.
* A ton of refactoring of code and tests due to some logic bugs, some test bugs, and some other misc. needs.


packages/terra-functional-testing/src/services/wdio-terra-service/WDIOTerraService.js
```diff
@@ -0,0 +1,280 @@
...
+        const gp = new GithubPr(
+          gitApiUrl,
+          gitToken,
+          metadata.owner,
+          metadata.repo,
+          issueNumber,
+        );
+
+        const baseBranch = await gp.getBaseBranchRef();
+        screenshotConfig = this.getRemoteScreenshotConfiguration(this.screenshotsSites, baseBranch);
```
benbcai, line 87: Should we add back this [check for getRemoteScreenshotConfiguration](https://github.com/cerner/terra-toolkit/pull/764/commits/82348fcecf4085e1a672cd79259e01464d17930d#diff-a6ba3c1d16c08e58e7ee3994f3ab541b63b2a3a8dafb0de9daed2c8090c948d9R66-R68) in case it is not configured correctly in the consumer's wdio.config.js?

benbcai: Nevermind. This is all in a try/catch block so it would catch if there is an error.

packages/terra-functional-testing/src/services/wdio-terra-service/GithubIssue.js
```diff
@@ -0,0 +1,28 @@
+const { Octokit } = require('@octokit/core');
+
+class GithubIssue {
+  constructor(url, token, owner, repo, number) {
+    this.ok = new Octokit({ baseUrl: url, auth: token });
```
benbcai: Can we rename `this.ok` to `this.octoKit` or something similar so it's more legible?

xenoworf: Sure no problem

packages/terra-functional-testing/src/services/wdio-terra-service/GithubPr.js
```diff
@@ -0,0 +1,10 @@
+const GithubIssue = require('./GithubIssue');
+
+class GithubPr extends GithubIssue {
+  async getBaseBranchRef() {
```
benbcai, line 4: I'm curious why we separate this into its own file and not just keeping it in GithubIssue.js?

xenoworf: Only to follow the one-class-per-file rule that the linter defaults to.

xenoworf: Also Issues don't have this branch data. PRs do though. According to the GitHub API doc, every PR is an Issue, but not every Issue is a PR. So that's why I have them as two separate classes.

benbcai: Thanks for the clarifications!

benbcai (approved):

packages/terra-functional-testing/src/services/wdio-terra-service/GithubIssue.js
xenoworf: done

BenBoersma (approved):

PR #791, opened 2023-07-10
[terra-functional-testing] Add option to use HTTPS when gridUrl is provided

artpark:
### Summary

**What was changed:**
A new flag for enabling an https connection when the `gridUrl` parameter is also provided.

**Why it was changed:**
This was added in response to needing to point to a orion-0 cluster, which all need an https connection. 

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [ ] Other (please describe below)
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

ION-47621 

---

@cerner/terra

### DIFF
--- a/packages/terra-functional-testing/tests/jest/webpack-server/webpack-server.test.js
+++ b/packages/terra-functional-testing/tests/jest/webpack-server/webpack-server.test.js
@@ -42,12 +42,13 @@ describe('Webpack Server', () => {
         port: 'mock',
         theme: 'mock',
         gridUrl: '1.1.1.1',
+        useHttps: true,
       });
 
       expect(server.config).toEqual('config');
       expect(server.host).toEqual('0.0.0.0');
       expect(server.port).toEqual('mock');
-      expect(server.gridStatusUrl).toEqual('http://1.1.1.1:80/status');
+      expect(server.gridStatusUrl).toEqual('https://1.1.1.1:80/status');
     });
   });
 
--- a/packages/terra-toolkit-docs/CHANGELOG.md
+++ b/packages/terra-toolkit-docs/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 * Added
-  * Added `useHttps` to use a secure http connection when the `gridUrl` flag is also provided.
+  * Added `useHttps` flag information to `terra-functional-testing` documentation.
 
 ## 2.13.0 - (April 27, 2023)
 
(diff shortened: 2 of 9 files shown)


sycombs (approved):

BenBoersma (approved):

packages/terra-functional-testing/tests/jest/webpack-server/webpack-server.test.js
```diff
@@ -42,12 +42,13 @@ describe('Webpack Server', () => {
...
         theme: 'mock',
         gridUrl: '1.1.1.1',
+        useHttps: true,
       });
 
       expect(server.config).toEqual('config');
       expect(server.host).toEqual('0.0.0.0');
       expect(server.port).toEqual('mock');
-      expect(server.gridStatusUrl).toEqual('http://1.1.1.1:80/status');
+      expect(server.gridStatusUrl).toEqual('https://1.1.1.1:80/status');
```
sdadn, line 51: We probably want to keep the existing unit test as that is still a valid use case; we should add a second test for https.

packages/terra-toolkit-docs/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added `useHttps` to use a secure http connection when the `gridUrl` flag is also provided.
```
sdadn: nit: this reads as if it's a source code change when this is the documentation package. We can probably rephrase this as:
```suggestion
  * Added `gridUrl` flag information to `terra-functional-testing` documentation.
```

packages/terra-functional-testing/tests/jest/webpack-server/webpack-server.test.js
artpark: I think this one is alright because the test previously already passed in gridUrl, so this is just building on top of that test since that will be the only use case for `useHttps` anyways

packages/terra-toolkit-docs/CHANGELOG.md
artpark: Changed [2e82aa7](https://github.com/cerner/terra-toolkit/pull/791/commits/2e82aa7fec0992bc29e76afdf2c04951163828d8)

packages/terra-functional-testing/tests/jest/webpack-server/webpack-server.test.js
sdadn: That makes sense then. +1

sdadn (approved):

PR #824, opened 2023-08-30
Fixed package-json-lint bug and updated eslint-config-terra devDependencies

sdadn:
### Summary

This PR has the following changes:

- Fixes an installation bug with `package-json-lint`.
- Updates `eslint-config-terrra` and `jest-config-terra` devDependency format from `^x.y.z` to `x`. Both of these formats are identical, this was changed so that changes to these packages do not cause minor version bumps in other terra-toolkit packages and unnecessary releases like in #823 .
- Locks `eslint-plugin-formatjs` to `2.19.1` as the latest 2.x version has a peerDependency of `eslint@8` which conflicts with the `eslint-config-terra` peerDependency of `eslint@7`.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] Other (please describe below)
- [ ] No tests are needed

This was tested by:
1. Running `npm install -D lerna@6` at the root of the package
2. Running `npx lerna run --parallel clean:install` at the root level. This will cause the dependencies to be installed at the package level for each package.
3. Running `npx lerna run --parallel compile`.
4. Running `npx lerna run --parallel lint`. This will test the linter dependency updates by running the linter at the package level.  

The result should be something like this:

CleanShot 2023-08-30 at 15 04 30

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review
- [x] Tech verification

---

@cerner/terra


adoroshk (approved):

sdadn:
> Fails
> 🚫	Please include a CHANGELOG entry for each changed package on this PR. Looks like a CHANGELOG entry is missing for:
> * browserslist-config-terra
> * duplicate-package-checker-webpack-plugin
> * jest-config-terra
> * package-json-lint-config-terra
> * stylelint-config-terra
> * terra-aggregate-translations
> * terra-cli
> * terra-enzyme-intl
> * terra-open-source-scripts
> * terra-toolkit-docs
> * webpack-config-terra
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [500cdbb](https://github.com/cerner/terra-toolkit/commit/500cdbbbc1aa26e7839e2ad1e07bc278be793860)

These changes do not need to be released

adoroshk:
Installed lerna, run clean:install, compile and lint commands, which successfully worked. 
### Clean:install
Screenshot 2023-08-31 at 9 41 15 AM

### Compile
Screenshot 2023-08-31 at 9 41 03 AM

### Lint
Screenshot 2023-08-31 at 9 38 45 AM

packages/browserslist-config-terra/package.json
```diff
@@ -35,7 +35,7 @@
     "lint:js:fix": "eslint --ext .js,.jsx . --fix"
   },
   "devDependencies": {
-    "@cerner/eslint-config-terra": "^5.5.1",
+    "@cerner/eslint-config-terra": "5",
```
sycombs, line 38: Why don't we want to consume the minor version dependency updates? Wouldn't we want to have the latest changes?

packages/duplicate-package-checker-webpack-plugin/package.json
```diff
@@ -50,8 +50,8 @@
   "devDependencies": {
     "@babel/cli": "^7.22.10",
     "@babel/core": "^7.22.10",
-    "@cerner/eslint-config-terra": "^5.5.1",
-    "@cerner/jest-config-terra": "^1.6.2",
+    "@cerner/eslint-config-terra": "5",
+    "@cerner/jest-config-terra": "1",
```
kenk2, line 54: I have the same question that @sycombs has here. This seems to still take on minor versions, but I think we should keep the specific versioning so consumers know more easily which working version they ought to roll back to.

packages/browserslist-config-terra/package.json
sycombs: Ok, looking at the [npm docs](https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept), this will still take minor versions, but not patch versions?

packages/eslint-config-terra/package.json
```diff
@@ -34,16 +34,16 @@
     "lint:js:fix": "eslint --ext .js,.jsx . --fix"
   },
   "dependencies": {
-    "@babel/eslint-parser": "^7.12.13",
-    "eslint-config-airbnb": "^18.2.1",
-    "eslint-plugin-compat": "^3.9.0",
-    "eslint-plugin-formatjs": "^2.17.8",
-    "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-jsx-a11y": "^6.4.1",
+    "@babel/eslint-parser": "7",
```
kenk2, line 37: Also here too, why are we not using the minor/patch versions anymore?

packages/browserslist-config-terra/package.json
sycombs: Actually I think I understand now. This is just so that the version bumps don't appear in package.json and trigger a new release, but running `npm install` will still install the latest version?

sdadn: Correct! Both these syntax are identical. This is so that these dependencies don't get updated and trigger release. Since these are devDependencies, they aren't relevant for consumers and don't need to be released. e.g. #823 where only eslint-config-terra had releasable changes but everything else had to be released to due to a simple dev dependency bump.

packages/duplicate-package-checker-webpack-plugin/package.json
sdadn: These syntax are identical according to the [npm docs](https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept). The reason this was changed is so that the numbers don't constantly get bumped when we make changes to these packages. Since these are devDependencies, these are not relevant for consumers. This change was made so that we don't have unnecessary releases simply because there are devDependency updates, such as in #823.

packages/eslint-config-terra/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Locked `eslint-plugin-formatjs` to `2.19.1` as it has a peerDependency of `eslint@7` and no longer conflicts with `eslint-config-terra`.
```
sycombs: Can we reword this to something like the following so that it is clear that the change is to ensure that both `eslint-plugin-formatjs` and `eslint-config-terra` have the same peer dependency version of `eslint`?

```suggestion
  * Locked `eslint-plugin-formatjs` to `2.19.1` so that the version of its peerDependency `eslint@7` matches that in `eslint-config-terra`.
```

packages/eslint-config-terra/package.json
sdadn: The syntax is identical. in this case, the change was made for readability as we are locking a few packages and it was difficult to identify the without going line by line.

packages/package-json-lint/package.json
```diff
@@ -33,8 +33,7 @@
     "compile": "babel src --out-dir lib --copy-files --ignore src",
     "jest": "jest",
     "lint": "eslint --ext .js,.jsx --ignore-path ../../.eslintignore .",
-    "lint:js:fix": "eslint --ext .js,.jsx . --fix",
-    "postinstall": "npm run compile"
```
sycombs, line 37: Just curious why we don't need this script.

sdadn: `postinstall`? It is redundant as the project already gets compiled at the monorepo level and regnerates the `lib` which already exists in the released package.

sycombs: Ok I thought this might be useful for package-level development. But we can just run the compile script separately if needed.

sycombs (approved):

packages/package-json-lint/package.json
sdadn: The other issue is that this can cause installation failures for consumers if we are not publishing `src` as it'll try to run this script during `npm install`. So this is something we would want to figure out once we normalize package-level development. But that would be a separate discussion/PR.

kenk2 (approved):

artpark (approved):

PR #827, opened 2023-09-06
Workspaces upgrade

kenk2:
### Summary

**What was changed:**
The build process now utilizes NPM workspaces.

**Why it was changed:**
To keep up with the node / npm ecosystem.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] Other (please describe below)
- [ ] No tests are needed

Running and building the application without lerna and utilizing NPM workspaces.
### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "version": "5.25.0",
   "description": "Utilities to help when developing terra modules.",
   "engines": {
-    "node": "10 || 12|| 14",
+    "node": "10 || 12 || 14",
     "npm": "6 || 9"
   },
   "repository": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -4,13 +4,16 @@
   "version": "5.25.0",
   "description": "Utilities to help when developing terra modules.",
   "engines": {
-    "node": "10 || 12|| 14",
+    "node": "10 || 12 || 14",
     "npm": "6 || 9"
```
sdadn: Since workspaces is a npm 9 feature, we should remove npm 6.

package.json
```diff
@@ -73,13 +76,12 @@
...
   },
   "scripts": {
-    "bootstrap:hoist": "lerna bootstrap --hoist --force-local",
     "clean": "npm run clean:root && npm run clean:lib",
     "clean:install": "rm package-lock.json; npm run clean && time npm install",
     "clean:lib": "rm -rf packages/**/lib",
     "clean:root": "$(lerna clean --yes || true) && echo 'removing ./node_modules' && rm -rf node_modules && echo 'removed ./node_modules'",
     "compile:prod": "rm -rf ./build; NODE_ENV=production webpack --mode=production --config webpack.dev-site.config.js",
-    "compile": "npm run clean:lib && lerna run compile --parallel",
+    "compile": "npm run clean:lib && lerna run compile",
```
sdadn, line 84: why is the parallel flag removed?

kenk2: @sdadn Parallel  is already a feature in lerna v6, so we don't really need to specify it: https://lerna.js.org/docs/lerna6-obsolete-options#--parallel

package.json
kenk2: Done!

sdadn (approved):

package.json
```diff
@@ -4,13 +4,16 @@
   "version": "5.25.0",
   "description": "Utilities to help when developing terra modules.",
   "engines": {
-    "node": "10 || 12|| 14",
-    "npm": "6 || 9"
```
nikhitasharma, line 8: Are we okay going ahead and removing npm6 support based on discussion during retro?

kenk2: We will be, because the upgrades won't affect consumers. @nikhitasharma

adoroshk (approved):

PR #839, opened 2023-12-20
[terra-functional-testing] Fixing Nexus screenshot bugs

BenBoersma:
### Summary
Fixed Nexus upload/download of screenshots for multiple nodes.
Fixed PR commenting when mismatch was detected. 
Fixed logging for when doing a PR build or Screenshot Branch commit build not showing which tests had mismatches.

**What was changed:**
Nexus screenshots are now stored as multiple directories for each node being run. Directories are named as following with undefined values not being present '$THEME-$LOCALE-$BROWSER-$FORMFACTOR$' I.E 'terra-default-theme-en-chrome-large'

PR commenting is now based on the existence of a file that is made during `toMatchReference` calls instead of via process.env variables. This file is deleted on test suite completion.

Validates.screenshot now passes in the test name to its 'toMatchReference' calls, so that when 'toMatchReference' is doing Nexus mismatch logic it can now output a log message of the test that fails due to mismatch so that in can be found by Nexus users.

**Why it was changed:**

### Testing

This change was tested using:

- [ ] WDIO
- [x] Jest
- [ ] Other (please describe below)
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9299 

---

@cerner/terra


packages/terra-functional-testing/src/commands/expect/toMatchReference.js
```diff
@@ -41,11 +47,19 @@ function toMatchReference(screenshot) {
   if (global.Terra.serviceOptions.useRemoteReferenceScreenshots && !pass
     && (global.Terra.serviceOptions.buildBranch.match(BUILD_BRANCH.pullRequest) || global.Terra.serviceOptions.buildType === BUILD_TYPE.branchEventCause)) {
     pass = true;
-    process.env.SCREENSHOT_MISMATCH_CHECK = true;
-    if (message.length > 0) {
-      message = message.concat('\n');
+    const outputDir = getOutputDir();
+    const fileName = path.join(outputDir, 'ignored-mismatch.json');
+    // Create the output directory if it does not already exist.
+    if (!fs.existsSync(outputDir)) {
```
kenk2, line 53: nit: Is it possible to turn this into an `async` function? the Sync API from node is infamous for being a blocking API (meaning that it needs to wait for this to complete before doing anything else in node). This will affect performance down the road, especially when we change this to run multiple suites at a time. I might suggest using the equivalent promises API's from node: https://nodejs.org/docs/latest-v14.x/api/fs.html#fs_promises_api

packages/terra-functional-testing/src/commands/utils/ScreenshotRequestor.js
```diff
@@ -80,6 +81,26 @@ class ScreenshotRequestor {
...
+   * Naming convention is {theme}-{locale}-{browser}-{formfactor} and if provided string is empty does not add it.
+   * @param {string} locale - the locale to use when downloading
+   * @param {string} theme - the theme to use when downloading
+   * @param {string} formFactor - the formFactor to use when downloading
+   * @param {string} browser - the browser to use when uploading
+   */
+  static makeReferenceName(locale, theme, formFactor, browser) {
+    const referenceName = [theme, locale, browser, formFactor].filter(
+      (str) => {
+        if (str !== '' || str !== undefined) {
```
kenk2: nit: Are there any other types of falsey's we have to worry about? What would happen here with `null`?

packages/terra-functional-testing/src/services/wdio-terra-service/WDIOTerraService.js
```diff
@@ -256,13 +263,21 @@ class WDIOTerraService {
       return;
     }
 
+    const fileName = path.join(getOutputDir(), 'ignored-mismatch.json');
+
     try {
-      if (process.env.SCREENSHOT_MISMATCH_CHECK === 'true' && buildBranch.match(BUILD_BRANCH.pullRequest)) {
+      if (fs.existsSync(fileName) && buildBranch.match(BUILD_BRANCH.pullRequest)) {
```
kenk2: Like the above comment: I'd try to avoid the `sync` api as much as possible. This is also already an `async` function so we should be using the async (Promises) api as much as possible.

kenk2 (changes requested):

packages/terra-functional-testing/src/commands/utils/ScreenshotRequestor.js
BenBoersma: There shouldn't be but reconsidering this, we're just checking for falsy values so might as well filter them all out for an easy code run so changed here
https://github.com/cerner/terra-toolkit/pull/839/commits/0f73f13811532ac2ef8026b400ae1551315c5ff9

packages/terra-functional-testing/src/commands/expect/toMatchReference.js
BenBoersma: I don't know if it is possible, at the very least I have a suspicion that would be a major version bump since we would be changing a main testing function to be `async` when it wasn't before.

packages/terra-functional-testing/src/services/wdio-terra-service/WDIOTerraService.js
BenBoersma: alright finally got a moment to updated this, moved off the sync api and onto the async ones
https://github.com/cerner/terra-toolkit/pull/839/commits/3b352ec7520e8b85f08ff21aafbc8ae8aaa9a2fb

packages/terra-functional-testing/tests/jest/commands/expect/toMatchReference.test.js
```diff
@@ -299,12 +326,69 @@ describe('toMatchReference', () => {
...
+        useRemoteReferenceScreenshots: true,
+      },
+    };
+    const receivedScreenshot = {
+      isSameDimensions: false,
+      misMatchPercentage: 0.10,
+    };
+    jest.spyOn(fs, 'existsSync').mockImplementationOnce(() => false);
+    jest.spyOn(fs, 'mkdirSync').mockImplementationOnce(() => false);
+    jest.spyOn(fs, 'writeFileSync').mockImplementationOnce(() => false);
```
sdadn: We can also use the shorthand [mockFn.mockReturnValueOnce](https://jestjs.io/docs/mock-function-api#mockfnmockreturnvalueoncevalue)

packages/terra-functional-testing/tests/jest/commands/expect/toMatchReference.test.js
```diff
@@ -240,12 +260,16 @@ describe('toMatchReference', () => {
       isSameDimensions: false,
       misMatchPercentage: 0.10,
     };
+    jest.spyOn(fs, 'existsSync').mockImplementation(() => true);
+    jest.spyOn(fs, 'existsSync').mockImplementation(() => true);
```
sdadn: We can use [mockFn.mockReturnValue(value)](https://jestjs.io/docs/mock-function-api#mockfnmockreturnvalueoncevalue) as well.

BenBoersma: Fixed here: https://github.com/cerner/terra-toolkit/pull/839/commits/e3f1c358ea6ee77af9b877d072b9ca9f77df5871#diff-4fc55674149c158d6de22cac25bcc78653afd13a67eb1270cdd05c57b4abe9ebR263

packages/terra-functional-testing/tests/jest/commands/expect/toMatchReference.test.js
BenBoersma: Fixed here: https://github.com/cerner/terra-toolkit/pull/839/commits/e3f1c358ea6ee77af9b877d072b9ca9f77df5871#diff-4fc55674149c158d6de22cac25bcc78653afd13a67eb1270cdd05c57b4abe9ebR350-R352

sdadn (approved):

kenk2 (approved):

